PR #25, opened 2024-08-13
Refactor fill time periods function

Sukh-P:
# Pull Request

## Description
Small refactor of `fill_time_periods` to remove using `iterrows()` to hopefully make it more efficient 

## How Has This Been Tested?

A unit test already existed for this which is great so just checked that still passes

### DIFF
--- a/ocf_data_sampler/select/fill_time_periods.py
+++ b/ocf_data_sampler/select/fill_time_periods.py
@@ -1,15 +1,12 @@
-"""Select time periods"""
+"""fill time periods"""
 
 import pandas as pd
 import numpy as np
 
 
-
 def fill_time_periods(time_periods: pd.DataFrame, freq: pd.Timedelta):
-    datetimes = []
-    for _, row in time_periods.iterrows():
-        start_dt = pd.Timestamp(row["start_dt"]).ceil(freq)
-        end_dt = pd.Timestamp(row["end_dt"])
-        datetimes.append(pd.date_range(start_dt, end_dt, freq=freq))
+    start_dts = time_periods["start_dt"].apply(lambda x: pd.Timestamp(x).ceil(freq))
+    end_dts = time_periods['end_dt'].apply(lambda x: pd.Timestamp(x))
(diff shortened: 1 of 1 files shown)


dfulu (changes requested):
Could we use:

```
 start_dts = pd.to_datetime(time_periods["start_dt"]).ceil(freq)
 end_dts = pd.to_datetime(time_periods['end_dt'])
```
?

Sukh-P:
> Could we use:
> 
> ```
>  start_dts = pd.to_datetime(time_periods["start_dt"]).ceil(freq)
>  end_dts = pd.to_datetime(time_periods['end_dt'])
> ```
> 
> ?

Yeah I thought of that and tried it but the tests failed, since they were strings in the test, but not sure if we always assume they are pd.Timestamps anyway and then we should change the test? If so then yeah removing that makes sense

Sukh-P:
> Could we use:
> 
> ```
>  start_dts = pd.to_datetime(time_periods["start_dt"]).ceil(freq)
>  end_dts = pd.to_datetime(time_periods['end_dt'])
> ```
> 
> ?

So that won't work for the start_dts but yeah that's a nicer way to do the end_dts, will change that now

dfulu:
> So that won't work for the start_dts but yeah that's a nicer way to do the end_dts, will change that now

Oh okay, why won't it work for the start_dts? I ran:

```
pd.to_datetime(["2020-01-01 11:22", "2020-01-01 11:55",]).ceil(pd.Timedelta("1h"))
>> DatetimeIndex(['2020-01-01 12:00:00', '2020-01-01 12:00:00'], dtype='datetime64[ns]', freq=None)
```

Am I missing something?

Sukh-P:
I was missing a .values! Works now

dfulu:
> I was missing a .values! Works now

Cool! Nice work. Sorry to nitpick but to keep it cleaner can we add `.values` to both, otherwise it raises the question of why only have it for one. Right now I think `start_dt` is a `pd.DatetimeIndex` but `end_dt` is a `pd.Series`

dfulu (approved):

Issue #15, opened 2024-08-13
Live: Inference - Make GSP not necessary needed


peterdudfield:
## Detailed Description
Currently GSP dataset is needed, but for live inference we infact dont need this. 
We will need to use a certain t0 in live

## Context
- prepare, this to be used for live

## Possible Implementation
- TODO

peterdudfield:
Im unsure where to just make GSP optional, all make a dummy zarr file with fake gsp results. 

The dummy file feels like we shouldnt need it, but then [adding sun position](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/datasets/pvnet.py#L393) gets hard without gsp.

@dfulu what do you reckon? Maybe the dummy zarr might be easier

dfulu:
I think we should refactor add_sun_position to not require the GSP data. That is planned in #20

dfulu:
I think this should be easier to do once #27 is merged

peterdudfield:
Yea thanks, yea it should be easier. Unfortunately it still needs the GSP config, which i was trying to get rid of in live. Mokcing the GSP zarr would still be an option. 

Other option is to have a `training=true`, but I like the idea of not using this as it runs the risk of the training and inference pipeline being very differnet

peterdudfield:
One other option could be to have a new sun configration, that we use for making sun coordiantes. We could have some deafult values here that would be basically the same forecast minutes as gsp

dfulu:
Yeh probably having a separate entry in the config would be best. Maybe we should go ahead and bring the datapipes config in here now?

peterdudfield:
> Yeh probably having a separate entry in the config would be best. Maybe we should go ahead and bring the datapipes config in here now?

yea why not, can you copy over tests e.t.c too?

Again, it would be great if it was backwards compatiable. I.e if we add sun config, lets add default values to it, so even if an old config doesnt have sun, there will be values

dfulu:
The more issues that come up, the more I think we shouldn't aim for backwards compatibility. In the old version we add the sun position to match the input modality. i.e. to match the datetimes for GSP, or PV, or even wind. So which of these modalities would we make the config default to? As far as I'm aware Windnet does use the solar coords with the wind modality, but PVNet UK regional uses the GSP modality, and I think PVNet Rajasthan uses the PV modality

dfulu:
And also it would be good to be able to change the config to match the parameter naming of #22.

[...]

PR #27, opened 2024-08-14
Refactor sun position calculation

### LINKED ISSUE #20: Refactor `add_sun_position`
This function should not modify the numpy batch if possible. We should find a cleaner way of adding the sun position

dfulu:
This refactors the sun position calculation to avoid needing GSP data to be in the batch. This will help us to run the model live and helps with #15. 

We still wrap the solar coords into a NumpyBatch to help ease the transition of downstream libraries from `ocf_datapipes` to this. This decision can be revisited if we decide to replace the NumpyBatch with something else

Closes #20


peterdudfield (approved):

ocf_data_sampler/torch_datasets/pvnet_uk_regional.py
```diff
@@ -386,12 +390,20 @@ def process_and_combine_datasets(dataset_dict: dict, config: Configuration) -> N
         )
     )
 
+    # Make sun coords NumpyBatch
+    datetimes = pd.date_range(
+        t0-minutes(gsp_config.history_minutes),
+        t0+minutes(gsp_config.forecast_minutes),
+        freq=minutes(gsp_config.time_resolution_minutes),
```
AUdaltsova, line 397: I'm probably missing something, but I thought the idea was to divorce it from gsp overall? Correct me if I'm wrong

dfulu: Yeh after the discussion in #15, I agree its not quite right. I think we need to wait for a resolution to #31 first

PR #37, opened 2024-08-16
Issue/more init

peterdudfield:
# Pull Request

## Description

One more init

## How Has This Been Tested?

Ci tests

- [ ] Yes

## Checklist:

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,7 +1,7 @@
 [bumpversion]
 commit = True
 tag = True
-current_version = 0.0.9
+current_version = 0.0.10
 message = Bump version: {current_version} → {new_version} [skip ci]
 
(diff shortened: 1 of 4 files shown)


dfulu (changes requested):
Looks like you've used `..py` rather than `.py`. 

Separate note, and I could be wrong, but I thought `__init__.py` files weren't necessary after a certain version of python?

peterdudfield:
> Looks like you've used `..py` rather than `.py`.
> 
> Separate note, and I could be wrong, but I thought `__init__.py` files weren't necessary after a certain version of python?

thanks, good spot. 

yea they done seem to be needed when install it with `pip install -e .`, but when deploying and isntall from pypi they do seemed to be needed. Let me see if i can check why

dfulu (approved):
Looks good

PR #40, opened 2024-08-23
remove satellite scaling

dfulu:
Remove the satellite standardisation since it is already normalised


AUdaltsova (approved):

AUdaltsova (changes requested):
probably need to remove from consts as well then?

dfulu:
> probably need to remove from consts as well then?

Well the consts come from `ocf_datapipes`. So I'd just leave it. But we won't port those across to data-sampler

AUdaltsova (approved):

PR #41, opened 2024-08-23
Allow GSP data to be optional

dfulu:
Allows the sampler to not use GSP data if the gsp_zarr_path is `None` or `""`

Should make GSP mocking in openclimatefix/uk-pvnet-app#118 unnecessary


peterdudfield:
Code looks great, is it worth adding a test with no gsp config?

peterdudfield (approved):

PR #43, opened 2024-08-30
add function to return dataArray from Dataset

AUdaltsova:
# Pull Request

## Description

Added a function to return a xr.DataArray from xr.Dataset when loading instead of hardcoding data variable name (e. g. as `return ds.ECMWF_UK`)

## How Has This Been Tested?

Ran the tests locally

- [X] Yes

## Checklist:

### DIFF
--- a/ocf_data_sampler/load/gsp.py
+++ b/ocf_data_sampler/load/gsp.py
@@ -4,8 +4,6 @@ import pkg_resources
 import pandas as pd
 import xarray as xr
 
-from ocf_data_sampler.load.utils import get_xr_data_array_from_xr_dataset
-
 
 def open_gsp(zarr_path: str | Path) -> xr.DataArray:
 
@@ -30,4 +28,4 @@ def open_gsp(zarr_path: str | Path) -> xr.DataArray:
 
     )
 
-    return get_xr_data_array_from_xr_dataset(ds, datavar='generation_mw')
+    return ds.generation_mw
--- a/ocf_data_sampler/load/gsp.py
+++ b/ocf_data_sampler/load/gsp.py
@@ -4,7 +4,7 @@ import pkg_resources
 import pandas as pd
 import xarray as xr
 
-from ocf_data_sampler.load.utils import underlying_array
+from ocf_data_sampler.load.utils import get_xr_data_array_from_xr_dataset
 
 
 def open_gsp(zarr_path: str | Path) -> xr.DataArray:
@@ -30,4 +30,4 @@ def open_gsp(zarr_path: str | Path) -> xr.DataArray:
 
     )
 
-    return underlying_array(ds, datavar='generation_mw')
+    return get_xr_data_array_from_xr_dataset(ds, datavar='generation_mw')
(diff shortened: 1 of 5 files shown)


ocf_data_sampler/load/utils.py
```diff
@@ -26,4 +26,12 @@ def make_spatial_coords_increasing(ds: xr.Dataset, x_coord: str, y_coord: str) -
...
\ No newline at end of file
+    return ds
+
+
+def underlying_array(ds: xr.Dataset, datavar: str | None = None) -> xr.DataArray:
+    if datavar:
+        return ds[datavar]
+    datavars = list(ds.var())
+    assert len(datavars) == 1, "Cannot open as xr.DataArray: dataset contains multiple variables"
+    return ds[datavars[0]]
```
peterdudfield, line 37: should this function be called change_xr_dataset_to_xr_data_array, or something like this?

ocf_data_sampler/load/satellite.py
```diff
@@ -4,7 +4,11 @@
...
 import xarray as xr
-from ocf_data_sampler.load.utils import check_time_unique_increasing, make_spatial_coords_increasing
+from ocf_data_sampler.load.utils import (
+    check_time_unique_increasing,
+    make_spatial_coords_increasing,
+    get_xr_data_array_from_xr_dataset
+)
 
 
 def _get_single_sat_data(zarr_path: Path | str) -> xr.Dataset:
```
peterdudfield, line 14: is this return type hint right?

ocf_data_sampler/load/gsp.py
```diff
@@ -28,4 +30,4 @@ def open_gsp(zarr_path: str | Path) -> xr.DataArray:
 
     )
 
-    return ds.generation_mw
+    return get_xr_data_array_from_xr_dataset(ds, datavar='generation_mw')
```
dfulu, line 33: I think we shouldn't use this function here. For GSP data I think we can assume that we will always call it `generation_mw` and we are hard coding the name here anyway by setting the `datavar` parameter. I'd reverse this change

ocf_data_sampler/load/utils.py
```diff
@@ -27,4 +27,20 @@ def make_spatial_coords_increasing(ds: xr.Dataset, x_coord: str, y_coord: str) -
...
+def get_xr_data_array_from_xr_dataset(ds: xr.Dataset, datavar: str | None = None) -> xr.DataArray:
+    """Return underlying xr.DataArray from passed xr.Dataset. If davar provided, returns specified variable. 
+    If datavar not provided, checks only one variable is present and returns it as an xr.DataArray.
+
+    Args:
+        ds: xr.Dataset to extract xr.DataArray from
+        datavar: (if provided) return specific variable from ds
+    """
+    
+    if datavar:
```
dfulu, line 42: I think we don't need `datavar` as a parameter. In the places where we are hard coding the variable we should just use `ds.varname` instead of going through this function

dfulu (changes requested):

dfulu:
Looks good, but I think we should simplify the function to remove the parameter which specifies the data variable name

ocf_data_sampler/load/utils.py
AUdaltsova: Good point, I'll reverse this

dfulu (approved):
Looks good!

Issue #22, opened 2024-08-13
Use `interval_[start/end]` exclusively


dfulu:
Get rid of history and forecast and move to only support `interval_[start/end]`

AUdaltsova:
this should then go into a separate config mixin along with time resolution, and validation of intervals' divisibility by time resolution should be moved there too

Issue #18, opened 2024-08-13
NWP dropout -> delay


dfulu:
Rephrase the NWP dropout functionality to "delay" and remove the fraction of dropout as we always use some delay

AUdaltsova:
There might be a way to do non-breaking renaming through pydantic functionality

Issue #7, opened 2024-08-02
TODO list


peterdudfield:
The aim is to move away from torch.datapipes which is in ocf_datapipes. 
We want to build a pipline for pvnet which would be ready for training and ready to deployed in the app

Things we will keep the same
- config
- BatchML strcuture

Things to do
- [x] #8
- [ ] Tidy up and refactor
- [x] Use Import ocf_datapipe, until its ready
- [x] CI, pytest
- [x] CI release
- [x] Update pvnet_app to use it - https://github.com/openclimatefix/uk-pvnet-app/pull/118
- [ ] Minimum documentation, example of batch in torch.dataloader
- [x] Make script than can save batches
- [x] Make sure pvnet can be trained
- [ ] Review, Sukh, Peter, Alex
- [x] Retrain PVnet to use ocf_datapipes
- [x] update pvnet to not use ocf_datapipes,

For later
- [x] remove ocf_datapipes from this repo
- [ ] Add pv functions
- [ ] https://github.com/openclimatefix/PVNet-summation/issues/26
- [ ] pvnet_sites, windnet
- [ ] refactor function, try for now to keep function how they are.

dfulu:
@peterdudfield I'm going to close this. I think its deprecated by #98

PR #45, opened 2024-08-30
add config model and corresponding tests

AUdaltsova:
# Pull Request

## Description

Copied over the relevant parts of the config model from ocf_datapipes (pv, nwp, gsp, satellite) and corresponding tests + data for them.

- There is quite a lot I had to strip out that is not relevant to what the code uses right now. Is there anything we want to add back before merging?
- Left several ToDo's for places I wasn't sure were still relevant but left untouched

## How Has This Been Tested?

Ran the test pack

- [X] Yes

## Checklist:


dfulu:
I think there is still a lot of stuff we could strip out of the config model. I think we should take it down to the absolute simplest it can be. I think its better to add in things as we need them rather than port everything across just because we have it

peterdudfield:
I wonder weather we should try to train a model first before bring in stuff from ocf_datapipes. 
This is what we I thought agreed here - https://github.com/openclimatefix/ocf-data-sampler/issues/7

AUdaltsova:
> I think there is still a lot of stuff we could strip out of the config model. I think we should take it down to the absolute simplest it can be. I think its better to add in things as we need them rather than port everything across just because we have it

Yeah I agree. I *think* I stripped out all mixins and properties we don't use minus [sequence length properties](https://github.com/openclimatefix/ocf-data-sampler/blob/b8648cc54837d80fa9bb11c2e8a0241eadbb0ac2/ocf_data_sampler/config/model.py#L91) (i think it's safe to strip out), [system dropout](https://github.com/openclimatefix/ocf-data-sampler/blob/b8648cc54837d80fa9bb11c2e8a0241eadbb0ac2/ocf_data_sampler/config/model.py#L150) (also at least not used in here, but correct me if I'm wrong), and [this change of filename thing](https://github.com/openclimatefix/ocf-data-sampler/blob/b8648cc54837d80fa9bb11c2e8a0241eadbb0ac2/ocf_data_sampler/config/model.py#L311) which I'm not sure if it's still in use. Unless you mean removing arguments inside the mixins, which is probably good to do but I wouldn't be sure what to remove (I guess I can just strip everything that tests/test_data/pvnet_test_config.yaml doesn't use?)

AUdaltsova:
@peterdudfield 
> I wonder weather we should try to train a model first before bring in stuff from ocf_datapipes. This is what we I thought agreed here - #7

Yes I agree that we want to avoid training a model on a different config to limit the searchspace for bugs, but if it's an exact copy I don't think there is any harm in bringing it over (though I guess there's always a risk of human error, especially since we're stripping stuff out... up for debate). I was thinking we can merge a copy and agree not to merge any changes to it before the model is trained, because depending on how long that takes not having a baseline for config here might block development of other stuff?

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+class Git(Base):
+    """Git model"""
+
+    hash: str = Field(
+        ..., description="The git hash of nowcasting_dataset when a dataset is created."
+    )
+    message: str = Field(..., description="The git message for when a dataset is created.")
+    committed_date: datetime = Field(
+        ..., description="The git datestamp for when a dataset is created."
+    )
```
peterdudfield, line 66: strip out

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+        ge=0,
+        description="how many historic minutes to use. "
+        "If set to None, the value is defaulted to InputData.default_history_minutes",
+    )
+
+    log_level: str = Field(
+        "DEBUG",
+        description="The logging level for this data source. "
+        "This is the default value and can be set in each data source",
+    )
```
peterdudfield, line 89: strip out

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+
+    @property
+    def history_seq_length_30_minutes(self):
+        """How many historical steps are there in 30 minute datasets"""
+        return int(np.ceil(self.history_minutes / 30))
+
+    @property
+    def history_seq_length_60_minutes(self):
+        """How many historical steps are there in 60 minute datasets"""
+        return int(np.ceil(self.history_minutes / 60))
```
peterdudfield, line 120: remove property

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+
+
+# noinspection PyMethodParameters
+class DropoutMixin(Base):
```
peterdudfield, line 124: remove it

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+    @field_validator("dropout_fraction")
+    def dropout_fraction_valid(cls, v: float) -> float:
+        """Validate 'dropout_fraction'"""
+        assert 0 <= v <= 1
+        return v
+
+
+# Todo: do we ever use this?
+# noinspection PyMethodParameters
+class SystemDropoutMixin(Base):
```
peterdudfield, line 152: remove out

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+        return self
+
+
+# noinspection PyMethodParameters
+class TimeResolutionMixin(Base):
+    """Time resolution mix in"""
+
+    time_resolution_minutes: int = Field(
+        5,
+        description="The temporal resolution (in minutes) of the satellite images."
```
peterdudfield, line 186: update

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+        """Validate 'label'"""
+        if v not in providers:
+            message = f"provider {v} not in {providers}"
+            logger.error(message)
+            raise Exception(message)
+        return v
+
+
+# noinspection PyMethodParameters
+class PV(
```
peterdudfield, line 262: strip out

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+            logger.error(message)
+            raise Exception(message)
+        return v
+
+
+class Satellite(DataSourceMixin, TimeResolutionMixin, DropoutMixin):
+    """Satellite configuration model"""
+
+    satellite_zarr_path: Union[str, tuple[str], list[str]] = Field(
+        "gs://solar-pv-nowcasting-data/satellite/EUMETSAT/SEVIRI_RSS/OSGB36/all_zarr_int16_single_timestep.zarr",  # noqa: E501
```
peterdudfield, line 354: remove default

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+        IMAGE_SIZE_PIXELS_FIELD.default // 3,
+        description="The number of pixels of the height of the region of interest"
+        " for non-HRV satellite channels.",
+    )
+    satellite_image_size_pixels_width: int = Field(
+        IMAGE_SIZE_PIXELS_FIELD.default // 3,
+        description="The number of pixels of the width of the region "
+        "of interest for non-HRV satellite channels.",
+    )
+    satellite_meters_per_pixel: int = Field(
```
peterdudfield, line 370: TODO rename variables in seperate PR, remove satellite

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+    )
+    satellite_channels: tuple = Field(
+        RSS_VARIABLE_NAMES[1:], description="the satellite channels that are used"
+    )
+    satellite_image_size_pixels_height: int = Field(
+        IMAGE_SIZE_PIXELS_FIELD.default // 3,
+        description="The number of pixels of the height of the region of interest"
+        " for non-HRV satellite channels.",
+    )
+    satellite_image_size_pixels_width: int = Field(
```
peterdudfield, line 365: change to just use `satellite_image_size_pixels_height` or seem if it works with out one

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+        IMAGE_SIZE_PIXELS_FIELD.default // 3,
+        description="The number of pixels of the width of the region "
+        "of interest for non-HRV satellite channels.",
+    )
+    satellite_meters_per_pixel: int = Field(
+        METERS_PER_PIXEL_FIELD.default * 3,
+        description="The number of meters per pixel for non-HRV satellite channels.",
+    )
+
+    is_live: bool = Field(
```
peterdudfield, line 375: remove

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+    )
+    nwp_channels: tuple = Field(
+        NWP_VARIABLE_NAMES["ukv"], description="the channels used in the nwp data"
+    )
+    nwp_accum_channels: tuple = Field([], description="the nwp channels which need to be diffed")
+    nwp_image_size_pixels_height: int = IMAGE_SIZE_PIXELS_FIELD
+    nwp_image_size_pixels_width: int = IMAGE_SIZE_PIXELS_FIELD
+    nwp_meters_per_pixel: int = METERS_PER_PIXEL_FIELD
+    nwp_provider: str = Field("ukv", description="The provider of the NWP data")
+    index_by_id: bool = Field(
```
peterdudfield, line 409: remove

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+    max_staleness_minutes: Optional[int] = Field(
+        None,
+        description="Sets a limit on how stale an NWP init time is allowed to be whilst still being"
+        " used to construct an example. If set to None, then the max staleness is set according to"
+        " the maximum forecast horizon of the NWP and the requested forecast length.",
+    )
+
+    coarsen_to_degrees: Optional[float] = Field(
+        0.1, description="The number of degrees to coarsen the NWP data to"
+    )
```
peterdudfield, line 422: remove

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+
+    def items(self):
+        """Returns dictionary-like items"""
+        return self.root.items()
+
+
+# noinspection PyMethodParameters
+class GSP(DataSourceMixin, TimeResolutionMixin, DropoutMixin):
+    """GSP configuration model"""
+
```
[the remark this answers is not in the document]
peterdudfield: and rm height and width

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+    nwp: Optional[MultiNWP] = None
+    gsp: Optional[GSP] = None
+
+    default_forecast_minutes: int = Field(
+        60,
+        ge=0,
+        description="how many minutes to forecast in the future. "
+        "This sets the default for all the data sources if they are not set.",
+    )
+    default_history_minutes: int = Field(
```
peterdudfield, line 540: remove

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+        "This sets the default for all the data sources if they are not set.",
+    )
+
+    @property
+    def default_seq_length_5_minutes(self):
+        """How many steps are there in 5 minute datasets"""
+        return int((self.default_history_minutes + self.default_forecast_minutes) / 5 + 1)
+
+    @model_validator(mode="after")
+    def set_forecast_and_history_minutes(self):
```
peterdudfield, line 553: remove

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+                if self.nwp[k].forecast_minutes is None:
+                    self.nwp[k].forecast_minutes = self.default_forecast_minutes
+
+                if self.nwp[k].history_minutes is None:
+                    self.nwp[k].history_minutes = self.default_history_minutes
+
+        return self
+
+    @classmethod
+    def set_all_to_defaults(cls):
```
peterdudfield, line 591: remove

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+        for cls_and_attr_name in path_attrs:
+            cls_name, attribute = cls_and_attr_name.split(".")
+            cls = getattr(self.input_data, cls_name)
+            path = getattr(getattr(self.input_data, cls_name), attribute)
+            path = base_path / path
+            setattr(cls, attribute, path)
+            setattr(self.input_data, cls_name, cls)
+
+
+def set_git_commit(configuration: Configuration):
```
peterdudfield, line 630: remove

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+            gsp=GSP(),
+        )
+
+
+class Configuration(Base):
+    """Configuration model for the dataset"""
+
+    general: General = General()
+    input_data: InputData = InputData()
+    git: Optional[Git] = None
```
peterdudfield, line 609: for a later PR, bring input data up a level

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,652 @@
...
+
+    def set_base_path(self, base_path: str):
```
peterdudfield, line 611: try to remove

ocf_data_sampler/config/README.md
```diff
@@ -0,0 +1,21 @@
+# Configuration
```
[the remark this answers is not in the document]
dfulu: I still think this should go

ocf_data_sampler/config/load.py
```diff
@@ -0,0 +1,32 @@
...
+import fsspec
+from pathy import Pathy
+from pyaml_env import parse_config
+
+from ocf_data_sampler.config.model import Configuration
+
+logger = logging.getLogger(__name__)
+
+
+def load_yaml_configuration(filename: Union[str, Pathy]) -> Configuration:
```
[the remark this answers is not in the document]
dfulu: I still think this should be done

ocf_data_sampler/config/load.py
```diff
@@ -0,0 +1,32 @@
...
+import logging
+from typing import Union
+
+import fsspec
+from pathy import Pathy
+from pyaml_env import parse_config
+
+from ocf_data_sampler.config.model import Configuration
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
dfulu: still agree

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,236 @@
...
+    description: str = Field(
+        "example configuration", description="Description of this configuration file"
+    )
+
+
+class DataSourceMixin(Base):
+    """Mixin class, to add forecast and history minutes"""
+
+    forecast_minutes: int = Field(
+        None,
```
dfulu, line 43: I think this should be a required field and so should be `...` instead of `None`?

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,236 @@
...
+        ge=0,
+        description="how many minutes to forecast in the future. "
+        "If set to None, the value is defaulted to InputData.default_forecast_minutes",
+    )
+    history_minutes: int = Field(
+        None,
```
dfulu, line 49: I think this should be a required field and so should be `...` instead of `None`?

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,236 @@
...
+    )
+
+    satellite_image_size_pixels_width: int = Field(
+        ...,
+        description="The number of pixels of the width of the region "
+        "of interest for non-HRV satellite channels.",
+    )
+
+    live_delay_minutes: int = Field(
+        30, description="The expected delay in minutes of the satellite data"
```
dfulu, line 117: I think this should not have a default value and should be required

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,236 @@
...
+    """NWP configuration model"""
+
+    nwp_zarr_path: Union[str, tuple[str], list[str]] = Field(
+        ...,
+        description="The path which holds the NWP zarr",
+    )
+    nwp_channels: tuple = Field(
+        ..., description="the channels used in the nwp data"
+    )
+    nwp_accum_channels: tuple = Field([], description="the nwp channels which need to be diffed")
```
dfulu, line 132: The type of this should be `list[str]` rather than `tuple`

ocf_data_sampler/config/model.py
```diff
@@ -0,0 +1,236 @@
...
+        """Returns dictionary-like items"""
+        return self.root.items()
+
+
+# noinspection PyMethodParameters
+class GSP(DataSourceMixin, TimeResolutionMixin, DropoutMixin):
+    """GSP configuration model"""
+
+    gsp_zarr_path: str = Field(..., description="The path which holds the GSP zarr")
+    gsp_image_size_pixels_height: int = Field(64, description="The size of GSP spacial crop in pixels")
```
dfulu, line 204: These  `gsp_image_size_pixels_*` can go

ocf_data_sampler/config/save.py
```diff
@@ -0,0 +1,35 @@
...
+import yaml
+from pathy import Pathy
+
+from ocf_data_sampler.config.model import Configuration
+
+logger = logging.getLogger(__name__)
+
+
+def save_yaml_configuration(
+    configuration: Configuration, filename: Optional[Union[str, Pathy]] = None
```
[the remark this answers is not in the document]
dfulu: I still agree with this

ocf_data_sampler/config/save.py
```diff
@@ -0,0 +1,35 @@
...
+import logging
+from typing import Optional, Union
+
+import fsspec
+import yaml
+from pathy import Pathy
+
+from ocf_data_sampler.config.model import Configuration
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
dfulu: still agree

tests/config/test_config.py
```diff
@@ -0,0 +1,95 @@
...
+        # save default config to file
+        save_yaml_configuration(Configuration(), filename)
+
+        # check the file can be loaded
+        _ = load_yaml_configuration(filename)
+
+
+def test_yaml_load_env(test_config_filename):
+    """
+    Check a configuration can be loaded with an env var
```
dfulu, line 51: I don't think the functionality to load from environmental variable exists - I also don't think it should. I'd remove this

tests/config/test_config.py
```diff
@@ -0,0 +1,95 @@
...
+
+    with tempfile.NamedTemporaryFile(suffix=".yaml") as fp:
+        filename = fp.name
+
+        # check that temp file can't be loaded
+        with pytest.raises(TypeError):
+            _ = load_yaml_configuration(filename)
+
+        # save default config to file
+        save_yaml_configuration(Configuration(), filename)
```
dfulu, line 43: This is too simple to be a useful test. Could we load the config in `test_yaml_load_test_config()` and test saving that?

tests/config/test_config.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    Check a configuration can be saved to a yaml file
+    """
+
+    with tempfile.NamedTemporaryFile(suffix=".yaml") as fp:
+        filename = fp.name
+
+        # check that temp file can't be loaded
+        with pytest.raises(TypeError):
+            _ = load_yaml_configuration(filename)
```
dfulu, line 40: I'm not sure why we'd want this check. I think we should get rid of it

dfulu (changes requested):
Looks really good but I think there is more we can remove

ocf_data_sampler/config/model.py
AUdaltsova: I think it's optional because of satellite? It's allowed to not have forecast, but yeah we can make it required and set to 0

ocf_data_sampler/config/model.py
AUdaltsova: Oh right it doesn't undergo spacial cropping does it

ocf_data_sampler/config/model.py
peterdudfield: I think `live` was all being stripped out

tests/config/test_config.py
peterdudfield: yea, I think we used to use it, but not any more. It was a nice feature meaing you didnt have to edit the config file with your ENV VARS, but to keep things simple, can strip this out

tests/config/test_config.py
peterdudfield: Good to check the saving works right?

tests/config/test_config.py
peterdudfield: yea, could strip it out, it does just check that we get an error properly when trying to load a file that doesnt have a config in it, but perhaps we can remove it for now, to keep it simple

tests/config/test_config.py
dfulu: Yeh definitely, but the default config has no input sources and so we are essentially saving out `{gsp: None, satellite: None, nwp:none}`. If we instead save out the config which we load in the `test_yaml_load_test_config()` there are some inputs to be saved

tests/config/test_config.py
dfulu: Okay, well since that functionality doesn't exist anymore I think we should get rid of it for now

ocf_data_sampler/config/model.py
dfulu: I see what you mean. I think that since it is required for some sources it should still be required here. I have been generally setting it to zero for the satellite inputs. Hopefully this should be cleaner once we switch to `interval_[start/end]`

ocf_data_sampler/config/model.py
dfulu: It is but not until we switch to `interval_[start/end]`

ocf_data_sampler/config/model.py
dfulu: No it doesn't. We use it similarly to selecting the PV site

ocf_data_sampler/config/model.py
AUdaltsova: this is just how delay/dropout is organised in sat currently, generally yes I think we are moving towards a) calling delay, especially for nwp, what it is (#18) and/or b) removing the need for that altogether with `interval_start/interval_end` (#22). As far as I understand it doesn't actually have anything to do with live as the environment, its live in a sense that this is the delay we expect to have in 'live' circumstances (do correct me if I'm wrong!) 

This is still needed so I think we're keeping it in until we resolve either of the issues above; we can rename it in the meantime but with it being scheduled to be chucked maybe not worth it

AUdaltsova:
Tbh now I'm questioning whether test_config.yaml is needed at all, if anything it has less parameters than pvnet_test_config.yaml. But I dislike the idea of substituting it with PVNet config (even if now they are equivalent) because moving forward it will have to be different and I think it's better to amend specific test_config.yaml than swap them in and then out.

Alternatively, we can make the config tests cycle through example configs for all pipelines? Right now it will just be PVNet but once the others are added there will be more. I think it offers higher fidelity to actual use-cases than just meshing every parameter into one file; but also might be too much of a faff for nothing.

AUdaltsova:
Also, should we make `dropout_timedeltas_mintes` a required field? I think it is useful to require it so it gets thought about, and putting [0] as a way of "turning it off" should provide the same behaviours as passing null currently does. It means that [this check making sure dropout is less than 0](https://github.com/openclimatefix/ocf-data-sampler/blob/d0d82907fbb9c3b080427f44d9438e458d0efc99/ocf_data_sampler/select/dropout.py#L15) will need to be changed, but something has to be done about this anyway bc currently config actually allows for dropout timedeltas to be 0, so one of them needs to be changed. Or we can just delete the check inside dropout since config already checks for that. Do you guys know why it was introduced in the firs place maybe?

dfulu:
> Tbh now I'm questioning whether test_config.yaml is needed at all, if anything it has less parameters than pvnet_test_config.yaml. But I dislike the idea of substituting it with PVNet config (even if now they are equivalent) because moving forward it will have to be different and I think it's better to amend specific test_config.yaml than swap them in and then out.
> 
> Alternatively, we can make the config tests cycle through example configs for all pipelines? Right now it will just be PVNet but once the others are added there will be more. I think it offers higher fidelity to actual use-cases than just meshing every parameter into one file; but also might be too much of a faff for nothing.

I'd leave it for now. At the moment the `Configuration` class is supposed to work for all pipelines and so I think just testing it once with an overcomplete file is good enough. Once the config has been refactored into lots of composable mixins, we might think about having separate pydantic `Configuration`  for each dataset if they really do need to diverge. Trying to force one config to work for mutliple datapipes is what made the config so messy in ocf_datapipes in the first place.

dfulu:
> Also, should we make `dropout_timedeltas_mintes` a required field? I think it is useful to require it so it gets thought about, and putting [0] as a way of "turning it off" should provide the same behaviours as passing null currently does. It means that [this check making sure dropout is less than 0](https://github.com/openclimatefix/ocf-data-sampler/blob/d0d82907fbb9c3b080427f44d9438e458d0efc99/ocf_data_sampler/select/dropout.py#L15) will need to be changed, but something has to be done about this anyway bc currently config actually allows for dropout timedeltas to be 0, so one of them needs to be changed. Or we can just delete the check inside dropout since config already checks for that. Do you guys know why it was introduced in the firs place maybe?

I'm not sure about this, but I also really feel strongly about it. The current UK PVNet on prod doesn't make use of dropout (except for NWP delay which we should distinguish), though some of the models on dev do. So in that sense I think it could be optional. But feel free to change it if you think it makes more sense. However my preference of an "off" setting would be an empty list rather than `[0]` if we don't use `None`

We should definitely at least align those checks! We could delete the check from the function itself if we think it won't be used independently from the config, which has always been true for us. It is a very computationally cheap check though so I'd choose whichever we think is better code quality. @peterdudfield might have some thoughts on this?

There is currently an additional way of turning off dropout which is to set `dropout_fraction=0`. It might actually be good in the config to check that the `dropout_fraction` and `dropout_timedeltas_mintes` are both set to "on" or both set to "off" and not allow one to be "on" and one set to "off" since that could be ambiguous?

dfulu (approved):
This is looking really good now @AUdaltsova. My last comments are only nice-to-have's. You can decide what to do with them. Great work!

AUdaltsova:
> However my preference of an "off" setting would be an empty list rather than [0] if we don't use None

I see what you mean and I would usually agree, but I think the whole concept of dropout being "off" is a bit of an illusion: there isn't really such thing as ambiguous delay, you either set it to something or it defaults to instant availability. Which is an "off" in the sense that it's not delayed, but it's still a very specific scenario that you are committing to, so I think it would be nice to think of that and consciously put 0 instead of thinking oh well, just do whatever it is you do, if that makes sense?

> It is a very computationally cheap check though so I'd choose whichever we think is better code quality. 

Personally, I would go for a config check. This is a restriction on an input parameter and should be checked on input (same as the NWP provider check, which turns out has not been working all this time :) Obviously incorrect providers were being caught down the line when consts are suddenly not there, so it is not even strictly necessary, but I think it should stay and flag this right away instead of half-way through selection)

> There is currently an additional way of turning off dropout which is to set dropout_fraction=0. It might actually be good in the config to check that the dropout_fraction and dropout_timedeltas_mintes are both set to "on" or both set to "off" and not allow one to be "on" and one set to "off" since that could be ambiguous?

That is a very good point! Especially since dropout fraction defaults to "off", so can accidentally negate dropout timedeltas and not fail anywhere (I think)

AUdaltsova:
> We should definitely at least align those checks! We could delete the check from the function itself if we think it won't be used independently from the config, which has always been true for us. It is a very computationally cheap check though so I'd choose whichever we think is better code quality. @peterdudfield might have some thoughts on this?

@peterdudfield thought it's good redundancy in case we change the config or use the fnction without it, so I put it back in but made it a leq to be consistent with config check.

>It might actually be good in the config to check that the dropout_fraction and dropout_timedeltas_mintes are both set to "on" or both set to "off" and not allow one to be "on" and one set to "off" since that could be ambiguous?

Added a check and test for that!

PR #49, opened 2024-09-16
Tweaks for production

dfulu:
# Pull Request

## Description

- Add nan filling
- Make the target GSP location added to batch independent of adding the GSP data
- Eager-load the GSP data - not necessarily for production but I think we should have this in memory since its small enough

## How Has This Been Tested?

- Has been run with PVNet app

## Checklist:

### DIFF
--- a/ocf_data_sampler/torch_datasets/pvnet_uk_regional.py
+++ b/ocf_data_sampler/torch_datasets/pvnet_uk_regional.py
@@ -403,6 +403,7 @@ def process_and_combine_datasets(
         numpy_modalities.append(convert_satellite_to_numpy_batch(da_sat))
 
     gsp_config = config.input_data.gsp
+    
     if "gsp" in dataset_dict:
         da_gsp = concat_xr_time_utc([dataset_dict["gsp"], dataset_dict["gsp_future"]])
         da_gsp = normalize_gsp(da_gsp)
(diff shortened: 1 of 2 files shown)


ocf_data_sampler/torch_datasets/pvnet_uk_regional.py
```diff
@@ -406,9 +425,18 @@ def process_and_combine_datasets(
...
-
-    # Combine all the modalities
+    
+    # Add coordinate data
+    # TODO: Do we need all of these?
+    numpy_modalities.append({
+        BatchKey.gsp_id: location.id,
+        BatchKey.gsp_x_osgb: location.x,
+        BatchKey.gsp_y_osgb: location.y,
+    })
```
peterdudfield, line 435: Id be tempted to keep them, nice to have when exploring a batch

ocf_data_sampler/torch_datasets/pvnet_uk_regional.py
```diff
@@ -423,22 +451,6 @@ def compute(xarray_dict: dict) -> dict:
     return xarray_dict
 
 
-def get_locations(ga_gsp: xr.DataArray) -> list[Location]:
```
peterdudfield, line 426: is this just not needed now? i cant remember what is was needed for?

peterdudfield (approved):

ocf_data_sampler/torch_datasets/pvnet_uk_regional.py
dfulu: No its not used anymore. `get_gsp_locations()` is used instead. I just forgot to delete this when that new function was added

ocf_data_sampler/torch_datasets/pvnet_uk_regional.py
```diff
@@ -344,6 +344,24 @@ def slice_datasets_by_time(
     return sliced_datasets_dict
 
 
+def fill_nans_in_arrays(batch: NumpyBatch) -> NumpyBatch:
```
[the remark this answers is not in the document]
dfulu: I'd leave it in here for now. I think things like this should stay in the dataset file until they are used in multiple datasets and thus prove themselves as being generally useful

PR #50, opened 2024-09-16
Fill nans

dfulu:
Add the option to specify the GSPs for the dataset


ocf_data_sampler/torch_datasets/pvnet_uk_regional.py
```diff
@@ -451,8 +451,12 @@ def compute(xarray_dict: dict) -> dict:
     return xarray_dict
 
 
-def get_gsp_locations() -> list[Location]:
+def get_gsp_locations(gsp_ids: list[int] | None = None) -> list[Location]:
     """Get list of locations of all GSPs"""
+    
+    if gsp_ids is None:
+        gsp_ids = [i for i in range(1, 318)]
```
peterdudfield, line 458: I like to put N_GSP at the top somewhere, just to keep it tidy and not lose hard coded values

PR #57, opened 2024-09-27
copy files over ocf_datapipes

peterdudfield:
# Pull Request

## Description

- Move over location, geospatial files
- tried to trim out as much as possible

Helps with #58 

## How Has This Been Tested?

CI test

- [x] Yes

## Checklist:


ocf_data_sampler/select/geospatial.py
```diff
@@ -0,0 +1,125 @@
...
+    Args:
+        x: osgb east-west
+        y: osgb north-south
+        xr_data: xarray object with geostationary area
+
+    Returns:
+        Geostationary coords: x, y
+    """
+    # Only load these if using geostationary projection
+    import pyproj
```
AUdaltsova, line 78: pyproj is definitely imported and used before this so we can probably remove that

ocf_data_sampler/select/geospatial.py
```diff
@@ -0,0 +1,125 @@
...
+    except KeyError:
+        area_definition_yaml = xr_data.data.attrs["area"]
+    geostationary_area_definition = pyresample.area_config.load_area_from_string(
+        area_definition_yaml
+    )
+    geostationary_crs = geostationary_area_definition.crs
+    osgb_to_geostationary = pyproj.Transformer.from_crs(
+        crs_from=OSGB36, crs_to=geostationary_crs, always_xy=True
+    ).transform
+    return osgb_to_geostationary(xx=x, yy=y)
```
AUdaltsova, line 92: out of curiosity, which datasets use this? I don't think I've seen it before (but maybe it was under the hood all this time!)

ocf_data_sampler/select/geospatial.py
```diff
@@ -0,0 +1,125 @@
...
+        return "lon_lat", "longitude", "latitude"
+    elif "x_geostationary" in available_coords:
+        return "geostationary", "x_geostationary", "y_geostationary"
+    elif "x_osgb" in available_coords:
+        return "osgb", "x_osgb", "y_osgb"
+    else:
+        raise ValueError(f"Unrecognized coordinate system: {available_coords}")
+
+
+def spatial_coord_type(ds: xr.Dataset):
```
AUdaltsova, line 106: the function fails hard if ds passed is not `xr.DataArray`, so the type hint should be different probably

ocf_data_sampler/select/geospatial.py
```diff
@@ -0,0 +1,125 @@
+"""Geospatial functions"""
+
+from datetime import datetime
```
AUdaltsova, line 3: I dont think this import is ever used

ocf_data_sampler/select/geospatial.py
```diff
@@ -0,0 +1,125 @@
...
+from numbers import Number
+from typing import Union
+
+import numpy as np
+import pandas as pd
+import pvlib
```
AUdaltsova, line 9: I think these are never used; I commented them out and none of the tests failed

ocf_data_sampler/select/geospatial.py
peterdudfield: I think it used for selecting spatial slices when starting with osgb locations, and we need to get the geostationary cooridnates (for example satellite). Does this make sense?

AUdaltsova: Yes, thank you!

ocf_data_sampler/select/select_spatial_slice.py
```diff
@@ -97,8 +93,8 @@ def _get_idx_of_pixel_closest_to_poi(
     x_index = da.get_index(x_dim)
     y_index = da.get_index(y_dim)
 
-    closest_x = x_index.get_indexer([x], method="nearest")[0]
-    closest_y = y_index.get_indexer([y], method="nearest")[0]
+    closest_x = float(x_index.get_indexer([x], method="nearest")[0])
+    closest_y = float(y_index.get_indexer([y], method="nearest")[0])
```
AUdaltsova, line 97: why is this needed now?

ocf_data_sampler/select/select_spatial_slice.py
```diff
@@ -45,9 +44,6 @@ def convert_coords_to_match_xarray(
         if from_coords == "osgb":
             x, y = osgb_to_geostationary_area_coords(x, y, da)
 
-        elif from_coords == "lon_lat":
-            x, y = lon_lat_to_geostationary_area_coords(x, y, da)
-
```
AUdaltsova, line 50: potentially silly question, but isn't this needed for satellite data and lat-lon poi? or do we not have that in pvnet => chuck it for now?

ocf_data_sampler/select/select_spatial_slice.py
```diff
@@ -130,13 +126,8 @@ def _get_idx_of_pixel_closest_to_poi_geostationary(
...
     # Get the index into x and y nearest to x_center_geostationary and y_center_geostationary:
-    x_index_at_center = searchsorted(
-        da[x_dim].values, center_geostationary.x, assume_ascending=True
-    )
-
-    y_index_at_center = searchsorted(
-        da[y_dim].values, center_geostationary.y, assume_ascending=True
-    )
+    x_index_at_center = np.searchsorted(da[x_dim].values, center_geostationary.x)
+    y_index_at_center = np.searchsorted(da[y_dim].values, center_geostationary.y)
```
AUdaltsova, line 130: oh that's nice!

AUdaltsova:
very much outside the scope of this pr and also I might be misreading this, but I think we never actually use spatial padding? The option for it defaults to false and cannot be controlled from the config I don't think... gonna go make an issue about this

ocf_data_sampler/select/geospatial.py
dfulu: Yeh to confirm we use this in UK PVNet where the GSPs (if I remember right) have OSGB coords, but we need to slice from the satellite data so we convert the coords on the fly

ocf_data_sampler/select/select_spatial_slice.py
dfulu: We don't use this in UK PVNet and we haven't brought across the `lon_lat_to_geostationary_area_coords()` function from ocf_datapipes yet. So maybe it is more in keeping with our minimalist approach for now

ocf_data_sampler/select/geospatial.py
```diff
@@ -0,0 +1,121 @@
...
+
+    Return: 2-tuple of OSGB x, y
+    """
+    return _lon_lat_to_osgb(xx=x, yy=y)
+
+
+def osgb_to_geostationary_area_coords(
+    x: Union[Number, np.ndarray],
+    y: Union[Number, np.ndarray],
+    xr_data: Union[xr.Dataset, xr.DataArray],
```
dfulu, line 62: We are limiting ourselves to only use `xr.DataArray` in data-sampler so we should update this type hint for that

ocf_data_sampler/select/geospatial.py
```diff
@@ -0,0 +1,121 @@
...
+    Returns:
+        Geostationary coords: x, y
+    """
+    # Only load these if using geostationary projection
+    import pyresample
+
+    try:
+        area_definition_yaml = xr_data.attrs["area"]
+    except KeyError:
+        area_definition_yaml = xr_data.data.attrs["area"]
```
dfulu, line 80: Since we are limiting to DataArrays we don't need this try-except block. The area string should always be on `xr_data.attrs["area"]`

dfulu (changes requested):
There are a couple of small things I think we should clean up, but otherwise looks good!

ocf_data_sampler/select/geospatial.py
peterdudfield: Thanks, yea, good to trim it down

ocf_data_sampler/select/geospatial.py
peterdudfield: Thanks, have done

AUdaltsova (approved):

peterdudfield:
> very much outside the scope of this pr and also I might be misreading this, but I think we never actually use spatial padding? The option for it defaults to false and cannot be controlled from the config I don't think... gonna go make an issue about this
> 
> UPD: #62

THanks @AUdaltsova and good to get this in another issue. 

@dfulu let me know if your happy for me to merge

ocf_data_sampler/select/select_spatial_slice.py
dfulu: @peterdudfield is this required?

peterdudfield: i got some pydnatic error before, but I can try removing it again

peterdudfield: i removed it

ocf_data_sampler/select/geospatial.py
```diff
@@ -0,0 +1,121 @@
...
+WGS84 = 4326
+
+
+_osgb_to_lon_lat = pyproj.Transformer.from_crs(
+    crs_from=OSGB36, crs_to=WGS84, always_xy=True
+).transform
+_lon_lat_to_osgb = pyproj.Transformer.from_crs(
+    crs_from=WGS84, crs_to=OSGB36, always_xy=True
+).transform
+_geod = pyproj.Geod(ellps="WGS84")
```
dfulu, line 28: I don't think this is used anywhere. If so can we strip it out?

dfulu (approved):
Looks good. Thanks for doing this

PR #61, opened 2024-09-27
Config refactor

### LINKED ISSUE #59: remove source prefixes from config parameters
## Detailed Description

Currently we have things like `pixel_width` or `channels` prefixed by source in config ( e g `satellite_image_size_pixels_height`, `nwp_channels`). This should be refactored to be the same across sources and potentially moved to a separate mixin 

## Context
This will make referencing easier and we can then make function calls for different data sources more uniform, leading to cleaner code

AUdaltsova:
# Pull Request

## Description

- Created a base data source config with path, image pixel sizes, and time and dropout mixins. 
- Moved forecast and history minutes into time mixin fo easier validation. 
- Refactored to remove data source prefixes

Fixes #59 

## How Has This Been Tested?

ran the tests locally

- [X] Yes

## Checklist:


AUdaltsova:
his should be merged after #45 is merged

AUdaltsova:
Interval start/end will ned to be added but other than that ready for review

ocf_data_sampler/config/model.py
```diff
@@ -208,27 +199,10 @@ def items(self):
         return self.root.items()
 
 
-# noinspection PyMethodParameters
-class GSP(DataSourceMixin, TimeResolutionMixin, DropoutMixin):
+class GSP(TimeWindowMixin, DropoutMixin):
```
peterdudfield, line 202: Should this be a DataSourceBase?

peterdudfield (changes requested):
Ive added one small comment, but in general I think this is a nice PR, makes things a bit simplier

ocf_data_sampler/config/model.py
peterdudfield: GSP(DataSourceMixin), does that work? Or perhaps there is a reason for it to be like it is. I may have missed something

AUdaltsova: I see what you mean - GFS doesn't have crop size parameters, which are required in DataSourceMixin. I wouldn't want to make them optional, and since GFS only needs zarr path from there, I thought this was better than dissolving DataSourceMixin/leaving only crop parameters in it as that means zarr path will have to be added individually to every source. But this might not be the best solution either, I'm open to suggestions here.

ocf_data_sampler/config/model.py
```diff
@@ -93,36 +77,66 @@ def dropout_instructions_consistent(self) -> Self:
...
-    satellite_image_size_pixels_width: int = Field(
+    image_size_pixels_width: int = Field(
         ...,
-        description="The number of pixels of the width of the region "
-        "of interest for non-HRV satellite channels.",
+        description="The number of pixels of the width of the region of interest",
+    )
+
+
+class Satellite(DataSourceBase):
```
AUdaltsova, line 135:
```suggestion
class Satellite(DataSourceBaseMixin):
```

ocf_data_sampler/config/model.py
```diff
@@ -131,21 +145,16 @@ class Satellite(DataSourceMixin, TimeResolutionMixin, DropoutMixin):
 
 
 # noinspection PyMethodParameters
-class NWP(DataSourceMixin, TimeResolutionMixin, DropoutMixin):
+class NWP(DataSourceBase):
```
AUdaltsova, line 148:
```suggestion
class NWP(DataSourceBaseMixin):
```

ocf_data_sampler/config/model.py
AUdaltsova: To be honest I'm not wild about how DataSourceMixin is grouped right now - it has both zarr path and spacial cropping which aren't really related, but we have to add zarr path to everything so I wanted to put it into a mixin, and a mixin for zarr path alone would increase unnecessary code. But this is clearly not a universal solution so maybe I shouldn't have done that

peterdudfield: I see `DataSourceMixin` has some spacial element to it too. Perhaps just leave it how it is for the moment with `TimeWindowMixin` and `DropoutMixin`. 

You could have `ImageDataSourceMixin` which inherits from `DataSourceMixin` and has spacial cropping elements. Just an idea, 

Potentially it might be simplier leaving it how it is right now in the PR

AUdaltsova: Yeah good point, done that now, thanks! I've added `zarr_path` back into each config for now, and I think that's for the best. We can add another mixin just for that but I don't think there's a point in doing it, and also this way it's clearer which parameters each one has without looking up mixins.

PR #64, opened 2024-10-07
switch to interval_[start/end] + allow forecast non-zero first step

### LINKED ISSUE #22: Use `interval_[start/end]` exclusively

dfulu:
In this PR we:

- Update the config and function parameters to use `interval_[start/end]`
- Allow the NWP to have a first forecast step which is not valid at the init-time
- Slight tidying

Closes #22

### DIFF
--- a/ocf_data_sampler/config/model.py
+++ b/ocf_data_sampler/config/model.py
@@ -15,6 +15,7 @@ from typing import Dict, List, Optional
 from typing_extensions import Self
 
 from pydantic import BaseModel, Field, RootModel, field_validator, ValidationInfo, model_validator
+
 from ocf_data_sampler.constants import NWP_PROVIDERS
 
 logger = logging.getLogger(__name__)
@@ -34,7 +35,7 @@ class Base(BaseModel):
 class General(Base):
     """General pydantic model"""
 
-    name: str = Field("example", description="The name of this configuration file.")
+    name: str = Field("example", description="The name of this configuration file")
     description: str = Field(
         "example configuration", description="Description of this configuration file"
     )
@@ -48,17 +49,17 @@ class TimeWindowMixin(Base):
         gt=0,
         description="The temporal resolution of the data in minutes",
     )
-
-    interval_end_minutes: int = Field(
-        ...,
-        description="Data interval ends at `t0 + interval_end_minutes`",
-    )
+    
     interval_start_minutes: int = Field(
         ...,
         description="Data interval starts at `t0 + interval_start_minutes`",
     )
 
-
+    interval_end_minutes: int = Field(
+        ...,
+        description="Data interval ends at `t0 + interval_end_minutes`",
+    )
+    
     @model_validator(mode='after')
     def check_interval_range(cls, values):
         if values.interval_start_minutes > values.interval_end_minutes:
@@ -78,6 +79,7 @@ class TimeWindowMixin(Base):
         return v
 
 
+
 # noinspection PyMethodParameters
 class DropoutMixin(Base):
     """Mixin class, to add dropout minutes"""
@@ -89,7 +91,7 @@ class DropoutMixin(Base):
     )
 
     dropout_fraction: float = Field(
-        default=0, 
+        default=0,
         description="Chance of dropout being applied to each sample",
         ge=0,
         le=1,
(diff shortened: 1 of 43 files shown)


ocf_data_sampler/config/model.py
```diff
@@ -40,21 +40,44 @@ class General(Base):
...
+    interval_start_minutes: int = Field(
+        ...,
+        description="Data interval starts at `t0 + interval_start_minutes`",
     )
 
 
+    @model_validator(mode='after')
+    def check_interval_range(cls, values):
+        if values.interval_start_minutes > values.interval_end_minutes:
+            raise ValueError('interval_start_minutes must be <= interval_end_minutes')
```
AUdaltsova, line 65: Nitpick, but the error message doesn't say what the test doe: one allows for equality, one doesn't

ocf_data_sampler/select/select_time_slice.py
```diff
@@ -39,23 +39,15 @@ def _sel_fillinterp(
...
     sample_period_duration: pd.Timedelta,
-    history_duration: pd.Timedelta | None = None,
-    forecast_duration: pd.Timedelta | None = None,
-    interval_start: pd.Timedelta | None = None,
-    interval_end: pd.Timedelta | None = None,
     fill_selection: bool = False,
     max_steps_gap: int = 0,
 ):
     """Select a time slice from a Dataset or DataArray."""
-    used_duration = history_duration is not None and forecast_duration is not None
```
AUdaltsova, line 51: I don't think this is used after assignment, so can be an assert?

ocf_data_sampler/torch_datasets/pvnet_uk_regional.py
```diff
@@ -279,8 +283,8 @@ def slice_datasets_by_time(
                 da_nwp,
                 t0,
                 sample_period_duration=minutes(nwp_config.time_resolution_minutes),
-                history_duration=minutes(nwp_config.history_minutes),
-                forecast_duration=minutes(nwp_config.forecast_minutes),
+                interval_start=minutes(nwp_config.interval_start_minutes),
+                interval_end=minutes(nwp_config.interval_end_minutes),
                 dropout_timedeltas=minutes(nwp_config.dropout_timedeltas_minutes),
                 dropout_frac=nwp_config.dropout_fraction,
```
AUdaltsova, line 289: Looking at the way sat delay is handled now, do we ever actually use the dropout fraction system for NWPs (or at all)? Because if we always just set it to one delay that is always used, then we can just do it through interval_end as well. 

Though to be honest I am a bit on the fence about factoring the delay into interval start/end, I think it can get messy and is very susceptible to human error. I like the way current NWP dropout system lets you say "I want x amount backward and y amount forward, latest you can get me" and it factors in the delay for you. When setting up sat config it used to be "I want an hour of history, but the delay is 30 minutes, so actually I need to request 90 minutes of history" etc, which is bad, and moving to "I want an hour of history with delay 30 min, so I need everything between -90 min and -30 min" is better, but still a long way away from "I want this much, you'll probably need to wait that much for it to come in, go figure it out".

What am I missing?

ocf_data_sampler/select/select_time_slice.py
AUdaltsova: This has wrong code context selected, it's referring to line 52 `used_intervals = interval_start is not None and interval_end is not None`

ocf_data_sampler/torch_datasets/pvnet_uk_regional.py
dfulu: No we don't use the dropout fraction for NWPs, you're right. But I think I'd tag that on to #18. Maybe we can make that issue to generally reconsider the delay in NWPs. 

> I like the way current NWP dropout system lets you say "I want x amount backward and y amount forward, latest you can get me" and it factors in the delay for you. 

I think that hasn't really changed in this PR. We are basically just renaming the history and forecast parameters. So after this PR it is "I want between [t0+x, and t0+y], the latest init-time you can get me". So for NWP it does still factor in the delay for you.

I think this PR gives a cleaner way of expressing the time slice. For the manchester prize satellite prediction input I have wanted to select between t0+15 minutes and t0+3 hours. Under our old system I's need to say I wanted negative 15 minutes of history which seems clumsy.

> When setting up sat config it used to be "I want an hour of history, but the delay is 30 minutes, so actually I need to request 90 minutes of history" etc, which is bad, and moving to "I want an hour of history with delay 30 min, so I need everything between -90 min and -30 min" is better, but still a long way away from "I want this much, you'll probably need to wait that much for it to come in, go figure it out".

Personally I like the explicitness of setting everything relative to t0. To me it better highlights the data we expect to be available in production -> i.e. we can't delete production data until it is more than 90 minutes stale. But I'm open to suggestions for how else we might parameterise this

AUdaltsova: > You're right we always set the NWP dropout to a constant value. I'm not sure how we would use `interval_end` for that though. Currently `interval_start/end` define the slice along the `step` dimension. `t0` and `dropout_timedeltas_minutes` define the slice along the `init-time` dimension.

Yeah good point, my bad!

>I think this PR gives a cleaner way of expressing the time slice. For the manchester prize satellite prediction input I want to slice between t0+15 minutes and t0+3 hours. Under our old system I'd need to say I wanted negative 15 minutes of history which seems clumsy

I think this is the use case I was missing, it makes a lot of sense now, thanks!

Not a hill I will die on, but for what it's worth: I think this intervaling is convenient, especially if you want an "unconventional" sat slice (might even enable 1 image of sat history that I previously had to hack around to get with forecast/history, which is great!) but also because for sat data `init_time` and `step` are the same thing, it kind of mixes the two together if that makes sense? Which is why it doesn't work for NWP delay setting, and why I initially thought it would. In that sense, I would maybe prefer for the delay to still be set separately and functioning the way NWP delay does and not how it was previously, so to be able to set interval (-60, -15) and delay -30 (btw, hot take but delay already implies moving back, so maybe shouldn't be set in negative? Seems kind of counter-intuitive to me. But there's probably a good reason and also, I digress), instead of interval (-90, -45), or how it was previously, history 90 delay 45 under the assumption that the tail will get cut off. 

Main reason: I know at least I am extremely prone to human error in config setting and would like as much automation as possible, and also prefer more single-function things and less multi-function things. 

I know we've talked a lot about eradicating live sat delay, and maybe this is a bit excessive parametrising for what it does, but to me it seems a bit more straightforward on user end. Also, feel free to tell me what I missed! Fully expect to be wrong on this one, I have way less experience with sat than you do.

>Personally I like the explicitness of setting everything relative to t0. To me it better highlights the data we expect to be available in production -> i.e. we can't delete production data until it is more than 90 minutes stale. But I'm open to suggestions for how else we might parameterise this

Here you'd 100% know better than I do! I get that this is a bit of a have your cake and eat it thing and if your prod cake is more useful than my user cake I'm happy for intervaling to cover delay as well.

By the way, thanks a lot for doing this! I really like the way it's turning out.

AUdaltsova: I also realised that sat config inherits from DropoutMixin (and always has been), so maybe that's the solution? I know we were going to change that to simplify into just delay, so unless there's something I'm missing I think sat and nwp can use the same mechanism here? 

Though dropout is also inherited by GSP and Site configs, and I'm not sure if there are use cases for those that will not be covered by delay.

dfulu: 1. **Switching to (-60, -15) and delay -30 instead of the current  (-90, -45)**

Personally I see that as slightly more complicated that it needs to be and I prefer having one less parameter. In the case you suggest  (-60, -15) and delay -30  is an identical slice to (-45, 0) and delay -45. I think it is messy that we could get the same time slice with different parameter settings. That feels like something which could be prone to human error too

2. **About the use of dropout and delay**

I'm not sure if I'm answering your question here, but I think the dropout and delay should be different things. Delay controls the time slice we are choosing and therefore the shape of the input tensor. Dropout controls whether those requested datetimes are actually available or are infilled with NaNs in the array. Admittedly we abuse the "dropout" in the NWP data but I think we should change that to make it clearer.

I think we should get rid of "delay" as we have it currently, and just use the intervals relative to t0. We can then discuss how we name "dropout" across different data sources

AUdaltsova: Hello again! 
1. I see what you're saying! I guess I just dislike not having the duration of the slice explicitly stated anywhere. Also, this is beside the point probably, but I think it's fine to have different parameter settings to give you the same slice: if I'm reading someone else's config and it says (-90, -45) I'd assume it's delay 45, slice of 45, and if it's actually delay 30 and we don't want to use the newest 15 min for some reason I'd want to know (don't think this would ever happen but I hope you can see what I'm getting at anyway; sorry if it's not too clear!). 
Honestly, I think this conversation doesn't mean we can't merge this, I think intervaling is a good update regardless and we can always redo delay later if we ever want to.

AUdaltsova (approved):

ocf_data_sampler/torch_datasets/pvnet_uk_regional.py
dfulu: Thanks! We can come back to this again if we want to change it and I'll add it to the agenda of our next data-sampler meeting

PR #68, opened 2024-10-18
Rm ocf datapipes

### LINKED ISSUE #58: remove ocf datapipes
Remove ocf datapipes
We should copy in and then strip out what we dont need. This is for

- configuration
- batch
- and some utils things

peterdudfield:
# Pull Request

## Description

remove ocf_datapipes

Fixes #58 

## How Has This Been Tested?

Ci tests
- [ ] Yes

## Checklist:

### DIFF
--- a/ocf_data_sampler/constants.py
+++ b/ocf_data_sampler/constants.py
@@ -25,7 +25,7 @@ class NWPStatDict(dict):
             return super().__getitem__(key)
         else:
             raise KeyError(
-                f"Values for {key} not yet available in ocf-datapipes {list(self.keys())}"
+                f"Values for {key} not yet available in ocf-data-sampler {list(self.keys())}"
             )
 
 # ------ UKV
--- a/ocf_data_sampler/constants.py
+++ b/ocf_data_sampler/constants.py
@@ -1,0 +1,135 @@
+import xarray as xr
+import numpy as np
+
+
+NWP_PROVIDERS = [
+    "ukv",
+    "ecmwf",
+]
+
+
+def _to_data_array(d):
+    return xr.DataArray(
+        [d[k] for k in d.keys()],
+        coords={"channel": [k for k in d.keys()]},
+    ).astype(np.float32)
+
+
+class NWPStatDict(dict):
+    """Custom dictionary class to hold NWP normalizarion stats"""
+
+    def __getitem__(self, key):
+        if key not in NWP_PROVIDERS:
+            raise KeyError(f"{key} is not a supported NWP provider - {NWP_PROVIDERS}")
+        elif key in self.keys():
+            return super().__getitem__(key)
+        else:
+            raise KeyError(
+                f"Values for {key} not yet available in ocf-datapipes {list(self.keys())}"
+            )
+
+# ------ UKV
+# Means and std computed WITH version_7 and higher, MetOffice values
+UKV_STD = {
+    "cdcb": 2126.99350113,
+    "lcc": 39.33210726,
+    "mcc": 41.91144559,
+    "hcc": 38.07184418,
+    "sde": 0.1029753,
+    "hcct": 18382.63958991,
+    "dswrf": 190.47216887,
+    "dlwrf": 39.45988077,
+    "h": 1075.77812282,
+    "t": 4.38818501,
+    "r": 11.45012499,
+    "dpt": 4.57250482,
+    "vis": 21578.97975625,
+    "si10": 3.94718813,
+    "wdir10": 94.08407495,
+    "prmsl": 1252.71790539,
+    "prate": 0.00021497,
+}
+UKV_MEAN = {
+    "cdcb": 1412.26599062,
+    "lcc": 50.08362643,
+    "mcc": 40.88984494,
+    "hcc": 29.11949682,
+    "sde": 0.00289545,
+    "hcct": -18345.97478167,
+    "dswrf": 111.28265039,
+    "dlwrf": 325.03130139,
+    "h": 2096.51991356,
+    "t": 283.64913206,
+    "r": 81.79229501,
+    "dpt": 280.54379901,
+    "vis": 32262.03285118,
+    "si10": 6.88348448,
+    "wdir10": 199.41891636,
+    "prmsl": 101321.61574029,
+    "prate": 3.45793433e-05,
+}
+
+UKV_STD = _to_data_array(UKV_STD)
+UKV_MEAN = _to_data_array(UKV_MEAN)
+
+# ------ ECMWF
+# These were calculated from 100 random init times of UK data from 2020-2023
+ECMWF_STD = {
+    "dlwrf": 15855867.0,
+    "dswrf": 13025427.0,
+    "duvrs": 1445635.25,
+    "hcc": 0.42244860529899597,
+    "lcc": 0.3791404366493225,
+    "mcc": 0.38039860129356384,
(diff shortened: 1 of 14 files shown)


pyproject.toml
```diff
@@ -18,14 +18,19 @@ maintainers = [
 ]
 
 dependencies = [ # Migration from requirements.txt
+    "torch<2.5.0",
```
Sukh-P, line 21: As discussed I don't believe this is needed after removing the pinned ocf_datapipes version

Sukh-P: Can try and remove it to double check the tests pass

ocf_data_sampler/constants.py
```diff
@@ -0,0 +1,135 @@
...
+
+def _to_data_array(d):
+    return xr.DataArray(
+        [d[k] for k in d.keys()],
+        coords={"channel": [k for k in d.keys()]},
+    ).astype(np.float32)
+
+
+class NWPStatDict(dict):
+    """Custom dictionary class to hold NWP normalizarion stats"""
```
AUdaltsova, line 19:
```suggestion
    """Custom dictionary class to hold NWP normalization stats"""
```

ocf_data_sampler/constants.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def __getitem__(self, key):
+        if key not in NWP_PROVIDERS:
+            raise KeyError(f"{key} is not a supported NWP provider - {NWP_PROVIDERS}")
+        elif key in self.keys():
+            return super().__getitem__(key)
+        else:
+            raise KeyError(
+                f"Values for {key} not yet available in ocf-datapipes {list(self.keys())}"
```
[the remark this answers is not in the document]
peterdudfield: Currently we always normalize. The NWP values for different channels have quite a big range, so its good to normalize them

pyproject.toml
peterdudfield: Nice good catch, it worked without the pin

ocf_data_sampler/constants.py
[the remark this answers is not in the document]
peterdudfield: Maybe I'm reading it wrong. But doesnt this bit of code ensure that we have put normalizing constants in. If we remove it, we might forget to add normalizing constant it when adding a new NWP?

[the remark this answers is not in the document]
AUdaltsova: oh yeah no, i meant remove the in providers check (the first if), sorry, wrong code context! Basically if there are no values, it doesn't matter if it's in providers (it shouldn't be, and actually if we get to this point then it is in providers for some reason since the config check didn't catch it), this is not an allowed nwp source. So I think it might be better if it just goes for the values straight away and if they are not there fails hard, instead of popping two different warnings when the provider is not in the list and when the values are not there, if that makes sense?

Tangentially, I think there should be a check that all channels were found in consts, as this can be an issue with accumulated channels getting diffed, but definitely not in this pr; I'l go make a new issue for that

ocf_data_sampler/numpy_batch/sun_position.py
```diff
@@ -38,7 +37,7 @@ def make_sun_position_numpy_batch(
         lon: float, 
         lat: float, 
         key_preffix: str = "gsp"
```
AUdaltsova, line 39:
```suggestion
        key_prefix: str = "gsp"
```

ocf_data_sampler/numpy_batch/sun_position.py
```diff
@@ -58,9 +57,9 @@ def make_sun_position_numpy_batch(
     elevation = elevation / 180 + 0.5
     
     # Make NumpyBatch
-    sun_numpy_batch: NumpyBatch = {
-        BatchKey[key_preffix + "_solar_azimuth"]: azimuth,
-        BatchKey[key_preffix + "_solar_elevation"]: elevation,
+    sun_numpy_batch = {
+        key_preffix + "_solar_azimuth": azimuth,
+        key_preffix + "_solar_elevation": elevation,
```
AUdaltsova, line 62:
```suggestion
        key_prefix + "_solar_azimuth": azimuth,
        key_prefix + "_solar_elevation": elevation,
```

ocf_data_sampler/numpy_batch/gsp.py
```diff
@@ -1,20 +1,33 @@
...
+    gsp_time_utc = 'gsp_time_utc'
+    gsp_t0_idx = 'gsp_t0_idx'
+    gsp_solar_azimuth = 'gsp_solar_azimuth'
+    gsp_solar_elevation = 'gsp_solar_elevation'
+    gsp_id = 'gsp_id'
+    gsp_x_osgb = 'gsp_x_osgb'
+    gsp_y_osgb = 'gsp_y_osgb'
+
+
+def convert_gsp_to_numpy_batch(da: xr.DataArray, t0_idx: int | None = None) -> dict:
```
Sukh-P, line 20: If we are removing the `NumpyBatch` and other batch types like the `NWPNumpyBatch` and just calling them dicts should we remove reference to the types as well as well? Like in the docstrings here and other files?

tests/torch_datasets/test_pvnet_uk_regional.py
```diff
@@ -3,7 +3,9 @@
 
 from ocf_data_sampler.torch_datasets.pvnet_uk_regional import PVNetUKRegionalDataset
 from ocf_data_sampler.config import load_yaml_configuration, save_yaml_configuration
-from ocf_datapipes.batch import BatchKey, NWPBatchKey
+from ocf_data_sampler.numpy_batch.nwp import NWPBatchKey
+from ocf_data_sampler.numpy_batch.gsp import GSPBatchKey
+from ocf_data_sampler.numpy_batch.satellite import SatelliteBatchKey
```
AUdaltsova, line 8: we can import those into numpy_batch __init__ for neater imports? I'm happy to do that if you want

AUdaltsova (approved):
Looks great, thank you sm for doing this! Out of curiosity: why opt to use three different batch key instances one per source instead of single batchkey class? Is it because it's getting reworked?

tests/numpy_batch/test_sun_position.py
```diff
@@ -71,11 +71,11 @@ def test_make_sun_position_numpy_batch():
 
     batch = make_sun_position_numpy_batch(datetimes, lon, lat, key_preffix="gsp")
```
AUdaltsova, line 72:
```suggestion
    batch = make_sun_position_numpy_batch(datetimes, lon, lat, key_prefix="gsp")
```

peterdudfield:
> Looks great, thank you sm for doing this! Out of curiosity: why opt to use three different batch key instances one per source instead of single batchkey class? Is it because it's getting reworked?

Not necessary a decision, and can be reworked. I thought it somehow sat quite well with each type of data input. But could be changed

AUdaltsova:
> Not necessary a decision, and can be reworked. I thought it somehow sat quite well with each type of data input. But could be changed

yeah I agree! One thing that I keep thinking: do we want to remove prefixes from keys then? Like `gsp_` in `gsp_t0_index`. It looks like it's always used like `GSPBatchKey.gsp_t0_index`, so should be clear what's used? But I could be missing something

peterdudfield:
Yea I definitely considered about it, And then decided it could be done in a separate PR. would you like to make an issue for this?

Issue #66, opened 2024-09-04
Load ICON data from HF


peterdudfield:
## Detailed Description
It would be great to be able to load the ICON data from HF in our nwp open dataset

## Context
- https://huggingface.co/datasets/openclimatefix/dwd-icon-eu
- Discussion on HF https://huggingface.co/datasets/openclimatefix/dwd-icon-eu/discussions/7
- most other nwp are one big zarr, or a few big zarr files. Howevere on HF there are lots 2 per day, of zarr.zip
- We need to lazily load the data, and then once we have selcted the time and location we can load the small amount of data in. 

## Possible Implementation
- Try to find a method that can lazily load the ICON data from HF in load nwp

peterdudfield:
Thanks for doing this, and nice to see the profile. Whats the current timings for making an xarray with the first 20? 
You totaly sure its only getting the metadata, not downloading all the data?

Do you know roughtly how many paths until you get a 429?

gabrielelibardi:
It takes 58 secs to create an xarray with 20. Pretty sure it is not downloading the data (that would be more than 60 GB in 60 secs a bit too crazy for my laptop). 
I think I was trying 50 or so and I was already getting a 429. Maybe it is because the .zmetadata are too small so maybe there are a lot of separate requests too quickly, if I download a whole .zip.zarr file it does not complain.

gabrielelibardi:
For training with ECMWF data you also use multiple zarr file or just one? You stream it from S3 buckets or keep it on the training server?

peterdudfield:
> For training with ECMWF data you also use multiple zarr file or just one? You stream it from S3 buckets or keep it on the training server?

For training ECMWF, we tend to join our daily ECMWF files into year (or monthly) files and then open multi open zarrs. We tend to try to keep the data close to where the model is running, so either locally if trianing localy, or in cloud, if we are training in the cloud. It would be nice to stream from HF though, so we dont have to re organise the files, and the live data is available. 

Thanks for these benchmark figures. Yea I agree it's just downloading metadata not the whole thing. 

Feels like there should be some caching we can do, so for example, loading metadata for each month one, and then quickly loading this new file in the future. I really dont know the solution for this sorry. 

@devsjc @Sukh-P @AUdaltsova might have some ideas?

jacobbieker:
[...]

PR #72, opened 2024-10-24
Added icon to nwp providers

gabrielelibardi:
# Pull Request

## Description

I added icon to the nwp providers. Specifically the changes should allow to an xarray lazily from a list of .zarr or .zarr.zip paths downloaded from here https://huggingface.co/datasets/openclimatefix/dwd-icon-eu. 
This pull request is created to address this issue https://github.com/openclimatefix/ocf-data-sampler/issues/66#issuecomment-2432560799.
In principle this should work even if ones uses remote paths directly to the the .zarr.zip files however because of the many request made to the hugging face server in a short time this may result in a 429 Error. There are ways around this as mentioned in the issue, that have not yet been implemented.    

Fixes #

## How Has This Been Tested?

see the test_load_icon_eu added to ocf-data-sampler/tests/load/test_load_nwp.py

- [X ] Yes

## Checklist:

### DIFF
--- a/ocf_data_sampler/load/nwp/nwp.py
+++ b/ocf_data_sampler/load/nwp/nwp.py
@@ -3,7 +3,7 @@ import xarray as xr
 
 from ocf_data_sampler.load.nwp.providers.ukv import open_ukv
 from ocf_data_sampler.load.nwp.providers.ecmwf import open_ifs
-
+from ocf_data_sampler.load.nwp.providers.icon import open_icon_eu
 
 def open_nwp(zarr_path: Path | str | list[Path] | list[str], provider: str) -> xr.DataArray:
     """Opens NWP Zarr
@@ -17,6 +17,8 @@ def open_nwp(zarr_path: Path | str | list[Path] | list[str], provider: str) -> x
         _open_nwp = open_ukv
     elif provider.lower() == "ecmwf":
         _open_nwp = open_ifs
+    elif provider.lower() == "icon-eu":
+        _open_nwp = open_icon_eu
     else:
         raise ValueError(f"Unknown provider: {provider}")
     return _open_nwp(zarr_path)
--- a/ocf_data_sampler/load/nwp/providers/icon.py
+++ b/ocf_data_sampler/load/nwp/providers/icon.py
@@ -1,0 +1,101 @@
+"""DWD ICON Loading"""
+
+import pandas as pd
+import xarray as xr
+import fsspec
+
+from ocf_data_sampler.load.nwp.providers.utils import open_zarr_paths
+
+def transform_to_channels(nwp : xr.Dataset):
+    """
+    Adds a channel dimension to the NWP data
+    
+    Args:
+        nwp: NWP data without channel dimension
+    
+    Returns:
+        NWP data with channel dimension
+    """
+    
+    channel_data = []
+    channel_names = []
+
+    for var_name in nwp.data_vars:
+        data_array = nwp[var_name]
+        expanded_data = data_array.expand_dims(dim={"channel": [var_name]})
+        
+        channel_data.append(expanded_data)
+        channel_names.append(var_name)
+
+    nwp_channels = xr.concat(channel_data, dim="channel")
+    nwp_channels["channel"] = channel_names
+    
+    return nwp_channels
+
+def remove_isobaric_lelvels_from_coords(nwp: xr.Dataset) -> xr.Dataset:
+    """
+    Removes the isobaric levels from the coordinates of the NWP data
+    
+    Args:
+        nwp: NWP data
+    
+    Returns:
+        NWP data without isobaric levels in the coordinates
+    """
+    variables_to_drop = [var for var in nwp.data_vars if 'isobaricInhPa' in nwp[var].dims]
+    return nwp.drop_vars(["isobaricInhPa"] + variables_to_drop)
+
+def open_icon_eu(zarr_path) -> xr.Dataset:
+    """
+    Opens the ICON data
+
+    ICON EU Data is on a regular lat/lon grid
+    It has data on multiple pressure levels, as well as the surface
+    Each of the variables is its own data variable
+
+    Args:
+        zarr_path: Path to the zarr to open
+
+    Returns:
+        Xarray DataArray of the NWP data
+    """
+    # Open the data
+    nwp = open_zarr_paths(zarr_path, time_dim="time")
+    nwp = nwp.rename({"time": "init_time_utc"})
+    # Sanity checks.
+    time = pd.DatetimeIndex(nwp.init_time_utc)
+    assert time.is_unique
+    assert time.is_monotonic_increasing
+    nwp = nwp.isel(step=slice(0, 48))
+    nwp = remove_isobaric_lelvels_from_coords(nwp)
+    nwp =  transform_to_channels(nwp)
+    print("loaded icon eu data with shape", nwp.shape)
+    return nwp.transpose('init_time_utc', 'step', 'channel', 'latitude', 'longitude')
+
+
+def open_icon_global(zarr_path) -> xr.Dataset:
+    """
+    Opens the ICON data
+
+    ICON Global Data is on an isohedral grid, so the points are not regularly spaced
+    It has data on multiple pressure levels, as well as the surface
+    Each of the variables is its own data variable
+
+    Args:
+        zarr_path: Path to the zarr to open
+
+    Returns:
+        Xarray DataArray of the NWP data
+    """
+    # Open the data
+    nwp = open_zarr_paths(zarr_path, time_dim="time")
+    nwp = nwp.rename({"time": "init_time_utc"})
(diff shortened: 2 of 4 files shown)


ocf_data_sampler/load/nwp/providers/icon.py
```diff
@@ -0,0 +1,101 @@
+"""DWD ICON Loading"""
+
+import pandas as pd
+import xarray as xr
+import fsspec
+
+from ocf_data_sampler.load.nwp.providers.utils import open_zarr_paths
+
+def transform_to_channels(nwp : xr.Dataset):
```
Sukh-P, line 9: Am I right in thinking that the input here is an xarray `Dataset` which has multiple data variables for each NWP variable and we want to go from that to a `DataArray` (e.g. one data variable but an extra channel dimension?)

I think a simpler approach might be to do something like what is done here https://github.com/openclimatefix/ocf_datapipes/blob/main/ocf_datapipes/load/nwp/providers/gfs.py#L26 where we use `to_array()` on the Dataset to convert it to a `DataArray` and then rename the `variable` dimension which is created with `to_array()` to `channel`

But I may have misunderstood the intention/need for this function

Sukh-P:
Thanks for creating this PR and the great work already done on trying to support ICON data in this library!

Something to note is that if this is added in as is that people may assume this library already supports ICON data but without some normalisation constants added and ICON listed as an NWP provider [here](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/constants.py) creating samples from it won't work, so my suggestion is that either this is added in this PR, or in a subsequent PR or this outstanding work is clearly documented in a Github issue or README, thanks!

ocf_data_sampler/load/nwp/providers/icon.py
gabrielelibardi: you are perfectly right, I deleted this and use to_array() instead, thx for pointing it out!

gabrielelibardi:
> Thanks for creating this PR and the great work already done on trying to support ICON data in this library!
> 
> Something to note is that if this is added in as is that people may assume this library already supports ICON data but without some normalisation constants added and ICON listed as an NWP provider [here](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/constants.py) creating samples from it won't work, so my suggestion is that either this is added in this PR, or in a subsequent PR or this outstanding work is clearly documented in a Github issue or README, thanks!

I can compute the std and mean constants, do you have a script to do this for others NWP? How large of a sample do you take?

Sukh-P:
> > Thanks for creating this PR and the great work already done on trying to support ICON data in this library!
> > Something to note is that if this is added in as is that people may assume this library already supports ICON data but without some normalisation constants added and ICON listed as an NWP provider [here](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/constants.py) creating samples from it won't work, so my suggestion is that either this is added in this PR, or in a subsequent PR or this outstanding work is clearly documented in a Github issue or README, thanks!
> 
> I can compute the std and mean constants, do you have a script to do this for others NWP? How large of a sample do you take?

Thanks, that would be great! So I don't think we have a script in Github currently so just created [this gist](https://gist.github.com/Sukh-P/e7748d170f1d68c0d22822d96baed4d3) to share some example code of how I have calculated some of the normalisation stats previously, in the example I used 200 samples I think that would be fine for this too

Sukh-P:
Hey @gabrielelibardi, hope you are doing well! Coming back to this one after a while and just wanted to check if you're planning on working on this still? If not maybe we can merge this in as is and then I can create another issue for someone to calculate the normalisation constants for icon-eu, thanks!

gabrielelibardi:
Hey @Sukh-P hope you are also doing well! Still planning to work on this, but a bit busy at the moment with other stuff. You can also merge if nothing more is needed, up to you.

ocf_data_sampler/load/nwp/providers/icon.py
```diff
@@ -0,0 +1,75 @@
...
+    Returns:
+        Xarray DataArray of the NWP data
+    """
+    # Open the data
+    nwp = open_zarr_paths(zarr_path, time_dim="time")
+    nwp = nwp.rename({"time": "init_time_utc"})
+    # Sanity checks.
+    time = pd.DatetimeIndex(nwp.init_time_utc)
+    assert time.is_unique
+    assert time.is_monotonic_increasing
```
Sukh-P, line 42: We actually have a helper function which does this [here](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/load/utils.py#L4) now so would be good to use in this file!

ocf_data_sampler/load/nwp/providers/icon.py
```diff
@@ -0,0 +1,75 @@
...
+    # 0–78 one hour steps, rest 3 hour steps
+    nwp = nwp.isel(step=slice(0, 78))
+    nwp = remove_isobaric_lelvels_from_coords(nwp)
+    nwp = nwp.to_array().rename({"variable": "channel"})
+    return nwp.transpose('init_time_utc', 'step', 'channel', 'latitude', 'longitude')
```
Sukh-P, line 47: Before returning this it would be good to include [this check](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/load/nwp/providers/ecmwf.py#L41) so that all NWPs have the same coordinate ordering lat/longs

Sukh-P:
> Hey @Sukh-P hope you are also doing well! Still planning to work on this, but a bit busy at the moment with other stuff. You can also merge if nothing more is needed, up to you.

Okay great, so I have left a few comments which would be good to address, I don't think we need to add anything else apart from those to this PR, so after that we can merge this in. 

There are some normalisation constants to be calculated as per my comments previously but I can create a separate issue for that, may need your input on how someone goes about fetching this data (as I believe you already fetched this from our HF) and then I can point to the gist I linked above for how someone can calculate these. 

Thanks for the input!

ocf_data_sampler/load/nwp/nwp.py
```diff
@@ -17,6 +17,8 @@ def open_nwp(zarr_path: Path | str | list[Path] | list[str], provider: str) -> x
         _open_nwp = open_ukv
     elif provider.lower() == "ecmwf":
         _open_nwp = open_ifs
+    elif provider.lower() == "icon-eu":
```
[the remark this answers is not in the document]
gabrielelibardi: right, I only have tried to run this with icon-eu data so I would stick to that, I now removed the icon-global

gabrielelibardi:
Hey @Sukh-P I made the changes you suggested, I also added some files to test the loading for icon (they are a bit heavy so we can remove those), also added the script to compute the mean and std_dev under utils, not sure if I should include it here or somewhere else or just leave it out.

gabrielelibardi:
Hey @Sukh-P, makes sense, I just changed that so the test data is created on the fly. Let me know if there is anything else!

Sukh-P:
All good I think! Just need to resolve some of the merge conflicts then should be good to go, thanks for all the work on this!

gabrielelibardi:
there was a test failing once I rebased (test_copy_batch_to_device so I had to tweak it, let me know if it is ok

Sukh-P:
@gabrielelibardi any chance you could try removing that change to the test just so I can see the CI tests run to see if that particular tests fails? We just updated our CI workflow so they weren't running previously, sorry

gabrielelibardi:
@Sukh-P sure, I reverted it now

gabrielelibardi:
@Sukh-P btw I did not include all the multi level variables for the icon eu model, I saw that these are also not always included in the other nwps. If I include them I would say that we need to compute the mean and std_var for each level separately, but then again higher levels are not really that useful for pv generation forecast. So is it ok if for now only a subset of the variables is included?

gabrielelibardi:
@Sukh-P could you share what commands from the nwp-consumer exactly do you run to download the icon-eu data you put on hugging face. I ran it to download the full parameters and somehow I don't get the same result as what I download from hugging face

Sukh-P:
@gabrielelibardi I think this is ready to merge in! I feel like the test should be okay because I have seen it run successfully before, if it fails after merging I will update it!



> @Sukh-P btw I did not include all the multi level variables for the icon eu model, I saw that these are also not always included in the other nwps. If I include them I would say that we need to compute the mean and std_var for each level separately, but then again higher levels are not really that useful for pv generation forecast. So is it ok if for now only a subset of the variables is included?

This makes sense to me, we usually just care about surface/ground level variables anyway for PV forecasting so all good, thank you! 


> @Sukh-P could you share what commands from the nwp-consumer exactly do you run to download the icon-eu data you put on hugging face. I ran it to download the full parameters and somehow I don't get the same result as what I download from hugging face. Opened a related issue here https://github.com/openclimatefix/nwp-consumer/issues/240

I think @devsjc will get back to you on this on that issue you raised, thanks for opening it!

PR #82, opened 2024-11-19
Site torch dataset update

Sukh-P:
# Pull Request

## Description

WIP PR to update the Site Torch Dataset to return samples as xarray Datasets for easier conversion into netcdf files which is the preferred current format of saving samples.  

This PR includes:

- Adding a new functions to process the sample dict (dict with xr DataArrays) into one Dataset
- Reordering of when .compute() is called since now we combine multiple DataArrays into a Dataset we can call compute after this is done
- Removed unused site specific parts from original process and combine function
- Updating unit tests now that the data type of the sample has changes in the Torch Dataset
- Updating some time interval syntax to stop a deprecation warning (unrelated to the above changes)

TODO 
- Removed saving solar coordinates data in samples for now, current idea is to use the numpy batch functions in here in PVNet to create this data when converting to a numpy batch (if this seems messy may add some logic in here to add to the solar position coordinates to the Dataset)
- Check this works by creating some samples and adding logic into PVNet to read the netcdfs and convert to NumpyBatch/TensorBatches and then train a model, will link PR here once that is done

### DIFF
--- a/ocf_data_sampler/torch_datasets/process_and_combine.py
+++ b/ocf_data_sampler/torch_datasets/process_and_combine.py
@@ -10,7 +10,6 @@ from ocf_data_sampler.numpy_batch import (
     convert_satellite_to_numpy_batch,
     convert_gsp_to_numpy_batch,
     make_sun_position_numpy_batch,
-    convert_site_to_numpy_batch,
 )
 from ocf_data_sampler.numpy_batch.gsp import GSPBatchKey
 from ocf_data_sampler.numpy_batch.nwp import NWPBatchKey
@@ -74,18 +73,6 @@ def process_and_combine_datasets(
             }
         )
 
-
-    if "site" in dataset_dict:
-        site_config = config.input_data.site
-        da_sites = dataset_dict["site"]
-        da_sites = da_sites / da_sites.capacity_kwp
-
-        numpy_modalities.append(
-            convert_site_to_numpy_batch(
-                da_sites, t0_idx=-site_config.interval_start_minutes / site_config.time_resolution_minutes
-            )
-        )
-
     if target_key == 'gsp':
         # Make sun coords NumpyBatch
         datetimes = pd.date_range(
@@ -95,16 +82,6 @@ def process_and_combine_datasets(
         )
 
         lon, lat = osgb_to_lon_lat(location.x, location.y)
-
-    elif target_key == 'site':
-        # Make sun coords NumpyBatch
-        datetimes = pd.date_range(
-            t0+minutes(site_config.interval_start_minutes),
-            t0+minutes(site_config.interval_end_minutes),
-            freq=minutes(site_config.time_resolution_minutes),
-        )
-
-        lon, lat = location.x, location.y
 
     numpy_modalities.append(
         make_sun_position_numpy_batch(datetimes, lon, lat, key_prefix=target_key)
--- a/ocf_data_sampler/torch_datasets/process_and_combine.py
+++ b/ocf_data_sampler/torch_datasets/process_and_combine.py
@@ -120,7 +120,17 @@ def process_and_combine_site_sample_dict(
     dataset_dict: dict,
     config: Configuration,
 ) -> xr.Dataset:
-    """Normalize and combine data to xr Dataset"""
+    """
+    Normalize and combine data into a single xr Dataset
+
+    Args:
+        dataset_dict: dict containing sliced xr DataArrays
+        config: Configuration for the model
+
+    Returns:
+        xr.Dataset: A merged Dataset with nans filled in.
+    
+    """
 
     data_arrays = []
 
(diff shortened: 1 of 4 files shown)


peterdudfield:
Thanks @Sukh-P great to push this forward. 

A few quick thoughts, and sorry if these seem obvious and have already been answered
1. is the ideal still to convert the site batch dataset to a dict of tensors ready for the model (PVnet)? if so, will this code sit in here, or in PVNet
2. For different torch dataloaders, do we have an idea of how to do this for the three different process? 1. make batches, 2. load batches and train model, 3. running inference. It would be a shame to have separate torch datasets for each loader, but perhaps there is a simple way to do this. This is very much in your TODO section, so perhaps you have thought about this already / going to next. 
3. This might be related to 2., but do you know where the combining samples to batch process fits in?

Sukh-P:
@peterdudfield thanks, I have tried to answer these: 
> 1. is the ideal still to convert the site batch dataset to a dict of tensors ready for the model (PVnet)? if so, will this code sit in here, or in PVNet

Yes that's still the plan, perhaps still making a NumpyBatch if we want to have a more generic intermediary format, and yes the code will probably be added to here but then called in PVNet, can make that clearer in the TODO list above 

> 2. For different torch dataloaders, do we have an idea of how to do this for the three different process? 1. make batches, 2. load batches and train model, 3. running inference. It would be a shame to have separate torch datasets for each loader, but perhaps there is a simple way to do this. This is very much in your TODO section, so perhaps you have thought about this already / going to next.

I think the longer term plan is to move towards one batch format (netcdf) and have a common interface with batches through a batch object, in this way things will be more generalised and we will have less of having a different way to do things each time, I imagine this will need a bit more thought and can be improved after adding in a working pipeline for sites, can create an issue/discussion around this after we have 

> 3. This might be related to 2., but do you know where the combining samples to batch process fits in?

So I think this is managed by having a function which does some stacking of samples like [here](https://github.com/openclimatefix/ocf_datapipes/blob/main/ocf_datapipes/batch/merge_numpy_examples_to_batch.py#L56) into a batch and the Torch DataLoader where you specify how many samples would be in a batch

tests/select/test_select_time_slice.py
```diff
@@ -197,7 +197,7 @@ def test_select_time_slice_nwp_with_dropout_and_accum(da_nwp_like, t0_str):
     t0 = pd.Timestamp(f"2024-01-02 {t0_str}")
     interval_start = pd.Timedelta(-6, "h")
     interval_end = pd.Timedelta(3, "h")
-    freq = pd.Timedelta("1H")
+    freq = pd.Timedelta("1h")
```
Sukh-P, line 200: To stop a syntax deprecation warning from popping up

peterdudfield (approved):
One small comment could be added, but otherwise looks great

ocf_data_sampler/torch_datasets/process_and_combine.py
```diff
@@ -115,6 +93,47 @@ def process_and_combine_datasets(
...
+    if "nwp" in dataset_dict:
+        for nwp_key, da_nwp in dataset_dict["nwp"].items():
+            # Standardise
+            provider = config.input_data.nwp[nwp_key].provider
+            da_nwp = (da_nwp - NWP_MEANS[provider]) / NWP_STDS[provider]
+            data_arrays.append((f"nwp-{provider}", da_nwp))
+          
+    if "sat" in dataset_dict:
+        # Satellite is already in the range [0-1] so no need to standardise
+        da_sat = dataset_dict["sat"]
```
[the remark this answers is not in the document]
peterdudfield: #87

ocf_data_sampler/torch_datasets/process_and_combine.py
```diff
@@ -124,6 +143,59 @@ def merge_dicts(list_of_dicts: list[dict]) -> dict:
...
+        datasets.append(dataset)
+
+    # Ensure all datasets are valid xarray.Dataset objects
+    for ds in datasets:
+        assert isinstance(ds, xr.Dataset), f"Object is not an xr.Dataset: {type(ds)}"
+
+    # Merge all prepared datasets
+    combined_dataset = xr.merge(datasets)
+
+    return combined_dataset
```
AUdaltsova, line 198: Out of curiosity bc I think I've missed a conversation somewhere: am I right in thinking that eventually gsp version will also output netcdfs and we will merge it into this function and move convert_to_numpy_batch into the training pipeline?

ocf_data_sampler/torch_datasets/site.py
```diff
@@ -152,10 +152,9 @@ def _get_sample(self, t0: pd.Timestamp, location: Location) -> dict:
         """
         sample_dict = slice_datasets_by_space(self.datasets_dict, location, self.config)
         sample_dict = slice_datasets_by_time(sample_dict, t0, self.config)
-        sample_dict = compute(sample_dict)
-
-        sample = process_and_combine_datasets(sample_dict, self.config, t0, location, target_key='site')
 
+        sample = process_and_combine_site_sample_dict(sample_dict, self.config)
+        sample = sample.compute()
```
AUdaltsova, line 157: nice!

AUdaltsova (approved):

AUdaltsova:
Looks super good, thanks a lot for doing all this and the tests!

ocf_data_sampler/torch_datasets/process_and_combine.py
Sukh-P: Yep exactly I think that is the current thinking, the convert_to_numpy_batch logic can still be in this repo and then we can call that in PVNet in the training pipeline

PR #97, opened 2024-12-20
Satellite normalisation

felix-e-h-p:
Normalisation introduced for satellite data utilising 2020 values for std and mean

Not entirely sure epsilon factor is truly necessary - yet left in as a safe guarding mechanism


peterdudfield:
I just re opened it to try and trigger the tests, but weirdly they dont seem to be running

ocf_data_sampler/torch_datasets/process_and_combine.py
```diff
@@ -37,18 +39,32 @@ def process_and_combine_datasets(
...
+
     if "sat" in dataset_dict:
-        # Satellite is already in the range [0-1] so no need to standardise
-        da_sat = dataset_dict["sat"]
 
-        # Convert to NumpyBatch
-        numpy_modalities.append(convert_satellite_to_numpy_batch(da_sat))
+        sat_numpy_modalities = dict()
+
+        for sat_key, da_sat in dataset_dict["sat"].items():
```
peterdudfield, line 54: did this work? Im surprised as I thought we only have one satellite unlike NWP where we have mulitple. This is basically because we have only ever used one satellite in one area, compared to using multie NWP for a model. 

Note in when loading datasets, satellite seems to just be one - https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/load/load_dataset.py#L42

dfulu: Yeh that is true, the satellite datasets aren't nested like the NWPs are since we only use one satellite product

ocf_data_sampler/constants.py
```diff
@@ -133,3 +154,49 @@ def __getitem__(self, key):
...
+    "IR_120": 0.89058584,
+    "IR_134": 0.944365,
+    "VIS006": 0.09633306,
+    "VIS008": 0.11426069,
+    "WV_062": 0.7359355,
+    "WV_073": 0.62479186,
+}
+
+# Specified to ensure calculation stability
+EPSILON = 1e-8
```
peterdudfield, line 191: remove

requirements.txt
```diff
@@ -0,0 +1,136 @@
+aiobotocore
```
peterdudfield, line 1: remove, they are already in pyproject.toml

tests/torch_datasets/test_process_and_combine.py
```diff
@@ -0,0 +1,143 @@
...
+    assert result["level1"]["level2"] == mock_dataarray
+    assert result["another_level1"] == mock_dataarray
+    mock_dataarray.compute.assert_called()
+
+
+def test_process_and_combine_datasets(da_sat_like, da_nwp_like):
+    # Dummy config with valid integer values
+    mock_config = MagicMock()
+    mock_config.input_data.nwp = {"ukv": MagicMock(provider="ukv")}
+    mock_config.input_data.satellite = {"rss": MagicMock(provider="rss")}
```
peterdudfield, line 117: remove provider

tests/torch_datasets/test_process_and_combine.py
```diff
@@ -0,0 +1,143 @@
...
+        interval_end_minutes=30,
+        time_resolution_minutes=15,
+    )
+
+    t0 = pd.Timestamp("2023-01-01 00:00:00")
+    mock_location = MagicMock(x=12345.6, y=65432.1, id=1)
+
+    dataset_dict = {
+        "nwp": {"ukv": da_nwp_like},
+        "sat": {"rss": da_sat_like},
```
peterdudfield, line 129: remove rss

ocf_data_sampler/torch_datasets/process_and_combine.py
```diff
@@ -25,8 +27,8 @@ def process_and_combine_datasets(
     location: Location,
     target_key: str = 'gsp'
 ) -> dict:
-    """Normalize and convert data to numpy arrays"""
 
+    """Normalise and convert data to numpy arrays"""
```
dfulu, line 31: Thanks for this catch 🇬🇧

tests/torch_datasets/test_process_and_combine.py
```diff
@@ -0,0 +1,143 @@
...
+    ]
+    
+    result = merge_dicts(dicts)
+    
+    # Prioritise later values for overlap
+    expected = {"a": 1, "b": 5, "c": 3, "d": 4, "e": 6}
+    assert result == expected
+
+
+def test_fill_nans_in_arrays():
```
dfulu, line 67: This is great

dfulu (changes requested):
Thanks for doing this @felix-e-h-p! It looks good, but I think there are a few things that we should clean up first

tests/torch_datasets/test_process_and_combine.py
```diff
@@ -0,0 +1,143 @@
...
+
+    np.testing.assert_array_equal(result["array_1"], expected["array_1"])
+    np.testing.assert_array_equal(result["array_2"], expected["array_2"])
+    np.testing.assert_array_equal(result["nested_dict"]["array_3"], expected["nested_dict"]["array_3"])
+    assert result["non_numeric"] == expected["non_numeric"]
+
+
+def test_compute():
+    mock_dataarray = MagicMock()
+    mock_dataarray.compute = MagicMock(side_effect=lambda scheduler=None: mock_dataarray)
```
dfulu, line 98: I'm not familiar with this way of testing. Will this test whether the `.compute()` method has been called on each object in the dict?

tests/torch_datasets/test_process_and_combine.py
```diff
@@ -0,0 +1,143 @@
...
+    coords = {
+        "init_time_utc": pd.date_range("2023-01-01", periods=3),
+        "step": pd.timedelta_range("0h", periods=3, freq="1h"),
+        "x_osgb": np.arange(4),
+        "y_osgb": np.arange(4),
+    }
+    return xr.DataArray(
+        data,
+        coords=coords,
+        dims=["init_time_utc", "step", "y_osgb", "x_osgb"],
```
dfulu, line 48: This is strictly not the data format that `process_and_combine_datasets()` expects. It expects that the NWP data will have been passed through the [`select_time_slice_nwp()` function](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/select/select_time_slice.py#L68) function which changes the dimensions from `["init_time_utc", "step", "channel", "x", "y"]` to `["target_time_utc", "channel", "x", y"]`.

Although admittedly the `process_and_combine_datasets()` doesn't really check this since it is only ever supposed to be used as part of the pipeline where `select_time_slice_nwp()` has been run

ocf_data_sampler/numpy_batch/satellite.py
```diff
@@ -13,6 +13,13 @@ class SatelliteBatchKey:
 
 def convert_satellite_to_numpy_batch(da: xr.DataArray, t0_idx: int | None = None) -> dict:
     """Convert from Xarray to NumpyBatch"""
+
+    # Missing coordinate checking stage
+    required_coords = ["x_geostationary", "y_geostationary"]
```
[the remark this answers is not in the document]
felix-e-h-p: Removed on both counts - suppose just unnecessary handling

tests/torch_datasets/test_process_and_combine.py
felix-e-h-p: Updated both constants.py and process_and_combine.py accordingly accounting for epsilon and rss

ocf_data_sampler/torch_datasets/process_and_combine.py
```diff
@@ -37,18 +39,32 @@ def process_and_combine_datasets(
...
-        numpy_modalities.append(convert_satellite_to_numpy_batch(da_sat))
+        sat_numpy_modalities = dict()
+
+        for sat_key, da_sat in dataset_dict["sat"].items():
+            # Standardise
+            provider = config.input_data.satellite[sat_key].provider
+
+            # Not entirely sure if epsilon is necessary considering mean and std values are consistently non-zero
+            # Purely a safety measure
+            da_sat = (da_sat - SAT_MEANS[provider]) / (SAT_STDS[provider] + EPSILON)
```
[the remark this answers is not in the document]
felix-e-h-p: Removed

tests/.DS_Store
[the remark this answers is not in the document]
felix-e-h-p: Yeah - in ignore now sorry, removed from the branch also

tests/torch_datasets/.DS_Store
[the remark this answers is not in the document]
felix-e-h-p: Check

tests/torch_datasets/test_process_and_combine.py
felix-e-h-p: Updated the test to ensure that NWP data is passed through the select_time_slice_nwp function initially before process and combine is called - would appreciate your feedback on this section if all OK

tests/torch_datasets/test_process_and_combine.py
felix-e-h-p: Was previously just to simulate the behaviour of the function on a single 'mock' object, but not on every object in a specified dict. Updated, with additional verification, to ensure that the function is called on each object in the dict

ocf_data_sampler/torch_datasets/process_and_combine.py
```diff
@@ -37,18 +39,23 @@ def process_and_combine_datasets(
...
-        # Satellite is already in the range [0-1] so no need to standardise
+        # Standardise
         da_sat = dataset_dict["sat"]
-
+        da_sat = (da_sat - SAT_MEANS) / SAT_STDS
         # Convert to NumpyBatch
-        numpy_modalities.append(convert_satellite_to_numpy_batch(da_sat))
+        sat_numpy_modalities = convert_satellite_to_numpy_batch(da_sat)
+        # Combine the Satellite into NumpyBatch
+        numpy_modalities.append({SatelliteBatchKey.satellite_actual: sat_numpy_modalities})
```
peterdudfield, line 57: This line, could be changed to back to `numpy_modalities.append(convert_satellite_to_numpy_batch(da_sat)`, and then I think itll be fine

peterdudfield:
Would you be able to pull `main` into this branch, and then add sat normalization to here too - https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/torch_datasets/process_and_combine.py#L122

ocf_data_sampler/constants.py
```diff
@@ -133,3 +136,42 @@ def __getitem__(self, key):
...
+    "IR_108": 0.71254843,
+    "IR_120": 0.89058584,
+    "IR_134": 0.944365,
+    "VIS006": 0.09633306,
+    "VIS008": 0.11426069,
+    "WV_062": 0.7359355,
+    "WV_073": 0.62479186,
+}
+
+SAT_STD = _to_data_array(SAT_STD)
```
dfulu, line 172: I'd prefer if we named these `RRS_[MEAN/STD]` everywhere. I like being more specific that these match to the RSS satellite data

ocf_data_sampler/constants.py
```diff
@@ -133,3 +136,42 @@ def __getitem__(self, key):
...
+SAT_MEAN = _to_data_array(SAT_MEAN)
+
+# SatStatDict wrapper not needed due to singular provider - direct assignment of meand and std
+SAT_STDS = SAT_STD
```
dfulu, line 176: I don't think we shouldn't assign these to new variables. We do this when we wrap all the NWP values into a dict, but since we only have one set of satellite constants I don't think it makes sense here

tests/torch_datasets/test_process_and_combine.py
```diff
@@ -0,0 +1,219 @@
...
+            channel=(["channel"], channels),
+            x_osgb=(["x_osgb"], x),
+            y_osgb=(["y_osgb"], y),
+        )
+    )
+    return da_nwp
+
+
+@pytest.fixture
+def mock_constants(monkeypatch):
```
dfulu, line 65: Is there a reason not to just import these?

tests/torch_datasets/test_process_and_combine.py
```diff
@@ -0,0 +1,219 @@
...
+                self.gsp = self.GSP()
+
+        def __init__(self):
+            self.input_data = self.InputData()
+    
+    return MockConfig()
+
+
+@pytest.fixture
+def mock_location():
```
dfulu, line 111: Since this is only used in a single test and its just a single line of code, I think it would be cleaner just to go i that test

tests/torch_datasets/test_process_and_combine.py
dfulu: For now I think I'd bypass the `select_time_slice_nwp()` function here and just update this fixture. It means this test is independent of errors in `select_time_slice_nwp()`. 

something like:

```
    data = np.random.rand(3, 3, 4, 4)
    coords = {
        "target_time_utc": pd.date_range("2023-01-01 12:00", freq="1h", periods=3),
        "channel": ["t", "dswrf", "lcc"],
        "x_osgb": np.arange(4),
        "y_osgb": np.arange(4),
    }
    xr.DataArray(
        data,
        coords=coords,
        dims=["target_time_utc", "channel", "y_osgb", "x_osgb"],
    )
```

tests/torch_datasets/test_process_and_combine.py
dfulu: Thanks for this addition

tests/torch_datasets/test_process_and_combine.py
felix-e-h-p: Sorted and  updated

ocf_data_sampler/constants.py
felix-e-h-p: Yeah - updated across the board

ocf_data_sampler/torch_datasets/process_and_combine.py
```diff
@@ -119,8 +125,9 @@ def process_and_combine_site_sample_dict(
             data_arrays.append((f"nwp-{provider}", da_nwp))
           
     if "sat" in dataset_dict:
-        # TODO add some satellite normalisation
+        # Satellite normalisation added
```
peterdudfield, line 128: thanks for this

tests/torch_datasets/test_process_and_combine.py
```diff
@@ -0,0 +1,217 @@
...
+from ocf_data_sampler.torch_datasets.process_and_combine import (
+    process_and_combine_datasets,
+    process_and_combine_site_sample_dict,
+    merge_dicts,
+    fill_nans_in_arrays,
+    compute,
+)
+
+
+# Currently leaving here for reference purpose - not strictly needed
```
peterdudfield, line 23: probably good to remove them, rather than have duplicate tests

tests/torch_datasets/test_process_and_combine.py
```diff
@@ -0,0 +1,217 @@
...
+    assert sample[NWPBatchKey.nwp]["ukv"][NWPBatchKey.nwp].shape == (4, 1, 2, 2)
+    # 3 hours of 30 minute data (inclusive)
+    assert sample[GSPBatchKey.gsp].shape == (7,)
+    # Solar angles have same shape as GSP data
+    assert sample[GSPBatchKey.solar_azimuth].shape == (7,)
+    assert sample[GSPBatchKey.solar_elevation].shape == (7,)
+
+
+# Currently leaving here for reference purpose - not strictly needed
+def test_pvnet_no_gsp(pvnet_config_filename):
```
peterdudfield, line 61: same for this one, I think its a duplicate of https://github.com/openclimatefix/ocf-data-sampler/blob/main/tests/torch_datasets/test_pvnet_uk_regional.py#L10

ocf_data_sampler/torch_datasets/process_and_combine.py
dfulu: @felix-e-h-p this is picky but can we change this comment to just `# Standardise` or something. I don't think "Satellite normalisation added" makes sense out of context now the TODO is gone

tests/torch_datasets/test_process_and_combine.py
```diff
@@ -0,0 +1,162 @@
...
+    
+    result = compute(nested_dict)
+    
+    # Ensure function applied - check if data is no longer lazy array and determine structural alterations
+    # Check that result is an xarray DataArray
+    assert isinstance(result["array1"], xr.DataArray)
+    assert isinstance(result["nested"]["array2"], xr.DataArray)
+    
+    # Check data is no longer lazy object
+    assert isinstance(result["array1"].data, np.ndarray)
```
dfulu, line 123: I think this part still needs a little fine tuning. Since `da` is initially filed with a numpy array with line would pass even without running the compute function - i.e. the xarray object was never lazy. 

I think we should use a dask array in `da` instead. This code snippet shows what I mean here

```
import xarray as xr
import numpy as np
import dask.array

da_numpy = xr.DataArray(np.random.rand(5, 5))
print(isinstance(da_numpy.data, np.ndarray))

da_dask = xr.DataArray(dask.array.random.random(5, 5))
print(isinstance(da_dask.data, np.ndarray))

print(isinstance(da_dask.compute().data, np.ndarray))

>>> True
>>> False
>>> True
```

tests/torch_datasets/test_process_and_combine.py
```diff
@@ -0,0 +1,162 @@
...
+    fill_nans_in_arrays,
+    compute,
+)
+
+
+def test_process_and_combine_datasets(pvnet_config_filename):
+
+    # Load in config for function and define location 
+    config = load_yaml_configuration(pvnet_config_filename)
+    t0 = pd.Timestamp("2024-01-01")
```
dfulu, line 27: Can we be explicit about the time here. I believe this defaults to 00:00. If so can we replace this with `t0 = pd.Timestamp("2024-01-01 00:00")`

tests/torch_datasets/test_process_and_combine.py
```diff
@@ -0,0 +1,162 @@
...
+
+    # Specify minimal structure for testing
+    raw_nwp_values = np.random.rand(4, 1, 2, 2)  # Single channel
+    site_dict = {
+        "nwp": {
+            "ukv": xr.DataArray(
+                raw_nwp_values,
+                dims=["time_utc", "channel", "y", "x"],
+                coords={
+                    "time_utc": pd.date_range("2024-01-01", periods=4, freq="h"),
```
dfulu, line 143: Can we be explicit about the time here too

tests/torch_datasets/test_process_and_combine.py
```diff
@@ -0,0 +1,162 @@
...
+    # Load in config for function and define location 
+    config = load_yaml_configuration(pvnet_config_filename)
+    t0 = pd.Timestamp("2024-01-01")
+    location = Location(coordinate_system="osgb", x=1234, y=5678, id=1)
+
+    nwp_data = xr.DataArray(
+        np.random.rand(4, 2, 2, 2),
+        dims=["time_utc", "channel", "y", "x"],
+        coords={
+            "time_utc": pd.date_range("2024-01-01", periods=4, freq="h"),
```
dfulu, line 34: Can we be explicit about the time here too

ocf_data_sampler/torch_datasets/process_and_combine.py
```diff
@@ -13,6 +13,7 @@
 )
 from ocf_data_sampler.numpy_batch.gsp import GSPBatchKey
 from ocf_data_sampler.numpy_batch.nwp import NWPBatchKey
+from ocf_data_sampler.numpy_batch.satellite import SatelliteBatchKey
```
peterdudfield, line 16: that used?

peterdudfield (approved):

ocf_data_sampler/torch_datasets/process_and_combine.py
dfulu: @felix-e-h-p can we change this comment

peterdudfield:
@all-contributors please add @felix-e-h-p  for code

Issue #71, opened 2024-10-21
Create batch class


Sukh-P:
Following on from discussions here #21 we plan to implement a Batch class/object for each torch Dataset where we add functions such as save, load and plot to help standardise things.

peterdudfield:
Shall we close this for now?

PR #125, opened 2025-01-15
Batch / Sample class - draft

felix-e-h-p:
## Overview

First stages of a standardised sample handling system, with unified interface across different dataset types - refer to issue #71. Initial focus regarding Dataset type is PVNet UK Regional. As of the moment, base.py and uk_regional.py function as effective parent and child classes, respectively.

## Main Components

1. Base Sample Class (SampleBase)
	•	Common interface for all sample types
	•	Standardised I/O operations (.nc, .zarr, .npz)
	•	Type conversion utilities (numpy ↔ torch)
	•	Nested data structure support (primarily for NWP)
	•	Save, load and plot implementation

2. PVNet Implementation (PVNetSample)
	•	Extends base class with PVNet UK Regional as ‘child’ class
	•	Implements required feature space validation
	•	Integration with UK Regional dataset

3. Utility Functions (utils.py)
	•	Sample stacking and merging operations
	•	Batch to sample conversion helpers

## Main Implementation

base.py
	•	Implements abstract SampleBase class
	•	Handles both flat and nested data structures
	•	Provides robust I/O and type conversion
	•	Implements NaN handling and validation

uk_regional.py
	•	Implements PVNetSample
	•	Defines PV-specific feature requirements
	•	Integrates with existing dataset structure
	•	Custom plotting functionality


felix-e-h-p:
To mention, any point in updating type hints to be effectively more robust?

I.e. using TypeVar:

```python
T = TypeVar('T', bound='SampleBase')
ArrayType = Union[np.ndarray, torch.Tensor, xr.DataArray]

class SampleBase(ABC, Generic[T]):
    def to_torch(self) -> T:
```

Suppose important for early and type related errors, particularly with multiple refactors.

ocf_data_sampler/sample/uk_regional.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+class PVNetSample(SampleBase):
+    """ Sample class specific to PVNet """
+    
+    # Feature space definitions
+    REQUIRED_KEYS = {
+        'nwp',
+        GSPSampleKey.gsp,
+        SatelliteSampleKey.satellite_actual,
```
peterdudfield, line 35: This could be optional

ocf_data_sampler/sample/utils.py
```diff
@@ -0,0 +1,96 @@
...
+import numpy as np
+import logging
+from typing import List, Dict, Any, Type
+
+from ocf_data_sampler.sample.base import SampleBase
+
+
+logger = logging.getLogger(__name__)
+
+
```
peterdudfield, line 17: Could be duplicate, to check

tests/test_sample/test_uk_regional.py
```diff
@@ -0,0 +1,212 @@
...
+)
+
+from ocf_data_sampler.torch_datasets.pvnet_uk_regional import PVNetUKRegionalDataset
+
+from ocf_data_sampler.sample.uk_regional import PVNetSample
+
+
+# Fixture define
+@pytest.fixture
+def pvnet_config_filename(tmp_path):
```
peterdudfield, line 26: move to json, check not in conftest already

ocf_data_sampler/sample/uk_regional.py
```diff
@@ -0,0 +1,213 @@
...
+            instance.validate()
+            logger.debug(f"Successfully loaded PVNetSample from {path}")
+            return instance
+        except Exception as e:
+            logger.error(f"Error loading from {path}: {str(e)}")
+            raise
+
+    def plot(self, **kwargs) -> None:
+        """ Sample visualisation definition """
+        logger.debug("Creating PVNetSample visualisation")
```
peterdudfield, line 180: Could update later using ocf-datapipes visulation.

ocf_data_sampler/sample/uk_regional.py
```diff
@@ -0,0 +1,213 @@
...
+    # Feature space definitions
+    REQUIRED_KEYS = {
+        'nwp',
+        GSPSampleKey.gsp,
+        SatelliteSampleKey.satellite_actual,
+        GSPSampleKey.solar_azimuth,
+        GSPSampleKey.solar_elevation
+    }
+
+    # REFERENCE
```
peterdudfield, line 40: remove

ocf_data_sampler/sample/base.py
```diff
@@ -0,0 +1,101 @@
...
+    # Supported array types / format(s) definition
+    VALID_ARRAY_TYPES = (np.ndarray, torch.Tensor, xr.DataArray)
+    SUPPORTED_FORMATS: ClassVar[set] = {'.pt'}
+
+    def __init__(self):
+        """ Initialise data container """
+        logger.debug("Initialising SampleBase instance")
+        self._data: Dict[str, Any] = {}
+    
+    def __getitem__(self, key: str) -> Any:
```
Sukh-P, line 36: For me the `__getitem__`, `___site_item__` and `keys` methods feel a bit much for functions which are one liners with dictionaries to implement, I know it runs the validation as well but I am not sure it is necessary currently

ocf_data_sampler/sample/base.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @abstractmethod
+    def save(self, path: Union[str, Path]) -> None:
+        """ Abstract method for saving sample data """
+        logger.debug(f"Save method called with path: {path}")
+        raise NotImplementedError("Save method must be implemented by subclass")
+
+    @classmethod
+    @abstractmethod
+    def load(cls, path: Union[str, Path]) -> 'SampleBase':
```
Sukh-P, line 98: I think there is an abstract method missing here, either called `to_model` or `to_numpy` or something similar which is essentially transforming into the preferred NumpySample type that the model can take (and get's tranformed into a tensor later)

ocf_data_sampler/sample/base.py
peterdudfield: I tend to agree, I wonder if we remove this for the moment?

ocf_data_sampler/sample/base.py
peterdudfield: Would just having `to_numpy` be fine for the moment? Then torch.dataloader, does the translating between numpy and a torch tensor

ocf_data_sampler/sample/uk_regional.py
peterdudfield: but, find to leave this as is at the moment

ocf_data_sampler/sample/uk_regional.py
```diff
@@ -0,0 +1,197 @@
...
+
+class PVNetSample(SampleBase):
+    """ Sample class specific to PVNet """
+    
+    # Feature space definitions
+    CORE_KEYS = {
+        'nwp',
+        GSPSampleKey.gsp,
+        GSPSampleKey.solar_azimuth,
+        GSPSampleKey.solar_elevation
```
peterdudfield, line 36: Would the GSP once be optional keys too? 
In live we wont need to load GSP at all

ocf_data_sampler/sample/base.py
```diff
@@ -0,0 +1,57 @@
...
+
+logger = logging.getLogger(__name__)
+
+class SampleBase(ABC):
+    """ 
+    Abstract base class for all sample types 
+    Provides core data storage functionality
+    """
+
+    def __init__(self):
```
peterdudfield, line 21: , should we add `data` to the _init_ so we can set `_data` when we create Sample straight away

tests/test_sample/test_uk_regional.py
```diff
@@ -0,0 +1,257 @@
...
+# PVNetUKRegionalDataset testing
+def test_pvnet_dataset(pvnet_config_filename):
+    """ Dataset initialisation """
+
+    # Temporal domain defined
+    start_time = "2024-01-01 00:00:00"
+    end_time = "2024-01-02 00:00:00"
+    
+    with pytest.raises(Exception):
+        dataset = PVNetUKRegionalDataset(
```
peterdudfield, line 203: check this raises an error, and if so add comment why

peterdudfield:
add matplotlib to requirements

ocf_data_sampler/sample/uk_regional.py
```diff
@@ -0,0 +1,188 @@
...
+        GSPSampleKey.solar_azimuth,
+        GSPSampleKey.solar_elevation,
+        SatelliteSampleKey.satellite_actual
+    }
+
+    def __init__(self):
+        logger.debug("Initialise PVNetSample instance")
+        super().__init__()
+
+    def validate(self) -> None:
```
dfulu, line 48: I'm not sure we should have a `validate()` method. There are no keys which are always present. Sometimes we'll want to train only satellite models with no NWP, sometimes only satellite, sometimes baseline models which just use solar coords etc. I don't think we should limit this since this limits the experiments we can do.

I think the rest of these checks belong in the pytests (I think most are already covered) rather than being checked at runtime

ocf_data_sampler/sample/uk_regional.py
```diff
@@ -0,0 +1,188 @@
...
+            
+            # Additional NWP-specific validations
+            nwp_sources = self._data[NWPSampleKey.nwp].keys()
+            for source in nwp_sources:
+                if not all(key in self._data[NWPSampleKey.nwp][source] 
+                           for key in ['nwp', NWPSampleKey.channel_names]):
+                    logger.error(f"Incomplete NWP data for source {source}")
+                    raise ValueError(f"Missing keys in NWP source {source}")
+        
+        # Validate timestep consistency
```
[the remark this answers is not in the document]
dfulu: Actually `GSPSampleKey.gsp` doesn't exist when we run this in production

ocf_data_sampler/sample/uk_regional.py
```diff
@@ -0,0 +1,188 @@
...
+            time_dependent_keys.append(SatelliteSampleKey.satellite_actual)
+        
+        for key in time_dependent_keys:
+            if len(self._data[key]) != gsp_timesteps:
+                logger.error(f"Validation failed - inconsistent timesteps for {key}")
+                raise ValueError("Inconsistent number of timesteps")
+                
+        logger.debug("PVNetSample validation successful")
+
+    def to_numpy(self) -> Dict[str, np.ndarray]:
```
dfulu, line 89: I think for this we should basically just wrap the functions [here](https://github.com/openclimatefix/ocf-data-sampler/tree/main/ocf_data_sampler/numpy_sample)

ocf_data_sampler/sample/uk_regional.py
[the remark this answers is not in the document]
peterdudfield: exacltly, so it should be optional

ocf_data_sampler/sample/uk_regional.py
peterdudfield: I like the idea of having a validate method on this. I agree you might not want to run it all the time in ML model experiments. But I like the idea that we can run it in production or even when making batches, to check things. 

Perhaps the validate method could run through the keys there are check for nans, for example.

dfulu: That is quite an expensive operation though to check for NaNs in all the inputs and would slow down batch creation. And if we were going to do this checking, I think it should go in the functions which create the sample rather than here. In the sampler we already have a [step to fill nans](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/torch_datasets/datasets/pvnet_uk_regional.py#L104) which would happen before the data got to this class. So there can never be NaNs here.

More generally I think this sample class should be a really light wrapper which basically just saves and loads. It should just be a convenience function basically

ocf_data_sampler/sample/uk_regional.py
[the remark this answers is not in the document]
dfulu: Oh I misread this. Good point

tests/test_sample/test_uk_regional.py
```diff
@@ -0,0 +1,257 @@
...
+
+def test_pvnet_dataset_get_sample(pvnet_config_filename):
+    """ Direct sample access - time and GSP ID """
+
+    # Empty spatial domain
+    with pytest.raises(Exception):
+        dataset = PVNetUKRegionalDataset(pvnet_config_filename)
+
+
+def test_pvnet_sample_to_numpy():
```
dfulu, line 218: This may be excessive testing since it basically tests whether the getitem and setitem into the internal dictionary work. I think that's simple enough that this isn't worthwhile

ocf_data_sampler/sample/uk_regional.py
peterdudfield: Maybe as a comprise lets remove it for the moment, and later add a function `validate_sample(Sample)` that can be called by who ever they want

ocf_data_sampler/sample/uk_regional.py
peterdudfield: just pass `data` back for the moment

felix-e-h-p:
Made some alterations - base.py removed validation and now have to_numpy as abstract. uk_regional.py removed validation function to outside class (and all related calls), to_numpy simplified also. Both tests also updated.

ocf_data_sampler/sample/base.py
```diff
@@ -0,0 +1,49 @@
...
+    Abstract base class for all sample types 
+    Provides core data storage functionality
+    """
+
+    def __init__(self, data: Optional[Dict[str, Any]] = None):
+        """ Initialise data container """
+        logger.debug("Initialising SampleBase instance")
+        self._data: Dict[str, Any] = data or {}
+
+    def keys(self):
```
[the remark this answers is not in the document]
felix-e-h-p: Both removed

felix-e-h-p:
Site inherited file pushed - core functionality is save loading in .nc format using h5netcdf, standard conversion of xr Dataset to numpy array and a placeholder for future plotting functionality

ocf_data_sampler/sample/base.py
```diff
@@ -0,0 +1,52 @@
...
+
+logger = logging.getLogger(__name__)
+
+class SampleBase(ABC):
+    """ 
+    Abstract base class for all sample types 
+    Provides core data storage functionality
+    """
+
+    # def __init__(self, data: Optional[Dict[str, Any]] = None):
```
peterdudfield, line 23: remove commented out code

ocf_data_sampler/sample/uk_regional.py
peterdudfield: or do NotImplementedYet warning/error

pyproject.toml
```diff
@@ -30,7 +30,9 @@ dependencies = [ # Migration from requirements.txt
     "pyproj",
     "pathy",
     "pyaml_env",
-    "pyresample"
+    "pyresample",
+    "matplotlib",
```
peterdudfield, line 34: can matplotlibbe put in a dev dependenices, as we dont always use it. We only need ti if we are plotting.

felix-e-h-p: Done

ocf_data_sampler/sample/base.py
felix-e-h-p: Done

ocf_data_sampler/sample/base.py
```diff
@@ -0,0 +1,52 @@
...
+    #     """ Initialise data container """
+    #     logger.debug("Initialising SampleBase instance")
+    #     self._data: Dict[str, Any] = data or {}
+
+    def __init__(self):
+        """ Initialise data container """
+        logger.debug("Initialising SampleBase instance")
+        self._data: Dict[str, Any] = {}
```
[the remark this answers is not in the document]
felix-e-h-p: Done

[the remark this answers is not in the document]
peterdudfield: i dont think data is always a dict

ocf_data_sampler/sample/uk_regional.py
peterdudfield: this comment @felix-e-h-p

peterdudfield (approved):

ocf_data_sampler/sample/__init__.py
```diff
@@ -0,0 +1,12 @@
+# __init__.py
```
Sukh-P, line 1: Small one but I think these comments at the top of the file with the file name can be removed

ocf_data_sampler/sample/base.py
```diff
@@ -0,0 +1,46 @@
...
+    def plot(self, **kwargs) -> None:
+        """ Abstract method for plotting """
+        raise NotImplementedError
+
+    @abstractmethod
+    def save(self, path: Union[str, Path]) -> None:
+        """ Abstract method for saving sample data """
+        raise NotImplementedError
+
+    @classmethod
```
Sukh-P, line 42: What's the reasoning around the load method being a classmethod?

ocf_data_sampler/sample/site.py
```diff
@@ -0,0 +1,98 @@
...
+            logger.error(f"Error saving - {path}: {str(e)}")
+            raise
+
+    @classmethod
+    def load(cls, path: Union[str, Path]) -> 'SiteSample':
+        """ Load site sample from netCDF """
+        logger.debug(f"Loading SiteSample from {path}")
+        path = Path(path)
+        
+        if path.suffix != '.nc':
```
Sukh-P, line 81: Nice check!

ocf_data_sampler/sample/site.py
```diff
@@ -0,0 +1,98 @@
...
+            logger.error(f"Invalid file format - {path.suffix}")
+            raise ValueError("Only .nc format is supported")
+        
+        try:
```
Sukh-P, line 85: It feels like this try/except block is a bit excessive perhaps, the only thing being called in it is xr.open_dataset() and I think the error logs from xarray would be sufficient for this if there was any issues with opening the the netcdf file, also given that there is already a check to see if the file is a necdf file I think that already reduces the risk of an error here that needs to be caught in the try/except

ocf_data_sampler/sample/__init__.py
felix-e-h-p: Removed

pyproject.toml
```diff
@@ -30,7 +30,8 @@ dependencies = [ # Migration from requirements.txt
     "pyproj",
     "pathy",
     "pyaml_env",
-    "pyresample"
+    "pyresample",
+    "h5netcdf",
```
Sukh-P, line 34: I am surprised this needs to be specified, is it not a dependency already in xarray?

ocf_data_sampler/sample/site.py
```diff
@@ -0,0 +1,98 @@
...
+        """ Convert sample numpy arrays - netCDF conversion """
+        logger.debug("Converting site sample to numpy format")
+        
+        try:
+            if not isinstance(self._data, xr.Dataset):
+                raise TypeError("Data must be xarray Dataset")
+            
+            numpy_data = convert_netcdf_to_numpy_sample(self._data)
+            
+            # Work around addition - to be checked!
```
Sukh-P, line 39: Think I am a bit confused around why this workaround is needed, would it be possible to explain, I thought that the `convert_netcdf_to_numpy_sample` will return a numpy sample (dictionary with numpy arrays) as required and that can be returned?

ocf_data_sampler/sample/uk_regional.py
```diff
@@ -0,0 +1,204 @@
...
+            logger.debug("Successfully created visualisation")
+        except Exception as e:
+            logger.error(f"Error creating visualisation: {str(e)}")
+            raise
+
+
+# TO DO - requirement to specify core and optional keys
+# Feature space definition required for validation implementation
+# Currently depreciated as of the moment
+def validate(data: Dict[str, Any], core_keys: set, optional_keys: set) -> None:
```
Sukh-P, line 142: For the sake of keeping the codebase clean I would remove this method for now as it's stated it is not used, not a fan of including code that might be used, better to remove and then add in if/when it is used IMHO

ocf_data_sampler/sample/base.py
felix-e-h-p: Wasn't sure to leave in but I think probably best to as of the moment. So with load creating an instance of a subclass from a file for example, class method allows it to operate at class level. So pretty much always creates an instance without needing to know the exact type of the subclass - kind of like an extra assurance so to speak whereby loading logic can dynamically work with the subclass. Also, if a new subclasses is introduced w/ unique loading logic / different format, it acts to never affect the base class or other subclasses.

tests/test_sample/test_site_sample.py
```diff
@@ -0,0 +1,195 @@
...
+import pandas as pd
+
+from pathlib import Path
+
+from ocf_data_sampler.torch_datasets.datasets.site import SitesDataset
+from ocf_data_sampler.sample.site import SiteSample
+
+
+@pytest.fixture
+def sample_with_data():
```
Sukh-P, line 20: Any FYI but I think there would be an easier way to create a site sample, as is done in this test [here](https://github.com/openclimatefix/ocf-data-sampler/blob/main/tests/torch_datasets/test_site.py#L10)

ocf_data_sampler/sample/uk_regional.py
```diff
@@ -0,0 +1,204 @@
...
+
+from ocf_data_sampler.numpy_sample import (
+    NWPSampleKey, 
+    GSPSampleKey,
+    SatelliteSampleKey
+)
+
+from ocf_data_sampler.sample.base import SampleBase
+
+try:
```
Sukh-P, line 30: This bit is duplicate of above block I think

ocf_data_sampler/sample/uk_regional.py
```diff
@@ -0,0 +1,204 @@
...
+    def __init__(self):
+        logger.debug("Initialise UKRegionalSample instance")
+        super().__init__()
+        self._data = {}
+
+    # TO DO - complete
+    def to_numpy(self) -> Dict[str, Any]:
+        """ Placeholder for numpy conversion """
+        logger.debug("to_numpy method not yet implemented")
+        raise NotImplementedError("Numpy conversion method is not yet implemented")
```
Sukh-P, line 53: I think we want to return self._data here no? We want to have the same data returned except it is already in the format needed so no transformation needed

ocf_data_sampler/sample/site.py
felix-e-h-p: Sorted

ocf_data_sampler/torch_datasets/datasets/__init__.py
```diff
@@ -1,2 +1,13 @@
 from .pvnet_uk_regional import PVNetUKRegionalDataset
-from .site import SitesDataset
+
+from .site import (
+    convert_netcdf_to_numpy_sample,
+    convert_from_dataset_to_dict_datasets,
```
Sukh-P, line 5: Is this needed? Not sure if it is imported anywhere in these changes, I may have missed it though

ocf_data_sampler/sample/site.py
```diff
@@ -0,0 +1,98 @@
...
+                engine="h5netcdf"
+            )
+            logger.debug(f"Successfully saved SiteSample - {path}")
+            
+        except Exception as e:
+            logger.error(f"Error saving - {path}: {str(e)}")
+            raise
+
+    @classmethod
+    def load(cls, path: Union[str, Path]) -> 'SiteSample':
```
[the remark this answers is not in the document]
felix-e-h-p: Done

ocf_data_sampler/sample/uk_regional.py
felix-e-h-p: Currently updated to return _data as specified

Sukh-P (approved):
Thanks for all the work on this!

PR #128, opened 2025-01-17
Remove Unused Parameters and Functions for Cleaner Dataset Slicing Logic

### LINKED ISSUE #107: Remove the `max_steps_gap` parameter and missing timestamp functionality in `select_time_slice()`
In the `slice_datasets_by_time()` function, the call to `select_time_slice()` used for the satellite data uses `max_steps_gap=2` - [see here](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/select/time_slice_for_dataset.py#L54). 

The value of `max_steps_gap` is entirely ignored in this function. This parameter is a [carry over from ocf_datapipes](https://github.com/openclimatefix/ocf_datapipes/blob/main/ocf_datapipes/select/select_time_slice.py#L57) and the slicing logic which uses this setting has not been ported across yet. Quite possibly we will never want to use this functionality since this was intended for use only in a production setting where data might be missing. We [do this in the app](https://github.com/openclimatefix/uk-pvnet-app/blob/main/pvnet_app/data/satellite.py#L159) now instead. During training, since we filter to t0 times where no satellite data may be missing, our current usage would never hit this anyway. We do not have the functionality in our t0 filtering to allow any missing time steps, so we should never need this.

I suggest we:

  - Strip out this parameter and the placeholder `_sel_fillinterp()` function which it was intended to be used with. 
  - Plus remove the `_sel_fillnan()` function and the `fill_selection` parameter. I don't think we use these or will ever use these for training.

Sachin-G13:
Fixes #107 

## Description

This PR removes redundant parameters and unused functionality from the dataset slicing logic to simplify and optimize the codebase.
 The max_steps_gap parameter was removed from the select_time_slice() call in the satellite data section of slice_datasets_by_time().
Also, both _sel_fillinterp() and _sel_fillnan() functions were removed.

## How Has This Been Tested?

Verified that dataset slicing works as expected with the updated logic.

- [x] Yes

## Checklist:

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,7 +1,7 @@
 [bumpversion]
 commit = True
 tag = True
-current_version = 0.0.47
+current_version = 0.0.48
 message = Bump version: {current_version} → {new_version} [skip ci]
 
 [bumpversion:file:pyproject.toml]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "ocf_data_sampler"
-version = "0.0.47"
+version = "0.0.48"
 license = { file = "LICENSE" }
 readme = "README.md"
 description = "Sample from weather data for renewable energy prediction"


ocf_data_sampler/select/select_time_slice.py
```diff
@@ -2,19 +2,6 @@
...
-def _sel_fillnan(
-        da: xr.DataArray, 
-        start_dt: pd.Timestamp, 
-        end_dt: pd.Timestamp, 
-        sample_period_duration: pd.Timedelta,
-    ) -> xr.DataArray:
-    """Select a time slice from a DataArray, filling missing times with NaNs."""
-    requested_times = pd.date_range(start_dt, end_dt, freq=sample_period_duration)
-    return da.reindex(time_utc=requested_times)
-
```
AUdaltsova, line 16: Good spot that this is not actually implemented! @dfulu just to double-check that we are ok to remove the infill options? I guess we could implement it as well but I think that'd be quite a bit of work to do properly (if I remember it correctly, t0 selection won't let a slice that needs filling through for example + we'll need user control over `fill_selection`, etc).

ocf_data_sampler/select/select_time_slice.py
```diff
@@ -24,18 +11,6 @@ def _sel_default(
...
-        da: xr.DataArray, 
-        start_dt: pd.Timestamp, 
-        end_dt: pd.Timestamp, 
-        sample_period_duration: pd.Timedelta,
-    ) -> xr.DataArray:
-    """Select a time slice from a DataArray, filling missing times with linear interpolation."""
-    return NotImplemented
-
-
 def select_time_slice(
```
AUdaltsova, line 14: If we're removing fill options, I think `fill_selection `and `max_step_gap` should also be stripped here, in uses, and, importantly, tests, as these parameters are only needed for selecting fill options.

ocf_data_sampler/select/select_time_slice.py
```diff
@@ -99,32 +66,23 @@ def select_time_slice_nwp(
...
     ).values
 
-    # Find the required steps for all target times
     steps = target_times - selected_init_times
-    
-    # We want one timestep for each target_time_hourly (obviously!) If we simply do
-    # nwp.sel(init_time=init_times, step=steps) then we'll get the *product* of
-    # init_times and steps, which is not what # we want! Instead, we use xarray's
-    # vectorized-indexing mode by using a DataArray indexer.  See the last example here:
-    # https://docs.xarray.dev/en/latest/user-guide/indexing.html#more-advanced-indexing
```
AUdaltsova, line 127: I would ask you to add all the doc strings and comments back in. They're very useful for new people trying to make sense of the code, and some of them, like this one, add very important context as to why it's written the way it's written, which would otherwise take a while to dig up!

AUdaltsova (changes requested):
I'd hold fire on the changes for a bit just until we're sure this is ok to remove, so you don't do unnecessary work. Thanks a lot for doing this!

AUdaltsova:
Hi again! Sorry, I missed that this was for an existing issue, so yeah, obviously these functions are fine to delete. Thanks a lot for pitching in!
There is another PR on that same issue which is getting merged now; it strips out `_sel_fillinterp` and `max_steps_gap`, but not `_sel_fillnan`, which I think is important to do and you PR covers. A couple of things I think should be done before we can merge:
- add the comments back in
- since this will strip out all `_sel` options, I think this can be simplified further and just do the default selection without hiding it under a function; this also means `fill_selection` can be removed
- as you already know, there is a test that depends on these options, so it will have to be adjusted, or even removed, if there is another one already testing for the default functionality

Let me know if you have any doubts/questions, and feel free to tag me to re-review when you're ready! Thanks again for your help!

PR #131, opened 2025-01-17
Site dataset clean

Sukh-P:
# Pull Request

## Description

This PR includes:

- Add id sample key for sites
- Remove some duplicated code
- Reorder functions in Site file slightly to make main function for loading presaved samples more obvious


ocf_data_sampler/torch_datasets/site.py
```diff
@@ -257,6 +257,8 @@ def process_and_combine_site_sample_dict(
         )
         combined_sample_dataset = xr.merge([combined_sample_dataset, sun_position_features_xr])
 
+        # TODO include t0_index in xr dataset?
```
AUdaltsova, line 260: Not sure what that means?

ocf_data_sampler/torch_datasets/site.py
```diff
@@ -317,6 +319,26 @@ def merge_data_arrays(self, normalised_data_arrays: list[Tuple[str, xr.DataArray
 
 # ----- functions to load presaved samples ------
 
+def convert_netcdf_to_numpy_sample(ds: xr.Dataset) -> dict:
```
AUdaltsova, line 322: Is there a reason this function got moved up? Maybe I'm missing something, but intuitively I'd think since it uses `convert_from_dataset_to_dict_datasets` it should go after it?

Sukh-P: It's the main function in this section so it felt natural for it to be first, I often find it more intuitive to have the main function first and then all the functions it calls afterwards

ocf_data_sampler/torch_datasets/site.py
Sukh-P: We sometimes include a t0_index in our samples like here: https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/numpy_sample/site.py#L28 to be able to actually pass that we would need to save that in the xarray Dataset somewhere, it isn't needed but I was unsure of its importance so left a note here

ocf_data_sampler/torch_datasets/site.py
AUdaltsova: ok!

ocf_data_sampler/torch_datasets/site.py
AUdaltsova: Ok if this is what I think it is, I think we usually store it as a variable attr? I'm actually not sure why the function you've linked (thanks btw!) expects it to be passed separately, as far as I can see they're all organised like this, instead of how it used to be (e g here https://github.com/openclimatefix/ocf_datapipes/blob/371b2bea2153d84834780fc056cd0d2410624a0b/ocf_datapipes/convert/numpy_batch/pv.py#L18), so I must've missed some decision on how we store it now. Do we at all?..

upd: found a use of t0_idx for gsp here https://github.com/openclimatefix/ocf-data-sampler/blob/ce50c5219bf3be398411f98e169d69d8dec411fd/ocf_data_sampler/torch_datasets/datasets/pvnet_uk_regional.py#L74
but I think it's the only source that does this currently, which is odd?

Sukh-P: I think it is there for GSP because that is a target variable and it's useful to have it for inference later, like [here](https://github.com/openclimatefix/PVNet/blob/main/scripts/backtest_uk_gsp.py#L246), so it may be useful to include for Sites as well based on the same reasoning, we can circle back to this, won't be included in this PR anyway, but I can add an issue to discuss this and whether it should be added / how we should be adding/storing it

AUdaltsova: re: its importance, I think it might be used somewhere to see where pv history ends and pv truth begins? Definitely in the backtest [here](https://github.com/openclimatefix/ocf-data-sampler/blob/ce50c5219bf3be398411f98e169d69d8dec411fd/ocf_data_sampler/torch_datasets/datasets/pvnet_uk_regional.py#L74), but probably somewhere else in pvnet for the same reason. The gsp one is used in this [metric function](https://github.com/openclimatefix/PVNet/blob/a3c225141f81f02d48effe982fd67eabddcfea05/pvnet/utils.py#L328), but I don't see anything like this that uses site t0_idx

AUdaltsova: Yep agreed. Generally I think we should also either set this for other input sources or remove the key for them, as right now as far as I can see we don't use it for nwp and sat (though I would be in favor of adding it to the .nc samples, I often check it when debugging)

AUdaltsova (approved):
LGTM! Thanks for doing this

ocf_data_sampler/torch_datasets/site.py
AUdaltsova: btw, while we are talking about this, we used to have [this helper function in dpipes](https://github.com/openclimatefix/ocf_datapipes/blob/371b2bea2153d84834780fc056cd0d2410624a0b/ocf_datapipes/transform/xarray/add_t0idx_and_sample_period_duration.py#L14) that added t0_idx as an attr to everything. Might be a good thing to have

peterdudfield (approved):

PR #132, opened 2025-01-18
Refactoring time slice selection

### LINKED ISSUE #107: Remove the `max_steps_gap` parameter and missing timestamp functionality in `select_time_slice()`

RupeshMangalam21:
# Pull Request

## Description
Removes unused code and parameters from time slice selection functionality to improve code maintainability.

Fixes #107 

The changes focus on removing unused functionality while preserving core features validated by the test suite. This includes removing the unimplemented `_sel_fillinterp()` function and unused `max_steps_gap` parameter, while retaining the essential `fill_selection` parameter needed for handling out-of-bounds timestamps.

Simplifying the selection logic since we no longer need to handle interpolation.

## How Has This Been Tested?
The full test suite has been executed to verify that the removal of unused code does not impact existing functionality. Particular attention was paid to tests involving out-of-bounds timestamp handling to ensure the retained `fill_selection` parameter continues to function correctly.

- [x] Yes

The changes do not affect data processing outputs, so no additional plotting verification was required.

- [x] N/A

## Checklist:
- [x] My code follows OCF's coding style guidelines

### DIFF
--- a/ocf_data_sampler/select/select_time_slice.py
+++ b/ocf_data_sampler/select/select_time_slice.py
@@ -3,7 +3,6 @@ import pandas as pd
 import numpy as np
 
 
-
 def _sel_fillnan(
         da: xr.DataArray, 
         start_dt: pd.Timestamp, 
@@ -25,17 +24,6 @@ def _sel_default(
     return da.sel(time_utc=slice(start_dt, end_dt))
 
 
-# TODO either implement this or remove it, which would tidy up the code
-def _sel_fillinterp(
-        da: xr.DataArray, 
-        start_dt: pd.Timestamp, 
-        end_dt: pd.Timestamp, 
-        sample_period_duration: pd.Timedelta,
-    ) -> xr.DataArray:
-    """Select a time slice from a DataArray, filling missing times with linear interpolation."""
-    return NotImplemented
-
-
 def select_time_slice(
     ds: xr.DataArray,
     t0: pd.Timestamp,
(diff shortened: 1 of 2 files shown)


dfulu (approved):
Thanks for these changes @RupeshMangalam21

PR #145, opened 2025-01-23
Add GFS Statistics to constants.py

siddharth7113:
# **Pull Request**

## **Description**

_Added GFS mean and standard deviation (STD) values to `constants.py` for integration with the new repository. Enhanced the `NWP_PROVIDERS` list and `NWPStatDict` to include GFS support._

**Key Changes:**
- Introduced `GFS_MEAN` and `GFS_STD` values, formatted using `_to_data_array`.
- Updated `NWP_PROVIDERS` to include `"gfs"`.
- Extended `NWP_MEANS` and `NWP_STDS` in `NWPStatDict` to handle GFS data.
- Maintained consistent structure and formatting with existing constants like UKV and ECMWF.
---
#### **Checklist**

- [x] My code follows [[OCF's coding style guidelines](https://github.com/openclimatefix/.github/blob/main/coding_style.md)](https://github.com/openclimatefix/.github/blob/main/coding_style.md).
- [x] I have performed a self-review of my own code.
- [ ] I have made corresponding changes to the documentation (if applicable).
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [x] I have checked my code and corrected any misspellings.


siddharth7113:
@Sukh-P ,
@peterdudfield 
Hi , I am working in `open-data-pvnet` which requeries working with `gfs-data` , Would it be possible to review the PR and merge it, if everything seems correct?

Sukh-P:
@siddharth7113 the code change looks good, just one thing to mention/look out for is that the value ranges of the data looks roughly equivalent after normalisation if you are using multiple NWP providers, I say this because the ECMWF normalisation constants for example have been calculated for the UK specifically, where as I am unsure whether the same has been done for the GFS constants (could have been done globally or for another region), I am going to create an issue/discussion as I think we may want to move to making the normalisation constants more specific for regions and then have some way of controlling which ones are used

Sukh-P (approved):

siddharth7113:
> @siddharth7113 the code change looks good, just one thing to mention/look out for is that the value ranges of the data looks roughly equivalent after normalisation if you are using multiple NWP providers, I say this because the ECMWF normalisation constants for example have been calculated for the UK specifically, where as I am unsure whether the same has been done for the GFS constants (could have been done globally or for another region), I am going to create an issue/discussion as I think we may want to move to making the normalisation constants more specific for regions and then have some way of controlling which ones are used

Yeah, it would be best to discuss them , I tried normalization using ecmwf [here](https://github.com/openclimatefix/open-data-pvnet/issues/1#issuecomment-2606028057) for some attributes and found some discripenacies, As for GFS values, I simply copied them from [here](https://github.com/openclimatefix/ocf_datapipes/blob/main/ocf_datapipes/utils/consts.py#L192)

PR #156, opened 2025-02-06
Minimal implementations of NumpySample and TensorSample

felix-e-h-p:
Basic implementation as of the moment - both new classes inherit from SampleBase in sample/base.py

Both NumpySample and TensorSample only implement required methods and utilise parent class without additional validation as of the moment

Move from enum keys to string keys - support for same data types and conversion between formats (kept type definitions similar)

### DIFF
--- a/ocf_data_sampler/sample/base.py
+++ b/ocf_data_sampler/sample/base.py
@@ -5,12 +5,18 @@ Handling of both flat and nested structures - consideration for NWP
 
 import logging
 import numpy as np
+import torch
 
 from pathlib import Path
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional, Union, TypeAlias
 from abc import ABC, abstractmethod
 
+
 logger = logging.getLogger(__name__)
+
+NumpyBatch: TypeAlias = Dict[str, Union[np.ndarray, Dict[str, np.ndarray]]]
+TensorBatch: TypeAlias = Dict[str, Union[torch.Tensor, Dict[str, torch.Tensor]]]
+
 
 class SampleBase(ABC):
     """ 
@@ -18,9 +24,16 @@ class SampleBase(ABC):
     Provides core data storage functionality
     """
 
-    def __init__(self):
+    def __init__(self, data: Optional[Dict[str, Any]] = None):
         """ Initialise data container """
         logger.debug("Initialising SampleBase instance")
+        self._data = data if data is not None else {}
+
+    def __getitem__(self, key: str) -> Any:
+        return self._data[key]
+    
+    def __setitem__(self, key: str, value: Any) -> None:
+        self._data[key] = value
 
     @abstractmethod
     def to_numpy(self) -> Dict[str, Any]:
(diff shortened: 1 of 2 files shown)


ocf_data_sampler/sample/base.py
```diff
@@ -5,22 +5,35 @@
...
+
+
 class SampleBase(ABC):
     """ 
     Abstract base class for all sample types 
     Provides core data storage functionality
     """
 
-    def __init__(self):
+    def __init__(self, data: Optional[Dict[str, Any]] = None):
```
Sukh-P, line 27: Nice idea to add it to the init I feel that makes sense, it can be an xarray DataSet as well though

ocf_data_sampler/sample/base.py
```diff
@@ -5,22 +5,35 @@
...
         """ Initialise data container """
         logger.debug("Initialising SampleBase instance")
+        self._data = data if data is not None else {}
```
Sukh-P, line 30: Is this not enough here?:
`self._data = data`

ocf_data_sampler/sample/base.py
```diff
@@ -42,3 +55,92 @@ def save(self, path: Union[str, Path]) -> None:
...
+
+    def save(self, path: Union[str, Path]) -> None:
+        raise NotImplementedError
+
+    @classmethod
+    def load(cls, path: Union[str, Path]) -> 'TensorSample':
+        raise NotImplementedError
+
+
+def sample_to_tensor(sample: NumpySample) -> TensorSample:
```
Sukh-P, line 118: I think these should still be called batch_to_tensor since when we call this the samples are already stacked into a batch

ocf_data_sampler/sample/base.py
```diff
@@ -42,3 +55,92 @@ def save(self, path: Union[str, Path]) -> None:
     def load(cls, path: Union[str, Path]) -> 'SampleBase':
         """ Abstract class method for loading sample data """
         raise NotImplementedError
+
+
+class NumpySample(SampleBase):
```
Sukh-P, line 60: For me I don't think we need this class or the TensorSample, I think the batch_to_tensor function by itself for now is fine and can be imported in to be used in PVNet, I don't quite see where these would be used

ocf_data_sampler/sample/base.py
```diff
@@ -5,22 +5,35 @@
...
     Provides core data storage functionality
     """
 
-    def __init__(self):
+    def __init__(self, data: Optional[Dict[str, Any]] = None):
         """ Initialise data container """
         logger.debug("Initialising SampleBase instance")
+        self._data = data if data is not None else {}
+
+    def __getitem__(self, key: str) -> Any:
```
Sukh-P, line 32: I wouldn't include these methods as mentioned in a previous PR I think they are too simple for their own methods

Sukh-P (changes requested):
Thanks for making a start on this, I think this can be simplified a bit more

felix-e-h-p:
Cheers for the feedback - made relevant updates. Reduced the helper/wrapper pattern also for batch_to_tensor and have some pretty basic testing implemented for it.

ocf_data_sampler/sample/base.py
```diff
@@ -5,22 +5,33 @@
...
 
 from pathlib import Path
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional, Union, TypeAlias
 from abc import ABC, abstractmethod
 
+
 logger = logging.getLogger(__name__)
 
+NumpyBatch: TypeAlias = Dict[str, Any]
```
Sukh-P, line 18: Wondering what your thoughts are on this vs the more specific type alias's we had in ocf_datapipes that could be used here

ocf_data_sampler/sample/base.py
```diff
@@ -5,22 +5,33 @@
...
     Provides core data storage functionality
     """
 
-    def __init__(self):
+    def __init__(self, data: Optional[Dict[str, Any] | xr.Dataset] = None):
         """ Initialise data container """
         logger.debug("Initialising SampleBase instance")
+        self._data = data
+    
+    def __setitem__(self, key: str, value: Any) -> None:
```
Sukh-P, line 33: Do we still need this `__setitem__` method?

ocf_data_sampler/sample/base.py
Sukh-P: Also perhaps this should be NumpySample rather than Batch, in my head NumpySamples and TensorBatches are the things we care more about

ocf_data_sampler/sample/base.py
```diff
@@ -5,22 +5,33 @@
...
+        self._data[key] = value
 
     @abstractmethod
     def to_numpy(self) -> Dict[str, Any]:
```
Sukh-P, line 37: If we have NumpySample this could be used here 
```suggestion
    def to_numpy(self) -> NumpySample:
```

ocf_data_sampler/sample/base.py
Sukh-P: Or we have TypeAlias's for all three, perhaps that is clearer, NumpySample and NumpyBatch are essentially the same except there is an extra dimension in the values of the dictionary in the Batch case

tests/test_sample/test_base.py
```diff
@@ -84,3 +87,69 @@ def test_sample_base_to_numpy():
...
+    """ Test nested dictionary conversion """
+    batch = {
+        'outer': {
+            'inner': np.array([1, 2, 3])
+        }
+    }
+    tensor_batch = batch_to_tensor(batch)
+    
+    assert isinstance(tensor_batch['outer']['inner'], torch.Tensor)
+    assert torch.equal(tensor_batch['outer']['inner'], torch.tensor([1, 2, 3]))
```
Sukh-P, line 102: Is the first check redundant given this second one only passes if it is a torch.Tensor?

tests/test_sample/test_base.py
```diff
@@ -84,3 +87,69 @@ def test_sample_base_to_numpy():
...
+    }
+    tensor_batch = batch_to_tensor(batch)
+    
+    assert isinstance(tensor_batch['tensor_data'], torch.Tensor)
+    assert isinstance(tensor_batch['string_data'], str)
+    assert isinstance(tensor_batch['nested']['numbers'], torch.Tensor)
+    assert isinstance(tensor_batch['nested']['text'], str)
+
+
+def test_batch_to_tensor_empty():
```
Sukh-P, line 123: Not quite sure about this test, do we ever expect an empty batch to be passed and if so is this test showing the correct behaviour?

ocf_data_sampler/sample/base.py
felix-e-h-p: Yeah, definitely - all three considered with most recent push

tests/test_sample/test_base.py
felix-e-h-p: Yeah - just being thorough I suppose. Removed.

tests/test_sample/test_base.py
felix-e-h-p: Wasn't sure about this RE reduced exception raising - have a ValueError implemented in batch_to_tensor instead.

Sukh-P (approved):

PR #167, opened 2025-02-12
Norm constants update

felix-e-h-p:
Variable naming updated in constants.py - new mean and std values for sd/sde following calculations (2018-23 zarrs)

Error handling added to process_and_combine_datasets (pvnet_uk.py) for NWP and sat - validation function situated in utils - relevant tests included for the update

Noticed the TO DO in process_and_combine_site_sample_dict (site.py) regarding normalisation for sat - have this included alongside validation and necessary testing updates


ocf_data_sampler/torch_datasets/datasets/site.py
```diff
@@ -223,12 +224,15 @@ def process_and_combine_site_sample_dict(
...
+                validate_channels(da_nwp, self.config, "nwp", provider)
                 da_nwp = (da_nwp - NWP_MEANS[provider]) / NWP_STDS[provider]
                 data_arrays.append((f"nwp-{provider}", da_nwp))
             
         if "sat" in dataset_dict:
-            # TODO add some satellite normalisation
+            # Standardise
             da_sat = dataset_dict["sat"]
+            validate_channels(da_sat, self.config, "sat")
+            da_sat = (da_sat - RSS_MEAN) / RSS_STD
```
peterdudfield, line 235: i'd put this in a different PR, to keep things simple

ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
```diff
@@ -47,12 +49,14 @@ def process_and_combine_datasets(
...
 
     if "nwp" in dataset_dict:
-
         nwp_numpy_modalities = dict()
 
         for nwp_key, da_nwp in dataset_dict["nwp"].items():
-            # Standardise
+            # Validate channels before standardisation
             provider = config.input_data.nwp[nwp_key].provider
+            validate_channels(da_nwp, config, "nwp", provider)
```
peterdudfield, line 57: I might have mis understood, but weren't we going to validate that `NWP_MEANS` has the correct channels in it. In the bug with `sde` / `sd` it did not and we lost and channel without knowing. 

So we could have something like 
```
assert set(da_nwp .channels) in set(NWP_MEANS[provider].keys())
```
(I havent tested this out, this is just a guess, but hopefully it helps)

peterdudfield: For checking the channels are in the data, we should do that when the channels are selected, this would be in [load](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/load/load_dataset.py#L32), but this might already cover it

ocf_data_sampler/torch_datasets/datasets/site.py
```diff
@@ -220,6 +224,8 @@ def process_and_combine_site_sample_dict(
         data_arrays = []
 
         if "nwp" in dataset_dict:
+            validate_nwp_channels(self.config, NWP_MEANS, NWP_STDS)
+
             for nwp_key, da_nwp in dataset_dict["nwp"].items():
                 # Standardise
```
peterdudfield, line 230: Could you not just call `_validate_channels` here? 

```
validate_channels(
            data_channels= da_nwp.channels.values,
            means_channels=nwp_means[nwp_key].channel.values,
            stds_channels=nwp_stds[nwp_key].channel.values,
        )
```

And then this function could be the same for `nwp` and `satellitle`

felix-e-h-p: Thanks for the feedback - just made the relevant changes

ocf_data_sampler/torch_datasets/utils/validate_channels.py
```diff
@@ -0,0 +1,46 @@
...
+            
+    # Find missing channels in stds
+    missing_in_stds = data_channels - stds_channels
+    if missing_in_stds:
+        raise ValueError(
+            f"The following channels for {source_name} are missing in normalisation stds: "
+            f"{missing_in_stds}"
+        )
+            
+    # Check if any extra channels in constants that are not in data
```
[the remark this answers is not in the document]
felix-e-h-p: Updated

tests/torch_datasets/test_validate_channels_utils.py
```diff
@@ -0,0 +1,70 @@
...
+            "norm_channels": ["IR_016"],
+            "source_name": "satellite",
+            "expect_error": True,
+            "error_match": "following channels for satellite are missing in normalisation means"  # Updated to match actual error
+        }
+    ])
+    def test_channel_validation(self, test_case, create_mock_xr_array):
+        """Test channel validation for both NWP and satellite data"""
+        data_channels = set(test_case["data_channels"])
+        norm_array = create_mock_xr_array(test_case["norm_channels"])
```
peterdudfield, line 52: could this be `norm_array = set(test_case["norm_channels"])`

peterdudfield: and then remove `create_mock_xr_array`

tests/torch_datasets/test_validate_channels_utils.py
```diff
@@ -0,0 +1,70 @@
+"""Tests - channel validation utility function"""
+
+import pytest
+import xarray as xr
+
+from ocf_data_sampler.torch_datasets.utils.validate_channels import validate_channels
+
+
+@pytest.fixture
```
peterdudfield, line 9: try to remove this

ocf_data_sampler/torch_datasets/utils/validate_channels.py
```diff
@@ -0,0 +1,35 @@
...
+
+    Args:
+        data_channels: Set of channels from the data
+        means_channels: Set of channels from means constants
+        stds_channels: Set of channels from stds constants
+        source_name: Name of data source (e.g., 'ecmwf', 'satellite') for error messages
+
+    Raises:
+        ValueError: If there's a mismatch between data channels and normalisation constants
+    """
```
peterdudfield, line 20: move `set` into this function. Then the inputs are just lists

felix-e-h-p: All sorted, cheers!

ocf_data_sampler/torch_datasets/datasets/site.py
```diff
@@ -221,8 +222,17 @@ def process_and_combine_site_sample_dict(
 
         if "nwp" in dataset_dict:
             for nwp_key, da_nwp in dataset_dict["nwp"].items():
-                # Standardise
                 provider = self.config.input_data.nwp[nwp_key].provider
+                
+                # Validate channels
+                validate_channels(
```
dfulu, line 228: I think these checks would be better in the data config pydantic checks. @peterdudfield what do you think?

ocf_data_sampler/constants.py
```diff
@@ -86,7 +86,7 @@ def __getitem__(self, key):
     "lcc": 0.3791404366493225,
     "mcc": 0.38039860129356384,
     "prate": 9.81039775069803e-05,
-    "sde": 0.000913831521756947,
+    "sd": 0.0011230289,
```
dfulu, line 89: We need to be careful in deploying this. This will break the compatibility to any previously trained models which use these constants

ocf_data_sampler/torch_datasets/datasets/site.py
peterdudfield: What so, when the Configuration is loaded, we go and check that they variables are in `constant.py` as a validation method on channels?

ocf_data_sampler/constants.py
Sukh-P: Could we not include both for now instead of renaming this one to try and get around this?

peterdudfield: yea both, with a comment, saying we are trying to use "sd"  for the future

ocf_data_sampler/torch_datasets/datasets/site.py
dfulu: Yeh something like that. Or even just check it in the init of the dataset

ocf_data_sampler/constants.py
dfulu: I think that keeping both will likely require care as well and could cause some really hard to trace forecast issues. In the app this variable is [assigned the name sde](https://github.com/openclimatefix/uk-pvnet-app/blob/main/pvnet_app/data/nwp.py#L202) which makes me wonder why we would want to change it here at all? If it has been assigned a different name in our training data can we not just correct that in the trainning data?

If we keep both variable names and we want to maintain backwards compatibility to already trained models then we'd need to add logic to the app to duplicate the sde/sd channel so it exists under both names in the data. That feels messy to me.

ocf_data_sampler/constants.py
```diff
@@ -110,7 +110,7 @@ def __getitem__(self, key):
     "lcc": 0.44901806116104126,
     "mcc": 0.3288780450820923,
     "prate": 3.108070450252853e-05,
-    "sde": 8.107526082312688e-05,
+    "sd": 0.0006500509,
```
peterdudfield, line 113: Change to "sd" but keep old value. This means PVnet DA model will still work. Retrain Pvnet INtraday using this values.

ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
```diff
@@ -168,6 +166,36 @@ def get_gsp_locations(gsp_ids: list[int] | None = None) -> list[Location]:
...
+        config: Configuration object containing provider information
+        
+    Raises:
+        ValueError: If there's a mismatch between data channels and normalisation constants
+    """
+    if "nwp" in datasets_dict:
+        for nwp_key, da_nwp in datasets_dict["nwp"].items():
+            provider = config.input_data.nwp[nwp_key].provider
+            validate_channels(
+                data_channels=set(da_nwp.channel.values),
```
peterdudfield, line 183: can you remove `set` as this is done in other functions. Similar for other calls of `validate_channels`

peterdudfield (changes requested):
One comment, other than that it looks good

ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
```diff
@@ -186,9 +214,11 @@ def __init__(
         """
         
         config = load_yaml_configuration(config_filename)
-
```
peterdudfield, line 189: Could you do it directly after loading the config, and not use dataset_dict. 
Inside the function you can use `config.input_data.satellite.channels` and similar for nwp. 
This would mean it would catch the error earlier, and is independent from `dataset_dict`

ocf_data_sampler/torch_datasets/datasets/site.py
```diff
@@ -47,6 +51,18 @@ def __init__(
         """
 
         config: Configuration = load_yaml_configuration(config_filename)
+
+        # Validate NWP channels if NWP is configured
+        if hasattr(config.input_data, "nwp"):
```
peterdudfield, line 56: how come you didnt check satellite here?

ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
```diff
@@ -168,6 +166,36 @@ def get_gsp_locations(gsp_ids: list[int] | None = None) -> list[Location]:
     return locations
 
 
+def validate_channels_from_config(config: Configuration) -> None:
```
peterdudfield, line 169: should this be moved to validate_channels.py?

ocf_data_sampler/torch_datasets/datasets/site.py
peterdudfield: you could use the function above `validate_channels_from_config`

ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
```diff
@@ -168,6 +166,36 @@ def get_gsp_locations(gsp_ids: list[int] | None = None) -> list[Location]:
...
+
+    # Validate satellite channels if satellite is configured
+    if hasattr(config.input_data, "satellite"):
+        validate_channels(
+            data_channels=config.input_data.satellite.channels,
+            means_channels=RSS_MEAN.channel.values,
+            stds_channels=RSS_STD.channel.values,
+            source_name="satellite"
+        )
+
```
peterdudfield, line 197: Refactor to two functions
- `validate_nwp_channels_from_config`
- `validate_satellite_channels_from_config`
then you can use `validate_nwp_channels_from_config`

ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
```diff
@@ -31,10 +31,16 @@
     merge_dicts,
     fill_nans_in_arrays,
 )
+from ocf_data_sampler.torch_datasets.utils.validate_channels import (
+    validate_channels,
+    validate_nwp_channels,
+    validate_satellite_channels,
+)
```
peterdudfield, line 38: do you use the function `validate_channels`?

ocf_data_sampler/torch_datasets/datasets/site.py
```diff
@@ -19,6 +20,10 @@
 from ocf_data_sampler.utils import minutes
 from ocf_data_sampler.torch_datasets.utils.valid_time_periods import find_valid_time_periods
 from ocf_data_sampler.torch_datasets.utils.merge_and_fill_utils import merge_dicts, fill_nans_in_arrays
+from ocf_data_sampler.torch_datasets.utils.validate_channels import (
+    validate_channels,
```
peterdudfield, line 24: do you use `validate_channels`?

peterdudfield (approved):

PR #182, opened 2025-02-21
Implements ruff linting, template workflows for versioning

devsjc:
Sets a consistent style for the code in the repo according to the OCF Style Guidelines. Updates workflows to use the new templates. (James I know you don't like full stops in docstrings - blame whoever decided on using google docstrings as the OCF style!)

Note: For VSCode people, use this extension https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff for ruff integration in the IDE. Also, feel free to set up a precommit as the following to avoid ever having to think about anything:

```
repos:
- repo: https://github.com/astral-sh/ruff-pre-commit
  # Ruff version.
  rev: v0.9.7
  hooks:
    # Run the linter.
    - id: ruff
      args: [ --fix ]
    # Run the formatter.
    - id: ruff-format
```

If not setting up pre-commit, then just listen to the ruff extension and fix them, or use `ruff check --fix .` periodically to clean up problems. Mob programming sessions for formatting can be significantly streamlined by using `ruff format .` from the outset to produce formatted code.


ocf_data_sampler/sample/base.py
```diff
@@ -1,61 +1,54 @@
...
+NumpyBatch: TypeAlias = dict[str, np.ndarray | dict[str, np.ndarray]]
+TensorBatch: TypeAlias = dict[str, torch.Tensor | dict[str, torch.Tensor]]
 
 
 class SampleBase(ABC):
-    """ 
-    Abstract base class for all sample types 
-    Provides core data storage functionality
-    """
+    """Abstract base class for all sample types.
```
felix-e-h-p, line 22: Likewise with full stops - but all good, best not to deviate from the google standard right.

Looks good: Ruff - integration - pre-commit hooks - CI enforcement

Cheers!

felix-e-h-p (approved):

ocf_data_sampler/select/select_time_slice.py
```diff
@@ -1,14 +1,17 @@
...
+
 
 def select_time_slice(
     ds: xr.DataArray,
     t0: pd.Timestamp,
     interval_start: pd.Timedelta,
     interval_end: pd.Timedelta,
     sample_period_duration: pd.Timedelta,
-):
+) -> xr.Dataset:
```
dfulu, line 14:
```suggestion
) -> xr.DataArray:
```

ocf_data_sampler/select/find_contiguous_time_periods.py
```diff
@@ -53,23 +58,24 @@ def find_contiguous_time_periods(
             periods.append(period)
         start_i = next_start_i
 
-    assert len(periods) > 0, (
-        f"Did not find an periods from {datetimes}. " f"{min_seq_length=} {max_gap_duration=}"
-    )
+    if len(periods) == 0:
+        raise ValueError(
+            f"Did not find an periods from {datetimes}. {min_seq_length=} {max_gap_duration=}",
```
dfulu, line 63:
```suggestion
            f"Did not find any periods from {datetimes}. {min_seq_length=} {max_gap_duration=}",
```

ocf_data_sampler/select/select_time_slice.py
```diff
@@ -27,23 +31,29 @@ def select_time_slice_nwp(
     sample_period_duration: pd.Timedelta,
     dropout_timedeltas: list[pd.Timedelta] | None = None,
     dropout_frac: float | None = 0,
-    accum_channels: list[str] = [],
+    accum_channels: list[str] | None = None,
```
dfulu, line 34: Is this a recommended style?

ocf_data_sampler/torch_datasets/utils/validate_channels.py
```diff
@@ -1,4 +1,4 @@
-import xarray as xr
+"""Functions for validating channels within a torch dataset."""
```
dfulu, line 1:
```suggestion
"""Functions for checking that normalisation statistics exist for the data channels requested"""
```

ocf_data_sampler/select/select_time_slice.py
devsjc: Yeah, you don't want to be using mutable arguments as defaults where you can help it: https://docs.astral.sh/ruff/rules/mutable-argument-default/

ocf_data_sampler/select/select_time_slice.py
devsjc: Are you sure just selecting a time slice turns it into a data array? Isn't it just a cut down dataset? You haven't chosen a single instance of the dataset's data_vars which would normally move between the two objects

tests/config/test_config.py
```diff
@@ -27,10 +28,11 @@ def test_incorrect_interval_start_minutes(test_config_filename):
 
     configuration = load_yaml_configuration(test_config_filename)
 
-    configuration.input_data.nwp['ukv'].interval_start_minutes = -1111
+    configuration.input_data.nwp["ukv"].interval_start_minutes = -1111
     with pytest.raises(
-        ValueError, 
-        match="interval_start_minutes.*must be divisible.*time_resolution_minutes.*"
+        ValueError,
```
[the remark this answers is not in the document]
dfulu: The pydantic config does check a lot of things. Do you not think its worth checking the specific message?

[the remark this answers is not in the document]
dfulu: That really IS a question. I'm happy not to have to match messages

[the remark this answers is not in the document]
devsjc: I guess it depends what we're actually checking for. If we just care about it failing validation, then there's no need, but if we care about the user seeing the right message and not just getting an error, then we should leave as is! Wasn't saying it should be changed, just wondered if it was actually a test for what we cared about.

tests/conftest.py
```diff
@@ -29,18 +29,25 @@ def session_tmp_path(tmp_path_factory):
 
 @pytest.fixture(scope="session")
 def sat_zarr_path(session_tmp_path):
-
     # Define coords for satellite-like dataset
     variables = [
-        'IR_016', 'IR_039', 'IR_087', 'IR_097', 'IR_108', 'IR_120', 
-        'IR_134', 'VIS006', 'VIS008', 'WV_062', 'WV_073',
+        "IR_016",
```
dfulu, line 34: I actually hate this

dfulu (approved):
Thanks for all this

ocf_data_sampler/select/select_time_slice.py
dfulu: Well the type hint on the input is a DataArray too

ocf_data_sampler/sample/base.py
dfulu: I hate the trailing full stop. But I do agree that its not worth the time

ocf_data_sampler/select/select_time_slice.py
dfulu: Okay cool. I do still hate it stylistically, but I'll take your point

tests/config/test_config.py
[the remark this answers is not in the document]
dfulu: Yeh I put in the regex since we use an f-string to put in what `interval_start_minutes` and `time_resolution_minutes` are in the message and I didn't want to have to hard code these. Happy to leave it as a general catchall

tests/conftest.py
devsjc: To be fair, me too - maybe we can put an ignore in for this spefic formatting style. I understand the benefit - it means adding and removing things from the list generate simpler diffs - but it does make your code harder to read! Maybe we can just handle it with code folding

tests/config/test_config.py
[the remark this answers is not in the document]
devsjc: I think it's fine as is and you can revisit it if it trips you up again in the future!

ocf_data_sampler/select/select_time_slice.py
devsjc: Feel free to revert with an ignore - I haven't seen it actually cause problems, but it is a footgun to be aware of I guess! Especially if you have a penchant for in-place modifications

ocf_data_sampler/select/select_time_slice.py
devsjc: Haha ahh good point! I just read the `ds` and made assumptions!

ocf_data_sampler/sample/base.py
devsjc: (I don't mind it you know, helps to differentiate it as a bit of plain English and not some code string - but then I'm already out of what my ideal codebase would look like by using python in the first place! So maybe it's an easier for me to make concessions when weighted against that...)

dfulu: Absolute Heresy

PR #186, opened 2025-02-24
Improve Readability of test_config.py

SophiaLi20:
Hey,
I’ve added inline comments to test_config.py to make it easier to understand what each test case is validating. The goal is to improve readability and help anyone working on this file quickly grasp the purpose of each test. No logic or some kind of functionality has been changed.
Advantages of the changes that has been made:

- Helps new contributors or anyone reviewing tests understand them faster.
- Reduces the time needed to figure out the intent of each test case.
- Makes the test suite more maintainable in the long run.

Testing:
These changes only add comments, so there are no new tests. The existing test cases remain the same and should run as expected without any issues.


Sukh-P:
@SophiaLi20 thanks for this but I think some of these changes to the script are covered in the other PR and for now we don't think the additional comments are quite necessary in the test config, so I will close this one if that's okay and we can focus on the other PR you have raised, thanks!

Issue #179, opened 2025-02-20
Decouple solar coordinates from specific inputs e.g. gsp/site


Sukh-P:
We want to decouple the solar coordinates from gsp or site for example, would need their own config sections so that these can be created independently of having another input.

PR #190, opened 2025-02-26
Solar coords

felix-e-h-p:
Decoupled solar coords - potential inclusion for further tests here, yet not sure if entirely necessary - updated tests relative to changes none the less

Relates to https://github.com/openclimatefix/ocf-data-sampler/issues/179


ocf_data_sampler/config/model.py
```diff
@@ -229,13 +229,28 @@ class Site(TimeWindowMixin, DropoutMixin):
     # TODO validate the csv for metadata
 
 
+class SolarPosition(TimeWindowMixin):
+    """Solar position configuration model."""
+
+    apply_to_gsp: bool = Field(
```
Sukh-P, line 235: I'm not sure about these apply_to bits, feels like it is still tying the solar positions/coordinates to gsp and site where as I think we wanted to separate this as much as possible

ocf_data_sampler/load/utils.py
```diff
@@ -46,3 +48,53 @@ def get_xr_data_array_from_xr_dataset(ds: xr.Dataset) -> xr.DataArray:
...
+
+    # For each location, calculate and add solar position
+    for location_id in data_array[id_dim].values:
+        # Get location coordinates
+        location_data = data_array.sel({id_dim: location_id})
+        lon = float(location_data[x_attr].values)
+        lat = float(location_data[y_attr].values)
+
+        # Calculate solar position and normalise
+        azimuth, elevation = calculate_azimuth_and_elevation(times, lon, lat)
```
Sukh-P, line 82: Have you tested this line? I feel like it won't work because for UK regional for example the coordinate system is osgb first and not lat long so there needs be a conversion to lat long before `calculate_azimuth_and_elevation`should be called on it, as is done previously [here](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py#L96)

ocf_data_sampler/numpy_sample/sun_position.py
```diff
@@ -35,7 +35,7 @@ def make_sun_position_numpy_sample(
     datetimes: pd.DatetimeIndex,
     lon: float,
     lat: float,
-    key_prefix: str = "gsp",
+    key_prefix: str | None = None,
```
Sukh-P, line 38: I feel like we want to remove key_prefixes altogether perhaps? Would need to make sure in other bits of code we account for that but I think that's what we are aiming for

Sukh-P (changes requested):
Thanks for making a start on this, think the logic of the code needs to be double checked and still more decoupling of the solar coordinates from GSP/site

ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
```diff
@@ -103,10 +103,31 @@ def process_and_combine_datasets(
...
+        )
+
+        # We already have lon, lat if target_key is "gsp", otherwise calculate them
+        if target_key != "gsp":
+            lon, lat = osgb_to_lon_lat(location.x, location.y)
+
+        # Calculate solar positions and add to modalities
+        solar_positions = make_sun_position_numpy_sample(datetimes, lon, lat)
+        prefixed_solar_positions = {
+            f"gsp_{key}": value for key, value in solar_positions.items()
```
Sukh-P, line 127: Hmm I was thinking that instead of having gsp or site as a prefix we wanted to move away from that so we would just have `solar_position` itself as the prefix and then either azimuth/elevation

Sukh-P:
@felix-e-h-p I think this looks good and pretty much ready to go, I think that perhaps it still might break things on the site side, namely I am not sure [this function](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/torch_datasets/datasets/site.py#L342C5-L342C35) will work now, so would it be possible to add a test which generates a site sample with the new solar configuration and then pass it through this function to see whether it gives you a numpy sample with the correct keys (namely a key that relates to solar), I think it won't and that function will need some tweaking, thanks! 

Also there definitely will need to be some changes in PVNet alongside this as you can see here https://github.com/openclimatefix/PVNet/blob/main/pvnet/models/multimodal/multimodal.py#L347-L351 it relies on a naming convention that we want to move away from

ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
```diff
@@ -103,10 +103,31 @@ def process_and_combine_datasets(
...
+    # Only add solar position if explicitly configured
+    has_solar_config = (
+        hasattr(config.input_data, "solar_position") and
+        config.input_data.solar_position is not None
     )
 
+    if has_solar_config:
+        # Create datetime range for solar position calculation
+        datetimes = pd.date_range(
+            t0 + minutes(gsp_config.interval_start_minutes),
```
dfulu, line 115: The initial idea of this was to stop the solar coords depending on the target data settings. We should use the start, end and time resolution minutes from the solar config

ocf_data_sampler/torch_datasets/datasets/site.py
```diff
@@ -261,17 +261,24 @@ def process_and_combine_site_sample_dict(
...
+        # Only add solar position if explicitly configured
+        has_solar_config = (
+            hasattr(self.config.input_data, "solar_position") and
+            self.config.input_data.solar_position is not None
         )
 
+        if has_solar_config:
+            # add sun features
+            sun_position_features = make_sun_position_numpy_sample(
+                datetimes=datetimes,
```
dfulu, line 273: Similarly to the other comment, we should decouple the solar coords from the target data config here

dfulu (changes requested):

felix-e-h-p:
Cheers for the feedback.

pvnet_uk.py now has a solar config object with  time parameters instead of borrowing from GSP settings (also updated relevant variable namings RE prefix).

site.py likewise has a solar config object with related timing parameters - hopefully properly handling site timestamp dimensions. t0 parameter is now passed from _get_sample method to process_and_combine_site_sample_dict so that the initial timestamp can be used for calculating the solar position timestamps accurately (no mismatch here).

Site tests updated accordingly - two additional tests.

dfulu:
@felix-e-h-p thanks for the changes. 

There is some duplication in [these lines](https://github.com/openclimatefix/ocf-data-sampler/blob/93e8106ceee58f704f054d25e19d9aae0d5d0d8c/ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py#L88-L123) which should be removed. Also whilst we're here, it might be best to strip out the `target_key` param from that `process_and_combine_datasets()` function. That param is no longer used

ocf_data_sampler/torch_datasets/datasets/site.py
```diff
@@ -261,18 +263,41 @@ def process_and_combine_site_sample_dict(
...
+
+            # Create datetime range using solar config params
+            solar_datetimes = pd.date_range(
+                t0 + minutes(solar_config.interval_start_minutes),
+                t0 + minutes(solar_config.interval_end_minutes),
+                freq=minutes(solar_config.time_resolution_minutes),
+            )
+
+            # Ensure matching of site time dimension length
+            solar_datetimes = solar_datetimes[:site_time_length]
```
dfulu, line 287: This couples the solar coords to the target datetimes. It also means the solar config doesn't reflect the datetimes of the solar coords. This should be removed

ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
```diff
@@ -103,10 +103,32 @@ def process_and_combine_datasets(
...
+            freq=minutes(solar_config.time_resolution_minutes),
+        )
+
+        if target_key != "gsp":
+            lon, lat = osgb_to_lon_lat(location.x, location.y)
+
+        # Calculate solar positions and add to modalities
+        solar_positions = make_sun_position_numpy_sample(datetimes, lon, lat)
+        solar_positions = {
+            f"solar_position_{key}": value for key, value in solar_positions.items()
```
dfulu, line 128: We shouldn't be changing the keys from what comes out of the `make_sun_position_numpy_sample()` function. If we want these key names this should be moved inside the function

tests/torch_datasets/test_pvnet_uk.py
```diff
@@ -74,15 +75,18 @@ def test_pvnet_uk_regional_dataset(pvnet_config_filename):
...
-        "gsp",
-        "gsp_solar_azimuth",
-        "gsp_solar_elevation",
-    ]:
+    # These keys should always be present
+    required_keys = ["nwp", "satellite_actual", "gsp"]
+    for key in required_keys:
         assert key in sample
 
+    # Check solar position keys only if present in the sample
```
dfulu, line 83: We shouldn't make this an if statement. For the test we should know whether the solar data is expected to be in this sample. We should definitely include it and test it here

tests/torch_datasets/test_pvnet_uk.py
```diff
@@ -129,15 +130,18 @@ def test_pvnet_uk_concurrent_dataset(pvnet_config_filename):
...
```
dfulu, line 138: Similarly, we should know if it will be and test for that

tests/torch_datasets/test_pvnet_uk.py
```diff
@@ -74,15 +75,18 @@ def test_pvnet_uk_regional_dataset(pvnet_config_filename):
...
+    solar_keys = ["solar_position_azimuth", "solar_position_elevation"]
+    if all(key in sample for key in solar_keys):
+        # Solar angles have same shape as GSP data
```
dfulu, line 86: I think this comment should not in general be true. We are trying to completely decouple the solar coords from the GSP datetimes. This will now depend on how the user configures the solar coords used

ocf_data_sampler/torch_datasets/datasets/site.py
felix-e-h-p: Wasn't sure with this, cheers. Now have created a dimension for solar position data - ensures solar coords are determined solely by config params.

ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
```diff
@@ -41,12 +41,102 @@
 xr.set_options(keep_attrs=True)
 
 
+# def process_and_combine_datasets(
```
dfulu, line 44: This commented out code should all go

ocf_data_sampler/sample/uk_regional.py
```diff
@@ -55,12 +55,6 @@ def plot(self) -> None:
             axes[0, 0].plot(self._data[GSPSampleKey.gsp])
             axes[0, 0].set_title("GSP Generation")
 
-        if GSPSampleKey.solar_azimuth in self._data and GSPSampleKey.solar_elevation in self._data:
```
dfulu, line 58: Is there any particular reason to get rid of this rather than just to update it to the new keys?

ocf_data_sampler/torch_datasets/datasets/site.py
```diff
@@ -261,18 +263,40 @@ def process_and_combine_site_sample_dict(
...
-        )
-        combined_sample_dataset = combined_sample_dataset.assign_coords(
-            {k: ("site__time_utc", v) for k, v in sun_position_features.items()},
+        # Only add solar position if explicitly configured
+        has_solar_config = (
+            hasattr(self.config.input_data, "solar_position") and
+            self.config.input_data.solar_position is not None
         )
 
-        # TODO include t0_index in xr dataset?
```
dfulu, line 275: Was there a reason to get rid of this TODO? Is it resolved?

tests/torch_datasets/test_site.py
```diff
@@ -197,3 +197,109 @@ def test_potentially_coarsen(ds_nwp_ecmwf):
...
+    # Verify solar position data exists in sample
+    solar_keys = ["solar_azimuth", "solar_elevation"]
+    for key in solar_keys:
+        assert key in loaded_sample.coords, f"Solar key {key} not found in loaded netCDF"
+
+    # Conversion and subsequent assertion
+    numpy_sample = convert_netcdf_to_numpy_sample(loaded_sample)
+    assert isinstance(numpy_sample, dict)
+
+    # Explicitly verify what is in sample
```
Sukh-P, line 302: Thanks for adding this test in, I think it still needs to check that the solar position values exist in the numpy sample which it doesn't do currently

ocf_data_sampler/sample/uk_regional.py
felix-e-h-p: Wasn't necessarily sure RE UKRegionalSample, felt perhaps something to address after full decoupling is complete - solar position calculations are now optional based on config right, have this updated just to reflect that

ocf_data_sampler/torch_datasets/datasets/site.py
felix-e-h-p: Sorry yeah - t0 parameter now passed to the method itself, the TODO for t0_index in the xarray dataset itself hasn't been fully addressed - separate PR after? Reinstated the comment

ocf_data_sampler/sample/uk_regional.py
dfulu: True, but the NWP, satellite and GSP are also all optional based on the config. So I'd leave it in

dfulu (approved):
Thanks for the changes!

Sukh-P (approved):

PR #198, opened 2025-03-03
Fix: Handle Single-Timestamp Case and Add Test Cases for other functions like intersection of 2 DF`

### LINKED ISSUE #192: Allow for single timestamp periods during contiguous time periods selection
Currently find_contiguous_time_periods function does not allow for sequence length = 1. We would like for it to be a valid time period (start and end time just have to be set to the same value in that case), to allow for e g single satellite picture selection. This will probably require changes in interval intersection selection in the same file.

MAYANK12SHARMA:
## Description

This PR updates the `find_contiguous_time_periods` function to allow for single-timestamp sequences when `min_seq_length` is set to `1`. Previously, the function required `min_seq_length` to be strictly greater than `1`, which prevented valid single-point time periods (e.g., single satellite image selection).

### Key Changes:
- **Validation Update:** `min_seq_length` can now be `1`.
- **Single-Timestamp Support:** If only one timestamp exists and `min_seq_length == 1`, a valid period is returned with `start_dt` and `end_dt` set to the same value.
- **Sequence Length Check:** Changed the condition to accept sequences `>= min_seq_length`, instead of strictly greater.

Fixes #192 

## How Has This Been Tested?

Tested with various time series scenarios:
- Single timestamp with `min_seq_length = 1`
- Single timestamp with `min_seq_length > 1` (raises expected error)
- Multiple timestamps with and without gaps

Added new test cases to cover these conditions.

- [x] Single timestamp with `min_seq_length = 1`
- [x] Multi-timestamp sequences with valid/invalid gaps
- [x] Edge cases for exact sequence lengths

## Checklist:


tests/select/test_location.py
```diff
@@ -8,29 +8,29 @@ def test_make_valid_location_object_with_default_coordinate_system():
     location = Location(x=x, y=y)
     assert location.x == x, "location.x value not set correctly"
     assert location.y == y, "location.x value not set correctly"
-    assert location.coordinate_system == "osgb", (
-        "location.coordinate_system value not set correctly"
-    )
+    assert (
+        location.coordinate_system == "osgb"
+    ), "location.coordinate_system value not set correctly"
```
AUdaltsova, line 13: Is there a reason for this change?

AUdaltsova: Oh my bad!

ocf_data_sampler/select/find_contiguous_time_periods.py
```diff
@@ -31,10 +31,19 @@ def find_contiguous_time_periods(
...
 
+    # Handle single timestamp case
+    if len(datetimes) == 1:
+        if min_seq_length == 1:
+            return pd.DataFrame([{"start_dt": datetimes[0], "end_dt": datetimes[0]}])
+        else:
+            raise ValueError(
+                "Only one timestamp found, but min_seq_length > 1. No valid periods.",
+            )
+
```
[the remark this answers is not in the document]
AUdaltsova: Yes this is the same Todo I meant! We have to allow overlap of 1 timestamp between intervals for single timestamp data to work correctly, otherwise some cases won't be covered - e g, when one source's interval ends exactly where the single timestamp of another source lies.

Speaking of edge cases, it would be great if you could add tests to this. I would check that a single timestamp is located correctly, that a t0 can be created when it requires just one timestamp from satellite, and that interval selection works when one of them is a single timestamp and is located inside the other, outside the other, and at the endpoints. The t0 one can be tricky, let me know if you would like me to help with it!

MAYANK12SHARMA:
Hi @AUdaltsova, can you take a moment to review it now?

Sukh-P:
@MAYANK12SHARMA thanks for having a go at this issue, I just want to flag that since this issue does not have the `contributions-welcome` tag on it which we are using as stated in our contributions guidelines [here](https://github.com/openclimatefix) this won't be a priority for us to review/implement as we have found in our experience that these sort of issues are more efficient to handle internally, we can try to get back to you when we can but due to this being a busy period at OCF I apologise if there is delay, thanks

MAYANK12SHARMA:
"Hi @Sukh-P, no worries at all. Whenever you have some time, I’d appreciate it if you could review it. I’ve been contributing to OCF since January. I have worked on ocf proejcts like open_data_pvnet, analysis dashboard, pvsitedatamodel, nowcasting datamodel and more.

MAYANK12SHARMA:
@peterdudfield, @Sukh-P  Could you please take a moment to review the PR. It's been 3 weeks.

PR #200, opened 2025-03-04
Spatial slice select update

felix-e-h-p:
Main refactor aspects are as follows:

CoordinateConverter Class - introduced to encapsulate coordinate transformations.

Combined Pixel Index Function - two pixel-index functions merged into one, using the CoordinateConverter class to handle different coordinate systems.

Improved Padding Logic - pre-existing padding logic simplified by calculating quantities explicitly.

New test added to cover cases with allow_partial_slice=False (not sure if truly necessary).

### DIFF
--- a/ocf_data_sampler/select/select_spatial_slice.py
+++ b/ocf_data_sampler/select/select_spatial_slice.py
@@ -187,7 +187,7 @@ def select_spatial_slice_pixels(
 
     Args:
         da: xarray DataArray to slice from
-        location: Location of interest
+        location: Location of interest that will be the center of the returned slice
         height_pixels: Height of the slice in pixels
         width_pixels: Width of the slice in pixels
         allow_partial_slice: Whether to allow a partial slice.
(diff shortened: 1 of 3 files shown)


felix-e-h-p:
Think all OK here - perhaps would benefit with the next mob programming session

ocf_data_sampler/select/select_spatial_slice.py
```diff
@@ -214,51 +162,52 @@ def _select_partial_spatial_slice_pixels(
...
 
     Args:
         da: xarray DataArray to slice from
-        center_idx: Location object describing the centre of the window with index coordinates
-        width_pixels: Window with in pixels
-        height_pixels: Window height in pixels
-        x_dim: Name of the x-dimension in `da`
-        y_dim: Name of the y-dimension in `da`
-        allow_partial_slice: Whether to allow a partially filled window
+        location: Location of interest
```
dfulu, line 190: This function is user-facing I think we should be more explicit about what the location is. Can we either this, or maybe better yet to docstring to mention that we slice a window **centred** on the location

dfulu:
@felix-e-h-p this looks really good! There are just a few little things I think we should tighten up on

dfulu (changes requested):

dfulu (approved):
Looks great @felix-e-h-p! Thanks for taking this on

PR #206, opened 2025-03-13
Move normalisation values to config

dfulu:
# Pull Request

This PR moves the normalisation values rom being stored in the library to being defined in the data config

## Checklist:


ocf_data_sampler/config/model.py
```diff
@@ -125,7 +127,35 @@ class SpatialWindowMixin(Base):
...
+    """Normalisation constants for multiple channels."""
+    normalisation_constants: dict[str, NormalisationValues]
+
+    @property
+    def channel_means(self) -> xr.DataArray:
+        """Return the channel means."""
+        return xr.DataArray(
+            [norm_values.mean for norm_values in self.normalisation_constants.values()],
+            coords={"channel": list(self.normalisation_constants.keys())},
+        ).astype(np.float32)
```
peterdudfield, line 146: should we just pass back a dictionary here? I know we use DataArrays later on, but it seems like extra code in this configuration, that perhaps isnt always needed

ocf_data_sampler/config/model.py
```diff
@@ -139,8 +169,20 @@ class Satellite(TimeWindowMixin, DropoutMixin, SpatialWindowMixin):
...
+    def check_all_channel_have_normalisation_constants(self) -> "Satellite":
+        """Check that all the channels have normalisation constants."""
+        normalisation_channels = set(self.normalisation_constants.keys())
+        missing_norm_values = set(self.channels) - set(normalisation_channels)
+        if len(missing_norm_values)>0:
+            raise ValueError(
+                "Normalsation constants must be provided for all channels. Missing values for "
+                f"channels: {missing_norm_values}",
+            )
+        return self
```
peterdudfield, line 182: nice

ocf_data_sampler/constants.py
```diff
@@ -1,350 +1,8 @@
...
-            return super().__getitem__(key)
-        else:
-            raise KeyError(
-                f"Values for {key} not yet available in ocf-data-sampler {list(self.keys())}",
-            )
-
-
-# ------ UKV
-# Means and std computed WITH version_7 and higher, MetOffice values
-UKV_STD = {
```
peterdudfield, line 40: Just so we dont lose them, is it worth putting this either in a readme, or a csv somewhere?

ocf_data_sampler/config/model.py
dfulu: Well the idea in returning a dataarray was to allow [these lines](https://github.com/openclimatefix/ocf-data-sampler/blob/4a2ef414c6f65217b4c608cd8bc6bde53540751b/ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py#L54-L57) where we apply the normalisation to be really clean. If we returned a dict we'd have to change those

ocf_data_sampler/constants.py
dfulu: I'm not too worried about losing these. They will always be in the history of this repo and the plan is to put them in the example configs in PVNet

ocf_data_sampler/config/model.py
peterdudfield: yea I undersatnd it makes those lines of code easy. Just feels slightly like that is an implementation and it makes the configuration a bit messy. If you prefer to keep it this way, then please leave it

peterdudfield (approved):
Looks good to me

peterdudfield:
Perhaps should be a minor changed as this causes a breaking change

zakwatts (approved):
Nice work

PR #183, opened 2025-02-21
Mob cleaning 2

dfulu:
# Pull Request

## Description

This PR comes from the  2nd mob programming session to clean up data sampler

## Checklist:

- [ ] My code follows [OCF's coding style guidelines](https://github.com/openclimatefix/.github/blob/main/coding_style.md)
- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked my code and corrected any misspellings

PR #212, opened 2025-03-14
Downstream update for `dropout_timedeltas_minutes` in config changing from `None`-> `[]` for null value

dfulu:
# Pull Request

## Description

Recently in one of the mob programming sessions we [updated the config model ](https://github.com/openclimatefix/ocf-data-sampler/pull/183/files#diff-7d8af53f8ccb0fe5ee493bd54771546ce767703b9d7ea18f09d70e1041563ae2) to update the null value of `dropout_timedeltas_minutes` from `None` to an empty list. This has broken some downstream code which still expects the null value of the list to be `None`.

This PR allows the functions where `dropout_timedeltas_minutes` is passed in to run with an empty list

## Checklist:

### DIFF
--- a/ocf_data_sampler/select/select_time_slice.py
+++ b/ocf_data_sampler/select/select_time_slice.py
@@ -55,7 +55,10 @@ def select_time_slice_nwp(
     if accum_channels is None:
         accum_channels = []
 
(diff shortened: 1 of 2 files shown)


Sukh-P (approved):
LGTM!

Issue #211, opened 2025-03-14
Cannot convert arrays with negative strides to torch tensors


dfulu:
When trying to create samples using the `save_samples.py` script in PVNet I came across an issue

```
  File "/home/jamesfulton/mambaforge/envs/pvnet0/lib/python3.12/site-packages/torch/utils/data/_utils/collate.py", line 64, in default_convert
    return torch.as_tensor(data)
           ^^^^^^^^^^^^^^^^^^^^^
ValueError: At least one stride in the given numpy array is negative, and tensors with negative strides are not currently supported. (You can probably work around this by making a copy of your array  with array.copy().) 
```

The basic problem is that the default collate function which stacks samples into batches cannot be used when any of the sample arrays are views with negative stride. i.e. where we are viewing an array in reverse. This is just a limitation of torch Tensors.

For a really simple example

```
import numpy as np
import torch

x = np.arange(10)
torch.as_tensor(x[::-1])

> Traceback (most recent call last):
>  File "<stdin>", line 1, in <module>
> ValueError: At least one stride in the given numpy array is negative, and tensors with negative strides are not currently supported. (You can probably work around this by making a copy of your array  with array.copy().) 
```

I did some digging and found that the only arrays with negative strides were the coordinate arrays (x/y) for datasets which had been flipped by [the `make_spatial_coords_increasing()` function](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/load/utils.py#L14). A really simple fix which does work is to just add lines like:

````
ds[x/y] = np.ascontiguousarray(ds[x/y].values)
```
to that function if it reverses the arrays. I've tested this locally and it does fix the issue.


I found it surprising that the values of the data array (i.e. not the coords but the actual data) do not have negative strides. But I suspect this is because the compute method returns a copy

It might be that we have more issues with negative strides and we might need to add `array.copy()` for each array in the NumpySample.

This stride issue disappears if the array is first saved to disk and reloaded before it is converted to a torch tensor. This is our main practice but I am still in support of putting in the minimal solution above since it is such a minimal cost

PR #213, opened 2025-03-14
Make coords contiguous if reversed in loading

dfulu:
# Pull Request

## Description

This PR is a proposal to fix issue #211. See the issue for full details

Fixes #

## How Has This Been Tested?

I have run this locally to create samples which are converted straight to torch tensors without being saved first

I have made sure the samples created by the UK PVNet sampler all the numpy arrays in the numpy batch have positive strides

Also this is a minimal example showing that the fix works:

```
import xarray as xr
import numpy as np
import dask

# Create a DataArray with a single dimension 'x'. The x coord is in decreading order: [9,8,7,...,1,0]
x_coords = np.arange(9, -1, -1)

# The underlying data is lazy - values [9,8,7,...,1,0]
values = dask.array.arange(9, -1, -1)

da = xr.DataArray(values, dims=['x'], coords={'x': x_coords}, name='example_data')

# Reverse the data in the x dimension
da = da.isel(x=slice(None, None, -1))

# Print the strides - x has a negative stride but the data has a positive stride
print("x-strides", da.x.data.strides)  # >> x-strides (-8,)
print("data-stride", da.values.strides)  # >> data-stride (8,)

# Make the x coord array contiguous
da["x"] = np.ascontiguousarray(da.x.values)

# Print the strides - both the data and the x coord have positive strides
print("x-strides", da.x.data.strides)  # >> x-strides (8,)
print("data-stride", da.values.strides)  # >> data-stride (8,)

# Sanity check to make sure we haven't muddled up the data and its coords somehow
for i in range(10):
    assert da.sel(x=i)==i
    assert da.isel(x=i)==i
```

## Checklist:

### DIFF
--- a/ocf_data_sampler/load/utils.py
+++ b/ocf_data_sampler/load/utils.py
@@ -1,5 +1,6 @@
 """Utility functions for working with xarray objects."""
 
+import numpy as np
 import pandas as pd
 import xarray as xr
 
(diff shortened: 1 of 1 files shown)


ocf_data_sampler/load/utils.py
```diff
@@ -22,8 +23,10 @@ def make_spatial_coords_increasing(ds: xr.Dataset, x_coord: str, y_coord: str) -
     # Make sure the coords are in increasing order
     if ds[x_coord][0] > ds[x_coord][-1]:
         ds = ds.isel({x_coord: slice(None, None, -1)})
+        ds[x_coord] = np.ascontiguousarray(ds[x_coord].values)
```
Sukh-P, line 26: A comment here about this is to avoid negative strides might be useful context

Sukh-P (approved):
Nice catch! Looks good

PR #223, opened 2025-04-03
patch dropout function for no dropout case

zakwatts:
# Pull Request

## Description

The current dropout functionality was patched in a recent PR #183, this changed how the dropout_time was set when no dropout was specified in the configuration to return t0, instead of None. This introduced NaNs in times after t0 which isn't the intended outcome.

This has been fixed by reinstating the previous functionality, setting dropout_time to None where dropout is not desired, with ds being returned in the apply_dropout function if the dropout_time is None.

## How Has This Been Tested?

The existing test, `test_draw_dropout_time_none` has also been updated to reflect the None case.

- [X] Yes

I have inspected and printed the generation locally to track and verify the changes.

- [X] Yes

## Checklist:

- [X] I have made corresponding changes to the documentation

### DIFF
--- a/ocf_data_sampler/select/dropout.py
+++ b/ocf_data_sampler/select/dropout.py
@@ -33,7 +33,7 @@ def draw_dropout_time(
         raise ValueError("dropout_frac must be between 0 and 1 inclusive")
 
     if (len(dropout_timedeltas) == 0) or (np.random.uniform() >= dropout_frac):
-        dropout_time = t0
+        dropout_time = None
     else:
         dropout_time = t0 + np.random.choice(dropout_timedeltas)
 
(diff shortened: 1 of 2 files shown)


Sukh-P (approved):
Nice work catching and fixing this!

ocf_data_sampler/select/dropout.py
```diff
@@ -42,13 +42,16 @@ def draw_dropout_time(
...
         ds: Xarray DataArray with 'time_utc' coordinate
         dropout_time: Time after which data is set to NaN
     """
-    # This replaces the times after the dropout with NaNs
-    return ds.where(ds.time_utc <= dropout_time)
+    if dropout_time is None:
+        return ds
+    else:
+        # This replaces the times after the dropout with NaNs
+        return ds.where(ds.time_utc <= dropout_time)
```
Sukh-P, line 57: Small syntax comment, the else isn't strictly needed, could be: 
```suggestion
    if dropout_time is None:
        return ds
    # This replaces the times after the dropout with NaNs
    return ds.where(ds.time_utc <= dropout_time)
```

zakwatts: Agreed the else isn't necessary but i think it could be clear with the else for new contributors so leaving it in for now.

Issue #180, opened 2025-02-20
Combine the draw and apply dropout time functions


Sukh-P:
We have two separate functions here https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/select/dropout.py which we can combine into one for simplicity

AUdaltsova:
This should be done after we merge the latest cleanup branch, I will update this comment when this is done.

Monson2002:
Hi @Sukh-P , wanted to contribute to this issue, is it open now?
PS: I also wanted some guidance on how to get this app running, want to set up a venv, but could not find a requirements.txt file to install dependencies. Thanks in advance!

Sukh-P:
Hi @Monson2002 yes it is, I can assign you to the issue, so there are some install instructions on the README itself and we use a [pyproject.toml ](https://github.com/openclimatefix/ocf-data-sampler/blob/main/pyproject.toml) file instead so can install with pip install . if you're in the ocf-data-sampler directory

Monson2002:
Does the setup take very long as its still going on (its been like 1 hour since I did `pip install .`)

MAYANK12SHARMA:
Hi @Sukh-P sir, I have combined both function and update the code as well. I want to raise the PR of it. before it can you please review my latest PR of issue #192 .

Dakshbir:
@Sukh-P sir is this issue still pen for contributions?
If yes can i solve this issue?

RupeshMangalam21:
@Monson2002 Are you still working on this? I can help you out if you're stuck somewhere.

Monson2002:
Actually I was not able to set up the app locally, that is why not able to test my fixes

RupeshMangalam21:
@Monson2002 Oh i get where you might be getting stuck, try tracing these steps --> Fork the ocf repository --> Open a new VScode Window(or any prefered editor on your machine) --> git clone https://github.com/RupeshMangalam21/ocf-data-sampler.git (clone YOUR FORK of the repo in the terminal) --> cd ocf-data-sampler --> pip install . --> git checkout -b new-branch --> apply your fixes --> pytest --> push your changes to the forked repo --> go to github forked repo and create a pr

PR #224, opened 2025-04-04
Refactor dropout

RupeshMangalam21:
# Pull Request  

## Description  
This PR combines the `draw_dropout_time` and `apply_dropout_time` functions into a single `simulate_dropout` function to simplify the dropout simulation workflow. The unified function retains the original functionality while reducing code complexity and improving maintainability. This change addresses the need to streamline the dropout process for time series data (GSP, Satellite, Site) as outlined in [issue #180](https://github.com/openclimatefix/ocf-data-sampler/issues/180), while incorporating critical fixes from [PR #223 ](https://github.com/openclimatefix/ocf-data-sampler/pull/223) to prevent unintended NaN values.

**Key changes**:  
- Replaced `draw_dropout_time` and `apply_dropout_time` with a consolidated `simulate_dropout` function.  
- Fixed critical issue where `dropout_time` defaults to `None` (instead of `t0`) when no dropout is applied, preserving data integrity.  
- Updated all affected code in `time_slice_for_dataset.py` to use the new unified function.  
- Refactored test cases to align with the new implementation while preserving test coverage.  

## How Has This Been Tested?  
- **Unit Tests**:  
  Updated `test_dropout.py` with comprehensive tests covering:  
  - Forced dropout times when `dropout_frac=1`  
  - No dropout behavior (`dropout_frac=0` and empty `dropout_timedeltas`)  
  - Partial dropout probability scenarios  
  - Input validation (negative timedeltas check, invalid configs)  
- **Sanity Checks**:  
  - Verified via xarray assertions that data remains unmodified when `dropout_frac=0`.  
  - Confirmed NaNs are only introduced after `dropout_time` when dropout is applied.  
- **Test Configuration**:  
  Tests run using pytest.

## Checklist:  
- [x] I have made corresponding changes to the documentation (function docstrings updated)  

**Fixes**: #180 (with follow-up fixes from #223)


RupeshMangalam21:
@Sukh-P Hi, could you please review. I've discarded the previous pr and created this one after pr #223 was merged with affecting changes.

ocf_data_sampler/select/time_slice_for_dataset.py
```diff
@@ -51,19 +51,14 @@ def slice_datasets_by_time(
             interval_end=minutes(sat_config.interval_end_minutes),
         )
 
-        # Randomly sample dropout
-        sat_dropout_time = draw_dropout_time(
+        # Apply dropout using the combined function
```
AUdaltsova, line 54:
```suggestion
        # Apply dropout
```

It's the only function now, so I don't think it needs to be specified

tests/select/test_dropout.py
```diff
@@ -3,69 +3,85 @@
...
-    dropout_time = pd.Timestamp(f"2024-01-01 {t0_str}")
-
-    da_dropout = apply_dropout_time(da_sample, dropout_time)
-
-    assert da_dropout.sel(time_utc=slice(None, dropout_time)).notnull().all()
-    assert (
-        da_dropout.sel(time_utc=slice(dropout_time + pd.Timedelta(5, "min"), None)).isnull().all()
-    )
+    times = pd.date_range(t0 - pd.Timedelta("1h"), t0 + pd.Timedelta("30min"), freq="10min")
+    ds = xr.DataArray(np.arange(len(times)), coords={"time_utc": times})
```
AUdaltsova, line 28: I think this is what `da_sample` does on L10? Is there a reason not to use it here?

ocf_data_sampler/select/dropout.py
```diff
@@ -9,19 +9,28 @@
...
     t0: pd.Timestamp,
     dropout_timedeltas: list[pd.Timedelta],
     dropout_frac: float,
-) -> pd.Timestamp:
-    """Randomly pick a dropout time from a list of timedeltas.
+) -> xr.DataArray:
+    """Combined function that draws a dropout time and applies dropout.
+
+    This function preserves the functionalities of the original two functions,
+    `draw_dropout_time` and `apply_dropout_time`.
```
AUdaltsova, line 21: The docstring should just explain what a new function does, I don't think context on what the code looked like before is usefule

tests/select/test_dropout.py
```diff
@@ -3,69 +3,85 @@
...
+
+    ds_dropout = simulate_dropout(ds, t0, dropout_timedeltas, dropout_frac=1)
+    non_nan = ds_dropout.where(~xr.ufuncs.isnan(ds_dropout), drop=True)
+    dropout_time = non_nan.time_utc.values[-1]
+    offset = pd.Timestamp(dropout_time) - t0
+    assert offset in dropout_timedeltas
```
AUdaltsova, line 34: This is a good test! I would repeat it a bunch of times and check that all dropout_timedeltas occur in the offsets you get instead maybe.

tests/select/test_dropout.py
```diff
@@ -3,69 +3,85 @@
...
+
+
+def test_simulate_dropout_draw_time_none(t0):
+    times = pd.date_range(t0 - pd.Timedelta("30min"), t0 + pd.Timedelta("30min"), freq="10min")
+    ds = xr.DataArray(np.arange(len(times)), coords={"time_utc": times})
```
AUdaltsova, line 39: Maybe I'm missing something, but I think it would be cleaner if they all used `da_sample`? I don't think it's used at all currently. If I'm missing something and it can't be used, then it should be removed

tests/select/test_dropout.py
```diff
@@ -3,69 +3,85 @@
...
+    # Test empty dropout_timedeltas
+    result = simulate_dropout(ds, t0, [], dropout_frac=0)
+    xr.testing.assert_equal(result, ds)
+
+
+@pytest.mark.parametrize(
+    "delta,expect_dropout",
+    [
+        (None, False),            # No dropout case
+        (pd.Timedelta(0), True),  # Dropout exactly at t0
```
AUdaltsova, line 54: This is an interesting edgecase actually, @dfulu I think we specifically allow for a 0 dropout, is t0 masking the desired behaviour for it?

AUdaltsova (changes requested):
Hi @RupeshMangalam21, thanks for your work on this! I've left some comments, but generally looks pretty good :)

RupeshMangalam21:
Thanks for the review! @AUdaltsova , i'll make the updates.

ocf_data_sampler/select/dropout.py
```diff
@@ -32,26 +41,15 @@ def draw_dropout_time(
...
+    # Return original data if no dropout
     if dropout_time is None:
-        return ds
-    else:
-        # This replaces the times after the dropout with NaNs
-        return ds.where(ds.time_utc <= dropout_time)
+        return ds.copy()
+
+    # Apply dropout
+    return ds.where(ds.time_utc <= dropout_time)
```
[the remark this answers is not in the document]
RupeshMangalam21: Thanks for catching this! I'll simplified the logic to remove redundancy. if `dropout_time` is None, we return `ds.copy()` directly, and the else clause is no longer needed. This would align with the intended behavior of preserving data integrity when no dropout is applied.

ocf_data_sampler/select/time_slice_for_dataset.py
RupeshMangalam21: I'll update the comment.

tests/select/test_dropout.py
RupeshMangalam21: Good catch! There’s no specific reason – the `da_sample` fixture can and should be used here. I initially recreated the dataset to align with `t0` dynamically, but since `da_sample` already provides valid time-coordinated data, I’ll refactor the test to reuse it.

tests/select/test_dropout.py
RupeshMangalam21: I think allowing dropout_time = t0 is intentional and tests a specific edge case where data at t0 is preserved, but all future data (times > t0) is masked, this would simulate scenarios where only historical data up to the exact forecast time is available.

the logic existed before, i just added tests for validation.

tests/select/test_dropout.py
RupeshMangalam21: Thanks for the feedback! I agree that repeating the test multiple times will help validate probabilistic behavior. For the test, do you have a recommended number of iterations? I was considering using 100 trials.

AUdaltsova: You can probably get away with less, at 50 trials you already have an 8e-16 probability of the test failing by chance, which I think is good enough

tests/select/test_dropout.py
RupeshMangalam21: i avoided da_sample here because it uses fixed timestamps unrelated to t0. This test requires dynamic time ranges centered on t0 to validate dropout offsets correctly. 

we can either keep this or modify da_sample to accept t0 as a parameter.

tests/select/test_dropout.py
RupeshMangalam21: What should we move forward with?

AUdaltsova:
Hi again @RupeshMangalam21, unfortunately, I think something went wrong with merge conflict resolution and now the PR is showing a lot of unrelated changes that have already happened. Would you mind fixing that so I can see the relevant changes? Maybe reopening the PR would be the easiest.

tests/select/test_dropout.py
AUdaltsova: Sorry if I'm missing something here, but t0 is still fixed to be a specific datetime - it's just hidden inside a fixture, right? I think the cleanest way would be to align da_sample and t0, this way the actual tests can be read as kind of "abstract" in terms of specific times they operate with and just have the logic in them. And you are right, I think it's best to make da_sample dependent on t0 for this.

tests/select/test_dropout.py
```diff
@@ -3,69 +3,85 @@
...
+
+    # Test empty dropout_timedeltas
+    result = simulate_dropout(ds, t0, [], dropout_frac=0)
+    xr.testing.assert_equal(result, ds)
+
+
+@pytest.mark.parametrize(
+    "delta,expect_dropout",
+    [
+        (None, False),            # No dropout case
```
AUdaltsova, line 59: For no dropout timedeltas [now default to `[]`, None is not accepted by the config anymore](https://github.com/openclimatefix/ocf-data-sampler/blob/105f21db8034be9309e8bf9d4d9c7f4af9da4ab5/ocf_data_sampler/config/model.py#L85-L89)

AUdaltsova: Oh I can see that you're passing an empty list in the actual test. My bad!

tests/select/test_dropout.py
AUdaltsova: Same here, I think it's worth aligning `da_sample` and `t0` and using those instead of creating a new dataArray

AUdaltsova (changes requested):
I tried to take a look at more changes and what I've managed to locate looks good! I think after you update the tests the changes will be good to go. But I can't merge a PR with things I haven't reviewed, and I'm sure you'll agree that me reviewing 20+ files that are not relevant to this PR and have been merged already just to make sure there's nothing unexpected there is a bit counterproductive, so if you could either open a new PR for this issue with the changes you've made or find another way to remove the irrelevant commits that would be great. Thanks so much for your work!

RupeshMangalam21:
> Hi again @RupeshMangalam21, unfortunately, I think something went wrong with merge conflict resolution and now the PR is showing a lot of unrelated changes that have already happened. Would you mind fixing that so I can see the relevant changes? Maybe reopening the PR would be the easiest.

Yes actually i rebased with the ocf main, which added the commits that were merged by other PRs during mine was open. I'll quickly finalize the changes in a new pr!

Issue #227, opened 2025-04-08
Validate samples after open


peterdudfield:
## Detailed Description
After loading samples, lets validate the samples

## Context
- open pvnet data want to know if there batches are ok

## Possible Implementation
- add something is https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/torch_datasets/sample/uk_regional.py#L16

peterdudfield:
Check it has the right fields, and the data is in the right shape

dfulu:
@peterdudfield I think this could be closed now thanks to #233?

PR #233, opened 2025-04-10
Validate samples after load

felix-e-h-p:
Alongside updating related tests, two new functions:

validate_sample (inside UKRegionalSample class): Checks if individual sample meets criteria by verifying required keys exist, data shapes match expected dimensions and validating structure for NWP.

validate_samples: Processes multiple UKRegionalSample objects - applies validation to each using provided configuration.

Related to: https://github.com/openclimatefix/ocf-data-sampler/issues/227

### DIFF
--- a/ocf_data_sampler/config/model.py
+++ b/ocf_data_sampler/config/model.py
@@ -320,17 +320,16 @@ class Configuration(Base):
     general: General = General()
     input_data: InputData = InputData()
     expected_shapes: dict = Field(default_factory=dict)
-    
+
     def get_expected_shapes(self) -> dict:
-        """
-        Returns dictionary of expected shapes based on configuration.
-        
+        """Returns dictionary of expected shapes based on configuration.
+
         This combines explicitly defined expected_shapes with those
         calculated from input_data configuration.
         """
         result = {}
-        
+
         if self.expected_shapes:
             result.update(self.expected_shapes)
-            
+
         return result
--- a/ocf_data_sampler/config/model.py
+++ b/ocf_data_sampler/config/model.py
@@ -319,4 +319,3 @@ class Configuration(Base):
 
     general: General = General()
     input_data: InputData = InputData()
-    expected_shapes: dict = Field(default_factory=dict)
--- a/ocf_data_sampler/config/model.py
+++ b/ocf_data_sampler/config/model.py
@@ -319,3 +319,18 @@ class Configuration(Base):
 
     general: General = General()
     input_data: InputData = InputData()
+    expected_shapes: dict = Field(default_factory=dict)
+    
+    def get_expected_shapes(self) -> dict:
+        """
+        Returns dictionary of expected shapes based on configuration.
+        
+        This combines explicitly defined expected_shapes with those
+        calculated from input_data configuration.
+        """
+        result = {}
+        
+        if self.expected_shapes:
+            result.update(self.expected_shapes)
+            
+        return result
--- a/ocf_data_sampler/config/model.py
+++ b/ocf_data_sampler/config/model.py
@@ -320,16 +320,3 @@ class Configuration(Base):
     general: General = General()
     input_data: InputData = InputData()
     expected_shapes: dict = Field(default_factory=dict)
-
-    def get_expected_shapes(self) -> dict:
-        """Returns dictionary of expected shapes based on configuration.
-
-        This combines explicitly defined expected_shapes with those
-        calculated from input_data configuration.
-        """
-        result = {}
-
-        if self.expected_shapes:
-            result.update(self.expected_shapes)
-
-        return result
--- a/ocf_data_sampler/torch_datasets/sample/uk_regional.py
+++ b/ocf_data_sampler/torch_datasets/sample/uk_regional.py
@@ -11,7 +11,10 @@ from ocf_data_sampler.numpy_sample import (
 )
 from ocf_data_sampler.numpy_sample.common_types import NumpySample
 from ocf_data_sampler.torch_datasets.sample.base import SampleBase
-from ocf_data_sampler.torch_datasets.utils.validation_utils import check_dimensions
+from ocf_data_sampler.torch_datasets.utils.validation_utils import (
+    calculate_expected_shapes,
+    check_dimensions,
+)
 
 
 class UKRegionalSample(SampleBase):
@@ -57,64 +60,18 @@ class UKRegionalSample(SampleBase):
         Returns:
             dict: Validation results with status and any validation errors.
         """
-
         # Check for required keys
         for key in config.get("required_keys", []):
             if key not in self._data:
                 raise ValueError(f"Missing required key: {key}")
 
-        # Start with configured expected shapes
-        expected_shapes = {}
-        if hasattr(config, "expected_shapes"):
-            expected_shapes.update(config.expected_shapes)
-        elif isinstance(config, dict) and "expected_shapes" in config:
-            expected_shapes.update(config["expected_shapes"])
-
-        # Calculate expected shapes from config if available
-        if hasattr(config, "input_data"):
-            input_data = config.input_data
-
-            # Calculate GSP shape
-            if hasattr(input_data, "gsp") and input_data.gsp is not None:
-                gsp_config = input_data.gsp
-                time_span = (
-                    gsp_config.interval_end_minutes -
-                    gsp_config.interval_start_minutes
-                )
-                resolution = gsp_config.time_resolution_minutes
-                expected_length = (time_span // resolution) + 1
-                expected_shapes[GSPSampleKey.gsp] = (expected_length,)
-
-            # Calculate NWP shape
-            if hasattr(input_data, "nwp") and input_data.nwp is not None:
-                for provider in input_data.nwp.values():
-                    expected_shapes[NWPSampleKey.nwp] = (
-                        provider.image_size_pixels_height,
-                        provider.image_size_pixels_width,
-                    )
-
-            # Calculate satellite shape
-            if hasattr(input_data, "satellite") and input_data.satellite is not None:
-                sat_config = input_data.satellite
-                channels = len(sat_config.channels)
-                time_span = (
-                    sat_config.interval_end_minutes -
-                    sat_config.interval_start_minutes
-                )
-                resolution = sat_config.time_resolution_minutes
-                time_steps = (time_span // resolution) + 1
-
-                expected_shapes[SatelliteSampleKey.satellite_actual] = (
-                    time_steps,
-                    channels,
-                    sat_config.image_size_pixels_height,
-                    sat_config.image_size_pixels_width,
-                )
+        # Calculate expected shapes from configuration
+        expected_shapes = calculate_expected_shapes(config)
 
         # Check GSP shape if specified
         if GSPSampleKey.gsp in self._data:
             gsp_data = self._data[GSPSampleKey.gsp]
-            
+
             # Check expected shape of GSP data if configured
             if GSPSampleKey.gsp in expected_shapes:
                 check_dimensions(
--- a/ocf_data_sampler/torch_datasets/sample/uk_regional.py
+++ b/ocf_data_sampler/torch_datasets/sample/uk_regional.py
@@ -121,8 +121,9 @@ class UKRegionalSample(SampleBase):
                     f"Satellite data ('{sat_key}') should have at least 2 dimensions, "
                     f"got shape {sat_data.shape}",
                 )
-            actual_spatial_dims = sat_data.shape[-2:]
-            expected_spatial_dims = expected_shapes[sat_key][-2:]
+
+            actual_spatial_dims = sat_data.shape
+            expected_spatial_dims = expected_shapes[sat_key]
 
             check_dimensions(
                 actual_shape=tuple(actual_spatial_dims),
--- a/ocf_data_sampler/torch_datasets/sample/uk_regional.py
+++ b/ocf_data_sampler/torch_datasets/sample/uk_regional.py
@@ -56,11 +56,6 @@ class UKRegionalSample(SampleBase):
         Returns:
             dict: Validation results with status and any validation errors.
         """
-        validation_result = {
-            "valid": True,
-            "errors": [],
-        }
-
         # Default config if none
         if config is None:
             config = {
@@ -73,14 +68,12 @@ class UKRegionalSample(SampleBase):
 
         # Check for required keys
         for key in config.get("required_keys", []):
-            if key not in self._data:
-                validation_result["valid"] = False
-                validation_result["errors"].append(f"Missing required key: {key}")
+            assert key in self._data, f"Missing required key: {key}"
 
         expected_shapes = config.get("expected_shapes", {}).copy()
 
+        # Calculate expected shapes from config if available
         if "input_data" in config:
-            # Calculate GSP shape if not in expected_shapes
             if GSPSampleKey.gsp not in expected_shapes and "gsp" in config["input_data"]:
                 gsp_config = config["input_data"]["gsp"]
                 time_span = (
@@ -126,72 +119,58 @@ class UKRegionalSample(SampleBase):
         for key, expected_shape in expected_shapes.items():
             if key in self._data:
                 actual_shape = self._data[key].shape
-                shape_valid = True
-                if len(actual_shape) != len(expected_shape):
-                    shape_valid = False
-                else:
-                    for actual_dim, expected_dim in zip(actual_shape, expected_shape, strict=False):
-                        if expected_dim is not None and actual_dim != expected_dim:
-                            shape_valid = False
-                            break
-
-                if not shape_valid:
-                    validation_result["valid"] = False
-                    validation_result["errors"].append(
-                        f"Shape mismatch for {key}: expected {expected_shape}, got {actual_shape}",
-                    )
+                
+                # Check dimensions match
+                assert len(actual_shape) == len(expected_shape), (
+                    f"Shape dimension mismatch for {key}: "
+                    f"expected {len(expected_shape)} dimensions, got {len(actual_shape)}"
+                )
+                
+                # Check each dimension matches
+                for i, (actual_dim, expected_dim) in enumerate(zip(actual_shape, expected_shape, strict=False)):
+                    if expected_dim is not None:
+                        assert actual_dim == expected_dim, (
+                            f"Shape mismatch for {key} at dimension {i}: "
+                            f"expected {expected_dim}, got {actual_dim}"
+                        )
 
         # Checks for NWP data - nested structure
         if NWPSampleKey.nwp in self._data:
             nwp_data = self._data[NWPSampleKey.nwp]
-            if not isinstance(nwp_data, dict):
-                validation_result["valid"] = False
-                validation_result["errors"].append("NWP data should be a dictionary")
-            else:
-                # Validate NWP data structure for each provider (ukv, ecmwf)
-                for provider, provider_data in nwp_data.items():
-                    if "nwp" not in provider_data:
-                        validation_result["valid"] = False
-                        validation_result["errors"].append(
-                            f"Missing 'nwp' key in NWP data for {provider}",
-                        )
+            assert isinstance(nwp_data, dict), "NWP data should be a dictionary"
+            
+            # Validate NWP data structure for each provider
+            for provider, provider_data in nwp_data.items():
+                assert "nwp" in provider_data, f"Missing 'nwp' key in NWP data for {provider}"
 
-                    # Check expected shape of NWP data if configured
-                    if config.get("nwp_shape") and "nwp" in provider_data:
-                        nwp_array = provider_data["nwp"]
-                        spatial_dims = nwp_array.shape[-2:]
+                # Check expected shape of NWP data if configured
+                if config.get("nwp_shape") and "nwp" in provider_data:
+                    nwp_array = provider_data["nwp"]
+                    spatial_dims = nwp_array.shape[-2:]
 
-                        if tuple(spatial_dims) != config.get("nwp_shape"):
-                            validation_result["valid"] = False
-                            validation_result["errors"].append(
-                                f"NWP shape mismatch for {provider}: "
-                                f"expected {config.get('nwp_shape')} for spatial dimensions, "
-                                f"got {spatial_dims}",
-                            )
+                    assert tuple(spatial_dims) == config.get("nwp_shape"), (
+                        f"NWP shape mismatch for {provider}: "
+                        f"expected {config.get('nwp_shape')} for spatial dimensions, "
+                        f"got {spatial_dims}"
+                    )
 
-        # Add satellite validation
+        # Validate satellite data
         if SatelliteSampleKey.satellite_actual in self._data:
             sat_data = self._data[SatelliteSampleKey.satellite_actual]
 
-            if len(sat_data.shape) >= 2:
-                spatial_dims = sat_data.shape[-2:]
-
-                # Check specific satellite shape if in config
-                if config.get("satellite_shape"):
-                    expected_spatial_dims = config.get("satellite_shape")[-2:]
-                    if spatial_dims != expected_spatial_dims:
-                        validation_result["valid"] = False
-                        validation_result["errors"].append(
-                            f"Satellite spatial dimensions mismatch: "
-                            f"expected {expected_spatial_dims}, got {spatial_dims}",
-                        )
-            else:
-                validation_result["valid"] = False
-                validation_result["errors"].append(
-                    f"Satellite data should have at least 2 dimensions, got shape {sat_data.shape}",
+            assert len(sat_data.shape) >= 2, (
+                f"Satellite data should have at least 2 dimensions, got shape {sat_data.shape}"
+            )
+            
+            spatial_dims = sat_data.shape[-2:]
+            if config.get("satellite_shape"):
+                expected_spatial_dims = config.get("satellite_shape")[-2:]
+                assert spatial_dims == expected_spatial_dims, (
+                    f"Satellite spatial dimensions mismatch: "
+                    f"expected {expected_spatial_dims}, got {spatial_dims}"
                 )
 
-        return validation_result
+        return True
 
     @override
     def plot(self) -> None:
--- a/ocf_data_sampler/torch_datasets/sample/uk_regional.py
+++ b/ocf_data_sampler/torch_datasets/sample/uk_regional.py
@@ -71,9 +71,6 @@ class UKRegionalSample(SampleBase):
                     NWPSampleKey.nwp,
                     SatelliteSampleKey.satellite_actual,
                 ],
-                "expected_shapes": {
-                    GSPSampleKey.gsp: (7,),
-                },
             }
 
         # Check for required keys
@@ -82,11 +79,55 @@ class UKRegionalSample(SampleBase):
                 validation_result["valid"] = False
                 validation_result["errors"].append(f"Missing required key: {key}")
 
-        # Check data shapes
-        for key, expected_shape in config.get("expected_shapes", {}).items():
+        expected_shapes = config.get("expected_shapes", {}).copy()
+
+        if "input_data" in config:
+            # Calculate GSP shape if not in expected_shapes
+            if GSPSampleKey.gsp not in expected_shapes and "gsp" in config["input_data"]:
+                gsp_config = config["input_data"]["gsp"]
+                time_span = (
+                    gsp_config["interval_end_minutes"] -
+                    gsp_config["interval_start_minutes"]
+                )
+                resolution = gsp_config["time_resolution_minutes"]
+                expected_length = (time_span // resolution) + 1
+                expected_shapes[GSPSampleKey.gsp] = (expected_length,)
+
+            # Calculate NWP shape
+            if "nwp" in config["input_data"]:
+                for provider in config["input_data"]["nwp"].values():
+                    if isinstance(provider, dict) and "image_size_pixels_height" in provider:
+                        config["nwp_shape"] = (
+                            provider["image_size_pixels_height"],
+                            provider["image_size_pixels_width"],
+                        )
+                        break
+
+            # Calculate satellite shape
+            if "satellite" in config["input_data"]:
+                sat_config = config["input_data"]["satellite"]
+                has_height = "image_size_pixels_height" in sat_config
+                has_width = "image_size_pixels_width" in sat_config
+                if has_height and has_width:
+                    channels = len(sat_config.get("channels", []))
+                    interval_end = sat_config["interval_end_minutes"]
+                    interval_start = sat_config["interval_start_minutes"]
+                    time_span = interval_end - interval_start
+                    resolution = sat_config["time_resolution_minutes"]
+                    time_steps = (time_span // resolution) + 1
+
+                    satellite_shape = (
+                        time_steps,
+                        channels,
+                        sat_config["image_size_pixels_height"],
+                        sat_config["image_size_pixels_width"],
+                    )
+                    config["satellite_shape"] = satellite_shape
+
+        # Check data shapes against calculated or config given shapes
+        for key, expected_shape in expected_shapes.items():
             if key in self._data:
                 actual_shape = self._data[key].shape
-                # Check if shapes match
                 shape_valid = True
                 if len(actual_shape) != len(expected_shape):
                     shape_valid = False
--- a/ocf_data_sampler/torch_datasets/sample/uk_regional.py
+++ b/ocf_data_sampler/torch_datasets/sample/uk_regional.py
@@ -57,71 +57,71 @@ class UKRegionalSample(SampleBase):
         Returns:
             dict: Validation results with status and any validation errors.
         """
-        # Default config if none
-        if config is None:
-            config = {
-                "required_keys": [
-                    GSPSampleKey.gsp,
-                    NWPSampleKey.nwp,
-                    SatelliteSampleKey.satellite_actual,
-                ],
-            }
 
         # Check for required keys
         for key in config.get("required_keys", []):
             if key not in self._data:
                 raise ValueError(f"Missing required key: {key}")
 
+        # Start with configured expected shapes
         expected_shapes = {}
-        if config.get("expected_shapes"):
+        if hasattr(config, "expected_shapes"):
+            expected_shapes.update(config.expected_shapes)
+        elif isinstance(config, dict) and "expected_shapes" in config:
             expected_shapes.update(config["expected_shapes"])
 
         # Calculate expected shapes from config if available
-        if "input_data" in config:
-            if GSPSampleKey.gsp not in expected_shapes and "gsp" in config["input_data"]:
-                gsp_config = config["input_data"]["gsp"]
+        if hasattr(config, "input_data"):
+            input_data = config.input_data
+
+            # Calculate GSP shape
+            if hasattr(input_data, "gsp") and input_data.gsp is not None:
+                gsp_config = input_data.gsp
                 time_span = (
-                    gsp_config["interval_end_minutes"] -
-                    gsp_config["interval_start_minutes"]
+                    gsp_config.interval_end_minutes -
+                    gsp_config.interval_start_minutes
                 )
-                resolution = gsp_config["time_resolution_minutes"]
+                resolution = gsp_config.time_resolution_minutes
                 expected_length = (time_span // resolution) + 1
                 expected_shapes[GSPSampleKey.gsp] = (expected_length,)
 
             # Calculate NWP shape
-            if "nwp" in config["input_data"]:
-                for provider in config["input_data"]["nwp"].values():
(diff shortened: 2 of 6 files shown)


ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +24,139 @@ def __init__(self, data: NumpySample) -> None:
...
+            }
+
+        # Check for required keys
+        for key in config.get("required_keys", []):
+            if key not in self._data:
+                validation_result["valid"] = False
+                validation_result["errors"].append(f"Missing required key: {key}")
+
+        # Check data shapes
+        for key, expected_shape in config.get("expected_shapes", {}).items():
```
peterdudfield, line 86: do you think its possible to calcualte the expected shape, rather than enter it in the config? Should be able to make it from the history_minutes e.t.c

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +24,139 @@ def __init__(self, data: NumpySample) -> None:
...
+            config: Optional configuration dict with expected shapes and required fields.
+                If None, uses default validation rules.
+
+        Returns:
+            dict: Validation results with status and any validation errors.
+        """
+        validation_result = {
+            "valid": True,
+            "errors": [],
+        }
```
peterdudfield, line 64: I can see why you have done this, im be tempted to first of all keep it simple, and just assert errors

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -58,14 +177,60 @@ def plot(self) -> None:
...
 
         if SatelliteSampleKey.satellite_actual in self._data:
             axes[1, 0].imshow(self._data[SatelliteSampleKey.satellite_actual])
             axes[1, 0].set_title("Satellite Data")
 
         plt.tight_layout()
         plt.show()
+
+
+def validate_samples(
```
peterdudfield, line 193: I'd probably be tempted to strip this out for the moment. We should just be able to use the code above to validate.

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -68,7 +68,8 @@ def validate_sample(self, config: dict | None = None) -> dict:
 
         # Check for required keys
         for key in config.get("required_keys", []):
-            assert key in self._data, f"Missing required key: {key}"
+            if key not in self._data:
+                raise ValueError(f"Missing required key: {key}")
```
AUdaltsova, line 72: Just curious, is this because ValueError is more informative than an assert error?

felix-e-h-p: It can be yeah - more so linting requires the ValueError as opposed to standard assertion

AUdaltsova: Oh okay! Thanks :)

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +22,167 @@ def __init__(self, data: NumpySample) -> None:
...
+                    SatelliteSampleKey.satellite_actual,
+                ],
+            }
+
+        # Check for required keys
+        for key in config.get("required_keys", []):
+            if key not in self._data:
+                raise ValueError(f"Missing required key: {key}")
+
+        expected_shapes = config.get("expected_shapes", {}).copy()
```
peterdudfield, line 74: start from `expected_shapes={}`

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +22,167 @@ def __init__(self, data: NumpySample) -> None:
...
+
+            # Calculate NWP shape
+            if "nwp" in config["input_data"]:
+                for provider in config["input_data"]["nwp"].values():
+                    if isinstance(provider, dict) and "image_size_pixels_height" in provider:
+                        config["nwp_shape"] = (
+                            provider["image_size_pixels_height"],
+                            provider["image_size_pixels_width"],
+                        )
+                        break
```
peterdudfield, line 96: good to have different expected shapes for different providers

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +22,167 @@ def __init__(self, data: NumpySample) -> None:
...
+
+            # Calculate satellite shape
+            if "satellite" in config["input_data"]:
+                sat_config = config["input_data"]["satellite"]
+                has_height = "image_size_pixels_height" in sat_config
+                has_width = "image_size_pixels_width" in sat_config
```
peterdudfield, line 102: if this is always, then just use

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +22,167 @@ def __init__(self, data: NumpySample) -> None:
...
+                    satellite_shape = (
+                        time_steps,
+                        channels,
+                        sat_config["image_size_pixels_height"],
+                        sat_config["image_size_pixels_width"],
+                    )
+                    config["satellite_shape"] = satellite_shape
+
+        # Check data shapes against calculated or config given shapes
+        for key, expected_shape in expected_shapes.items():
```
peterdudfield, line 120: explicitly write this for GSP

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +22,167 @@ def __init__(self, data: NumpySample) -> None:
...
+            if key in self._data:
+                actual_shape = self._data[key].shape
+
+                # Check dimensions match
```
peterdudfield, line 124: could make check for dimmesionas and check a function, then gets re used in GSP, Satellite and NWP

tests/torch_datasets/sample/test_uk_regional_sample.py
```diff
@@ -111,3 +111,90 @@ def test_to_numpy(numpy_sample):
     assert "nwp" in numpy_data
     assert isinstance(numpy_data["nwp"]["ukv"]["nwp"], np.ndarray)
     assert numpy_data[GSPSampleKey.gsp].shape == (7,)
+
+
+# Validation config file as fixture
+@pytest.fixture
+def validation_config_file(tmp_path):
```
peterdudfield, line 118: remove if not used

tests/torch_datasets/sample/test_uk_regional_sample.py
```diff
@@ -111,3 +111,90 @@ def test_to_numpy(numpy_sample):
...
+def test_validate_sample(numpy_sample):
+    """Test the validate_sample method with default config"""
+    sample = UKRegionalSample(numpy_sample)
+
+    # Assert validation structure
+    result = sample.validate_sample()
+    assert result is True
+
+
+def test_validate_sample_with_custom_config(numpy_sample):
```
peterdudfield, line 125: def test_validate_sample_with_custom_config(configuration)

# change nwp window to 3,3,
...

sample = UKRegionalSample(numpy_sample, config)
result = sample.validate_sample()

tests/torch_datasets/sample/test_uk_regional_sample.py
```diff
@@ -111,3 +111,90 @@ def test_to_numpy(numpy_sample):
...
+    # Create copy of sample without satellite data
+    modified_data = {
+        "nwp": numpy_sample["nwp"],
+        GSPSampleKey.gsp: numpy_sample[GSPSampleKey.gsp],
+    }
+
+    sample = UKRegionalSample(modified_data)
+
+    # Create config with satellite data
+    config = {
```
peterdudfield, line 154: removed

tests/torch_datasets/sample/test_uk_regional_sample.py
```diff
@@ -111,3 +111,90 @@ def test_to_numpy(numpy_sample):
...
+def test_validate_sample_with_missing_keys(numpy_sample):
+    """Test validation with missing required keys"""
+    # Create copy of sample without satellite data
+    modified_data = {
+        "nwp": numpy_sample["nwp"],
+        GSPSampleKey.gsp: numpy_sample[GSPSampleKey.gsp],
+    }
+
+    sample = UKRegionalSample(modified_data)
+
```
peterdudfield, line 152: sample['satellite_actually'].pop() - removes satelltie data

felix-e-h-p:
Updates made - ready to merge if all OK

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +22,186 @@ def __init__(self, data: NumpySample) -> None:
...
+            path: Path to the saved sample file.
+
+        Returns:
+            A UKRegionalSample instance with the loaded data.
+        """
         # Loads from .pt format
         # TODO: We should move away from using torch.load(..., weights_only=False)
         return cls(torch.load(path, weights_only=False))
 
+    def _check_dimensions(
```
peterdudfield, line 49: i'd probably move it to a seperate function, potentially out of this file

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +22,186 @@ def __init__(self, data: NumpySample) -> None:
...
+        # Check each dimension matches
+        zipped_dims = zip(actual_shape, expected_shape, strict=False)
+        for i, (actual_dim, expected_dim) in enumerate(zipped_dims):
+            if expected_dim is not None and actual_dim != expected_dim:
+                raise ValueError(
+                    f"{name} shape mismatch at dimension {i}: "
+                    f"expected {expected_dim}, got {actual_dim}",
+                )
+
+    def validate_sample(self, config: dict | None = None) -> dict:
```
peterdudfield, line 81: I reckon config should be a requirement argument, and its `config:Configuration`

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +22,186 @@ def __init__(self, data: NumpySample) -> None:
...
+            }
+
+        # Check for required keys
+        for key in config.get("required_keys", []):
+            if key not in self._data:
+                raise ValueError(f"Missing required key: {key}")
+
+        expected_shapes = {}
+        if config.get("expected_shapes"):
+            expected_shapes.update(config["expected_shapes"])
```
peterdudfield, line 108: for me config is a COnfiguration, so we should add extra things like expected shape into it, or get it from there

ocf_data_sampler/torch_datasets/sample/uk_regional.py
peterdudfield: see https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/config/model.py#L317

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +22,186 @@ def __init__(self, data: NumpySample) -> None:
...
+
+            # Calculate NWP shape
+            if "nwp" in config["input_data"]:
+                for provider in config["input_data"]["nwp"].values():
+                    if isinstance(provider, dict) and "image_size_pixels_height" in provider:
+                        config["nwp_shape"] = (
+                            provider["image_size_pixels_height"],
+                            provider["image_size_pixels_width"],
+                        )
+
```
peterdudfield, line 130: im not sure `expected_shapes` is added to here

peterdudfield: I think is alays in provider so you dont need that check

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +22,186 @@ def __init__(self, data: NumpySample) -> None:
...
+                resolution = sat_config["time_resolution_minutes"]
+                time_steps = (time_span // resolution) + 1
+
+                satellite_shape = (
+                    time_steps,
+                    channels,
+                    sat_config["image_size_pixels_height"],
+                    sat_config["image_size_pixels_width"],
+                )
+                config["satellite_shape"] = satellite_shape
```
peterdudfield, line 147: nice

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +22,186 @@ def __init__(self, data: NumpySample) -> None:
...
+                if "nwp" not in provider_data:
+                    raise ValueError(f"Missing 'nwp' key in NWP data for {provider}")
+
+                # Check expected shape of NWP data if configured
+                if config.get("nwp_shape") and "nwp" in provider_data:
+                    nwp_array = provider_data["nwp"]
+                    spatial_dims = nwp_array.shape[-2:]
+
+                    if tuple(spatial_dims) != config.get("nwp_shape"):
+                        raise ValueError(
```
peterdudfield, line 174: why dont use you use `check_dimensions` function

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +22,186 @@ def __init__(self, data: NumpySample) -> None:
...
+        # Validate satellite data
+        if SatelliteSampleKey.satellite_actual in self._data:
+            sat_data = self._data[SatelliteSampleKey.satellite_actual]
+
+            if len(sat_data.shape) < 2:
+                raise ValueError(
+                    f"Satellite data should have at least 2 dimensions, got shape {sat_data.shape}",
+                )
+
+            if config.get("satellite_shape"):
```
peterdudfield, line 189: should you get it form `expected_shape` not config

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +23,154 @@ def __init__(self, data: NumpySample) -> None:
...
+
+        Args:
+            config: Optional configuration dict with expected shapes and required fields.
+                If None, uses default validation rules.
+
+        Returns:
+            dict: Validation results with status and any validation errors.
+        """
+        # Default config if none
+        if config is None:
```
dfulu, line 61: I think it would be better if the config wasn't optional. We can't really assume that any of these sources are generally included - even GSPSampleKey.gsp which is not included in the inference samples

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +23,154 @@ def __init__(self, data: NumpySample) -> None:
...
+            path: Path to the saved sample file.
+
+        Returns:
+            A UKRegionalSample instance with the loaded data.
+        """
         # Loads from .pt format
         # TODO: We should move away from using torch.load(..., weights_only=False)
         return cls(torch.load(path, weights_only=False))
 
+    def validate_sample(self, config: Configuration | None = None) -> dict:
```
dfulu, line 50: The type hint here is a `Configuration` object but in the tests and the code in the function you allow it to be a general dictionary. Either the type hint here should be updated to what you expect it to be or we should be strict that this needs to be a `Configuration` object

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +23,154 @@ def __init__(self, data: NumpySample) -> None:
...
+                raise ValueError(f"Missing required key: {key}")
+
+        expected_shapes = {}
+        if config.get("expected_shapes"):
+            expected_shapes.update(config["expected_shapes"])
+
+        # Calculate expected shapes from config if available
+        if "input_data" in config:
+            if GSPSampleKey.gsp not in expected_shapes and "gsp" in config["input_data"]:
+                gsp_config = config["input_data"]["gsp"]
```
dfulu, line 82: If config here should be a `Configuration` object, I think its cleaner to use the form `config.input_data.gsp` like we stick to elsewhere

ocf_data_sampler/config/model.py
```diff
@@ -319,3 +319,17 @@ class Configuration(Base):
 
     general: General = General()
     input_data: InputData = InputData()
+    expected_shapes: dict = Field(default_factory=dict)
+
+    def get_expected_shapes(self) -> dict:
```
peterdudfield, line 324: I dont think this should be in config,

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +26,108 @@ def __init__(self, data: NumpySample) -> None:
...
+
+        Args:
+            config: Optional configuration dict with expected shapes and required fields.
+                If None, uses default validation rules.
+
+        Returns:
+            dict: Validation results with status and any validation errors.
+        """
+        # Check for required keys
+        for key in config.get("required_keys", []):
```
peterdudfield, line 64: is required_keys in the config? im not sure it is

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +26,108 @@ def __init__(self, data: NumpySample) -> None:
...
+            if key not in self._data:
+                raise ValueError(f"Missing required key: {key}")
+
+        # Calculate expected shapes from configuration
+        expected_shapes = calculate_expected_shapes(config)
```
peterdudfield, line 69: nice, i liked this

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +26,108 @@ def __init__(self, data: NumpySample) -> None:
...
+
+        # Check GSP shape if specified
```
peterdudfield, line 71: i like all this below

ocf_data_sampler/torch_datasets/utils/validation_utils.py
```diff
@@ -0,0 +1,102 @@
...
+            f"(Actual shape: {actual_shape}, Expected shape: {expected_shape})",
+        )
+
+    zipped_dims = zip(actual_shape, expected_shape, strict=True)
+    for i, (actual_dim, expected_dim) in enumerate(zipped_dims):
+        if expected_dim is not None and actual_dim != expected_dim:
+            raise ValueError(f"{name} shape mismatch at dimension {i}")
+
+
+def calculate_expected_shapes(
```
peterdudfield, line 37: ncie its a seperate function

tests/torch_datasets/sample/test_uk_regional_sample.py
```diff
@@ -48,15 +48,13 @@ def pvnet_config_filename(tmp_path):
...
     """Synthetic data generation"""
-
     # Field / spatial coordinates
     nwp_data = {
-        "nwp": np.random.rand(4, 1, 2, 2),
-        "x": np.array([1, 2]),
-        "y": np.array([1, 2]),
+        "nwp": np.random.rand(4, 1, 3, 3),
+        "x": np.array([1, 2, 3]),
+        "y": np.array([1, 2, 3]),
```
peterdudfield, line 55: im not sure why this has changed?

tests/torch_datasets/sample/test_uk_regional_sample.py
```diff
@@ -66,6 +64,58 @@ def numpy_sample():
...
+    """Configuration with expected shapes for testing"""
+    return {
+        "expected_shapes": {
+            GSPSampleKey.gsp: (7,),
+        },
+    }
+
+
+@pytest.fixture
+def input_data_configuration():
```
peterdudfield, line 91: i would use a configuration from conftest already

tests/torch_datasets/sample/test_uk_regional_sample.py
```diff
@@ -111,3 +161,64 @@ def test_to_numpy(numpy_sample):
...
+    # Use config that requires satellite data
+    config_with_satellite = custom_configuration.copy()
+    config_with_satellite["required_keys"] = [
+        GSPSampleKey.gsp,
+        NWPSampleKey.nwp,
+        SatelliteSampleKey.satellite_actual,
+    ]
+
+    with pytest.raises(ValueError, match="Missing required key: satellite_actual"):
+        sample.validate_sample(config_with_satellite)
```
peterdudfield, line 212: nice, maybe dont need "required_keys" as discussed above

tests/torch_datasets/sample/test_uk_regional_sample.py
```diff
@@ -111,3 +161,64 @@ def test_to_numpy(numpy_sample):
...
+def test_validate_sample_with_wrong_shapes(numpy_sample, shape_configuration):
+    """Test validation with incorrect data shapes"""
+    # Create copy of sample with wrong GSP shape
+    modified_data = numpy_sample.copy()
+    modified_data[GSPSampleKey.gsp] = np.random.rand(10)
+
+    sample = UKRegionalSample(modified_data)
+
+    with pytest.raises(ValueError, match="GSP shape mismatch at dimension 0"):
+        sample.validate_sample(shape_configuration)
```
peterdudfield, line 224: nice

ocf_data_sampler/torch_datasets/utils/validation_utils.py
```diff
@@ -0,0 +1,121 @@
...
+                num_non_accum = len(non_accum_channels)
+                num_accum = len(provider_config.accum_channels)
+
+                num_channels = num_non_accum + num_accum
+                height = provider_config.image_size_pixels_height
+                width = provider_config.image_size_pixels_width
+                nwp_shapes[provider_key] = (num_timesteps, num_channels, height, width)
+
+            if nwp_shapes:
+                expected_shapes[NWPSampleKey.nwp] = nwp_shapes
```
peterdudfield, line 101: expected_shapes[NWPSampleKey.nwp][provider_key]

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +26,116 @@ def __init__(self, data: NumpySample) -> None:
...
+            path: Path to the saved sample file.
+
+        Returns:
+            A UKRegionalSample instance with the loaded data.
+        """
         # Loads from .pt format
         # TODO: We should move away from using torch.load(..., weights_only=False)
         return cls(torch.load(path, weights_only=False))
 
+    def validate_sample(self, config: Configuration | None = None) -> dict:
```
peterdudfield, line 53: should `config` be required? it errors if its None now

ocf_data_sampler/torch_datasets/utils/validation_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    expected_shapes = {}
+
+    if config is None:
+        return expected_shapes
+
+    # Start with configured expected shapes
+    if hasattr(config, "expected_shapes"):
+        expected_shapes.update(config.expected_shapes)
+    elif isinstance(config, dict) and "expected_shapes" in config:
+        expected_shapes.update(config["expected_shapes"])
```
peterdudfield, line 58: should we remove this bit now?

peterdudfield (changes requested):
a few small changes, then I think we are there

ocf_data_sampler/torch_datasets/sample/uk_regional.py
felix-e-h-p: Yeah wasn't sure with this - updated so always a pre-requirement

ocf_data_sampler/torch_datasets/utils/validation_utils.py
felix-e-h-p: Yeah - felt perhaps could be worthwhile if any additional tests without Config objects, but suppose not needed technically right

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +26,115 @@ def __init__(self, data: NumpySample) -> None:
...
+        # Check satellite shape if data exists and is expected
+        if sat_key in self._data and sat_key in expected_shapes:
+            sat_data = self._data[sat_key]
+            if len(sat_data.shape) < 2:
+                raise ValueError(
+                    f"Satellite data ('{sat_key}') should have at least 2 dimensions, "
+                    f"got shape {sat_data.shape}",
+                )
+            actual_spatial_dims = sat_data.shape[-2:]
+            expected_spatial_dims = expected_shapes[sat_key][-2:]
```
peterdudfield, line 125: why -2 here?

peterdudfield (approved):
Looks great, ive left one other comment. 
Perhaps get @dfulu and @AUdaltsova to look at it too

ocf_data_sampler/torch_datasets/sample/uk_regional.py
felix-e-h-p: Ah just for spatial dimensions only - not full tensor shape. Can definitely upgrade/change though.

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +26,116 @@ def __init__(self, data: NumpySample) -> None:
...
+            nwp_data_all_providers = self._data[nwp_key]
+            if not isinstance(nwp_data_all_providers, dict):
+                raise ValueError(f"NWP data ('{nwp_key}') should be a dictionary.")
+
+            # Loop through providers present in actual data
+            for provider, provider_data in nwp_data_all_providers.items():
+                if "nwp" not in provider_data:
+                    raise ValueError(f"Missing array key in NWP data for provider '{provider}'.")
+
+                if nwp_key in expected_shapes and provider in expected_shapes[nwp_key]:
```
AUdaltsova, line 99: Maybe this has been discussed already, but I wonder if it would be useful to raise a warning if an unexpected provider is found in a sample? I don't think we want to fail here, especially given we often leave extra providers in the sample and then "turn them off" in the model config, but I think it might be useful for bug-catching to know if there is somehow an unexpected data source in your sample. Same for non-NWP sources as well.

I can see that this might be excessive (as in, it's really unlikely dsampler will add something it was not asked to), but for example if there is an accidental config/sample mismatch it will weed it out. But if you guys think that's too much that's fine

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +26,116 @@ def __init__(self, data: NumpySample) -> None:
...
+            path: Path to the saved sample file.
+
+        Returns:
+            A UKRegionalSample instance with the loaded data.
+        """
         # Loads from .pt format
         # TODO: We should move away from using torch.load(..., weights_only=False)
         return cls(torch.load(path, weights_only=False))
 
+    def validate_sample(self, config: Configuration) -> dict:
```
AUdaltsova, line 53: Depending on how we envision this being used, is it maybe easier to pass a config path instead of a config object? I think from a user perspective doing something like
```
sample = UKRegionalSample.load(sample_path)
sample.validate(config_path)
```
is more intuitive than
```
import load_yaml_configuration

sample = UKRegionalSample.load(sample_path)
config = load_yaml_configuration(config_path)
sample.validate(config)
```

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +26,116 @@ def __init__(self, data: NumpySample) -> None:
...
+            actual_spatial_dims = sat_data.shape
+            expected_spatial_dims = expected_shapes[sat_key]
+
+            check_dimensions(
+                actual_shape=tuple(actual_spatial_dims),
+                expected_shape=expected_spatial_dims,
+                name="Satellite spatial dims",
+            )
+
+        return True
```
AUdaltsova, line 134: Nitpicking here, but if this is supposed to be a message for the user that validation was successful, I think it could be more informative/obvious

ocf_data_sampler/torch_datasets/utils/validation_utils.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+
+    zipped_dims = zip(actual_shape, expected_shape, strict=True)
+    for i, (actual_dim, expected_dim) in enumerate(zipped_dims):
+        if expected_dim is not None and actual_dim != expected_dim:
+            raise ValueError(f"{name} shape mismatch at dimension {i}")
+
+
+def calculate_expected_shapes(
+    config: Configuration | dict | None = None,
```
AUdaltsova, line 38: I think we now require an actual Configuration object to be passed to validate_sample() so maybe dict and None are not needed anymore/this should be required as well? But maybe there is a use-case I'm not seeing

ocf_data_sampler/torch_datasets/utils/validation_utils.py
```diff
@@ -0,0 +1,122 @@
...
+
+                non_accum_channels = [
+                    c for c in provider_config.channels
+                    if c not in provider_config.accum_channels
+                ]
+
+                num_non_accum = len(non_accum_channels)
+                num_accum = len(provider_config.accum_channels)
+
+                num_channels = num_non_accum + num_accum
```
AUdaltsova, line 92: I would assume `len(provider_config.channels)` would give you the same number, since `accum_channels` should just be a subset of `channels`? No?

ocf_data_sampler/torch_datasets/utils/validation_utils.py
```diff
@@ -0,0 +1,122 @@
...
+                gsp_config.interval_start_minutes
+            )
+            resolution = gsp_config.time_resolution_minutes
+            expected_length = (time_span // resolution) + 1
+            expected_shapes[GSPSampleKey.gsp] = (expected_length,)
+
+        # Calculate NWP shape
+        if hasattr(input_data, "nwp") and input_data.nwp is not None:
+            if NWPSampleKey.nwp not in expected_shapes:
+                expected_shapes[NWPSampleKey.nwp] = {}
```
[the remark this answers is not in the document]
felix-e-h-p: Ah nah - only will be added to depending on provider.

ocf_data_sampler/torch_datasets/utils/validation_utils.py
felix-e-h-p: Yeah 'num_channels = len(provider_config.channels)' - suppose just as a 'safe measure' right, depending on config utilised. But definitely, updated.

ocf_data_sampler/torch_datasets/sample/uk_regional.py
felix-e-h-p: Yeah definitely - relatively basic message included now

ocf_data_sampler/torch_datasets/sample/uk_regional.py
felix-e-h-p: Yeah, definitely see this. Do you think an object could be better on the basis of clearer separation i.e. just validation as opposed to loading also? Quite like having any in/out dependencies away from main validation logic.

ocf_data_sampler/torch_datasets/sample/uk_regional.py
peterdudfield: yea, the batches often have extra providers in right. 

I would probably only raise an error in the provider in the config is not in the batch, and allow the batch to have extra things in it

peterdudfield: yea maybe a warning if there is a provider in the sample, but not in the config. This could be done in a different PR though

ocf_data_sampler/torch_datasets/sample/uk_regional.py
AUdaltsova: Yeah had the same thought, honesty very minor preference for me, can stay as is

ocf_data_sampler/torch_datasets/utils/validation_utils.py
```diff
@@ -0,0 +1,122 @@
...
+        Dictionary mapping data keys to their expected shapes (tuples of integers).
+    """
+    expected_shapes = {}
+
+    if config is None:
+        return expected_shapes
+
+    # Calculate expected shapes from input_data if available
+    if hasattr(config, "input_data"):
+        input_data = config.input_data
```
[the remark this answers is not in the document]
AUdaltsova: Actually I think we do need to fail hard if there is no input data. Currently, here it will skip calculation and return an empty dict, and because we don't chack for extra keys and everything is conditioned on being in expected_shapes in validate_sample, it will return success and check nothing for an empty config if I'm reading this right.

ocf_data_sampler/torch_datasets/utils/validation_utils.py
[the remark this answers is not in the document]
AUdaltsova: I still think the if statement can be removed, lmk if I'm missing something

AUdaltsova (changes requested):
Looks super good, thanks so much for doing this! Requesting changes for the sake of the input_data check, sorry if I'm reading it wrong!

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +26,112 @@ def __init__(self, data: NumpySample) -> None:
...
+    def validate_sample(self, config: Configuration) -> dict:
+        """Validates that the sample has the expected structure and data shapes.
+
+        Args:
+            config: Configuration dict with expected shapes and required fields.
+
+        Returns:
+            dict: Validation results with status and any validation errors.
+        """
+        if config is None:
```
dfulu, line 62: This seems like a strange check now. The type hint is that the config needs to be a `Configuration` object. Why check its in case it is `None`? A user equally could have set it to a string path or anything. I think this should go or we should assert the config is a `Configuration` object

ocf_data_sampler/torch_datasets/utils/validation_utils.py
```diff
@@ -0,0 +1,110 @@
...
+        config: Configuration object or dictionary with shape information.
+            If None, returns an empty dictionary.
+
+    Returns:
+        Dictionary mapping data keys to their expected shapes (tuples of integers).
+    """
+    expected_shapes = {}
+
+    # Calculate expected shapes from input_data if available
+    if hasattr(config, "input_data"):
```
dfulu, line 52: I think we could probably cut out this if statement. The `Configuration` object always has an `input_data` attribute, even if it is None.

ocf_data_sampler/torch_datasets/utils/validation_utils.py
```diff
@@ -0,0 +1,110 @@
...
+            time_span = (
+                gsp_config.interval_end_minutes -
+                gsp_config.interval_start_minutes
+            )
+            resolution = gsp_config.time_resolution_minutes
+            expected_length = (time_span // resolution) + 1
+            expected_shapes[GSPSampleKey.gsp] = (expected_length,)
+
+        # Calculate NWP shape
+        if hasattr(input_data, "nwp") and input_data.nwp is not None:
```
dfulu, line 67: Similar to the above, I think the Configuration.input_data object always has an `nwp` attribute even if it is None - [see here](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/config/model.py#L311)

dfulu: We could remove some of these hasattr checks for simplicity

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +26,112 @@ def __init__(self, data: NumpySample) -> None:
...
+        # Check satellite shape if data exists and is expected
+        if sat_key in self._data and sat_key in expected_shapes:
+            sat_data = self._data[sat_key]
+            check_dimensions(
+                actual_shape=sat_data.shape,
+                expected_shape=expected_shapes[sat_key],
+                name="Satellite data",
+            )
+
+        result = {
```
dfulu, line 125: What is the purpose of this returned object? I don't think it's used anywhere and it doesn't match the return description of containing status and validation errors. I think it might just be simpler to return True or None depending on how we want to frame the function

ocf_data_sampler/torch_datasets/sample/uk_regional.py
```diff
@@ -22,22 +26,107 @@ def __init__(self, data: NumpySample) -> None:
...
+            gsp_data = self._data[gsp_key]
+            check_dimensions(
+                actual_shape=gsp_data.shape,
+                expected_shape=expected_shapes[gsp_key],
+                name="GSP",
+            )
+
+        # Checks for NWP data - nested structure
+        nwp_key = NWPSampleKey.nwp
+        if nwp_key in expected_shapes and expected_shapes[nwp_key] and nwp_key not in self._data:
```
dfulu, line 85: I think it would be cleaner to not let `expected_shapes[nwp_key]`  be an empty dictionary if the config doesn't have NWP in it. I think `calculate_expected_shapes()` should be amended for this and then we can get rid of `expected_shapes[nwp_key]` from this if statement

dfulu (approved):

PR #238, opened 2025-04-15
Fix bug for incorrect dropout applied to generation

zakwatts:
# Pull Request

## Description

The previous code removed all generation after the time t0 for the case where no dropout is applied. 

This resulted in only times before the specified dropout to be kept, and the rest to be kept as NaNs.

This fix changes the result of no dropout to None so that all the generation values are kept.

## How Has This Been Tested?

This change has been tested locally via printing the results to ensure they are being kept.

- [X] Yes

Printed results

- [X] Yes

## Checklist:

- [] I have made corresponding changes to the documentation
- [] I have added tests that prove my fix is effective or that my feature works


zakwatts:
Code already merged so closing

PR #241, opened 2025-04-20
Validate channels

### LINKED ISSUE #236: Validate accum_channels is a subset of channels in config model
We expect `accum_channels` to be a subset of `channels` for any NWP source: after loading [we select only `channels`](https://github.com/openclimatefix/ocf-data-sampler/blob/6c171a4e88b81a4a2e80c6d981f36a2a6782fe54/ocf_data_sampler/load/load_dataset.py#L30), so if something is in `accum_channels` but not in `channels`, it will have no data to diff. I think we should validate that in the config.

Solution:
Add a validation function to the [NWP class](https://github.com/openclimatefix/ocf-data-sampler/blob/6c171a4e88b81a4a2e80c6d981f36a2a6782fe54/ocf_data_sampler/config/model.py#L189) in the config that checks that what is passed to `accum_channels` is a subset of what is passed to `channels`.

RupeshMangalam21:
# Pull Request

## Description

This PR adds validation to ensure `accum_channels` specified in the NWP configuration are a subset of the configured `channels`. This prevents runtime errors when attempting to compute differences for channels not present in the data.

Key changes:
- Added Pydantic validator to check `accum_channels ⊂ channels`
- Added required normalization constant checks for diff_ prefixed channels
- Added comprehensive unit tests for both valid and invalid cases

Fixes #236

## How Has This Been Tested?

Added 2 new test cases in `tests/config/test_config.py`:
1. `test_valid_accum_channels`: Verifies valid configuration with:
   - accum_channels subset of channels
   - Proper diff_ normalization constants
2. `test_invalid_accum_channels`: Checks proper ValueError is raised when:
   - accum_channels contains non-existent channel
   - Error message specifies missing channels

Tests run via pytest with command:
```bash
pytest tests/config/test_config.py -k accum_channels
```

## Checklist:

### DIFF
--- a/ocf_data_sampler/config/model.py
+++ b/ocf_data_sampler/config/model.py
@@ -7,7 +7,7 @@ Prefix with a protocol like s3:// to read from alternative filesystems.
 from collections.abc import Iterator
 from typing import Literal
 
-from pydantic import BaseModel, Field, RootModel, ValidationInfo, field_validator, model_validator
+from pydantic import BaseModel, Field, RootModel, field_validator, model_validator
 from typing_extensions import override
 
 NWP_PROVIDERS = [
@@ -212,16 +212,16 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
         " the maximum forecast horizon of the NWP and the requested forecast length.",
     )
 
-    @field_validator("accum_channels")
-    def validate_accum_channels_subset(cls, v: list[str], values: ValidationInfo) -> list[str]:
+    @model_validator(mode="after")
+    def validate_accum_channels_subset(self) -> "NWP":
         """Validate accum_channels is subset of channels."""
-        channels = values.data.get("channels", [])
-        invalid_channels = set(v) - set(channels)
+        invalid_channels = set(self.accum_channels) - set(self.channels)
         if invalid_channels:
             raise ValueError(
-                f"accum_channels contains channels not present in 'channels': {invalid_channels}",
+                f"NWP provider '{self.provider}': accum_channels contains "
+                f"invalid channels: {invalid_channels}",
             )
-        return v
+        return self
 
     @field_validator("provider")
     def validate_provider(cls, v: str) -> str:
--- a/ocf_data_sampler/config/model.py
+++ b/ocf_data_sampler/config/model.py
@@ -218,8 +218,8 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
         invalid_channels = set(self.accum_channels) - set(self.channels)
         if invalid_channels:
             raise ValueError(
-                f"NWP provider '{self.provider}': accum_channels contains "
-                f"invalid channels: {invalid_channels}",
+                f"NWP provider '{self.provider}': all values in 'accum_channels' should "
+                f"be present in 'channels'. Extra values found: {invalid_channels}",
             )
         return self
 
--- a/ocf_data_sampler/config/model.py
+++ b/ocf_data_sampler/config/model.py
@@ -7,7 +7,7 @@ Prefix with a protocol like s3:// to read from alternative filesystems.
 from collections.abc import Iterator
 from typing import Literal
 
-from pydantic import BaseModel, Field, RootModel, field_validator, model_validator
+from pydantic import BaseModel, Field, RootModel, ValidationInfo, field_validator, model_validator
 from typing_extensions import override
 
 NWP_PROVIDERS = [
@@ -211,6 +211,17 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
         " used to construct an example. If set to None, then the max staleness is set according to"
         " the maximum forecast horizon of the NWP and the requested forecast length.",
     )
+
+    @field_validator("accum_channels")
+    def validate_accum_channels_subset(cls, v: list[str], values: ValidationInfo) -> list[str]:
+        """Validate accum_channels is subset of channels."""
+        channels = values.data.get("channels", [])
+        invalid_channels = set(v) - set(channels)
+        if invalid_channels:
+            raise ValueError(
+                f"accum_channels contains channels not present in 'channels': {invalid_channels}",
+            )
+        return v
 
     @field_validator("provider")
     def validate_provider(cls, v: str) -> str:
--- a/tests/config/test_config.py
+++ b/tests/config/test_config.py
@@ -135,8 +135,10 @@ def test_accum_channels_validation(test_config_filename):
 
     # Verify exact error message
     expected_error = (
-        rf"NWP provider '{nwp_name}': accum_channels contains channels "
-        rf"not found in 'channels': {{'invalid_channel'}}"
+        r"input_data.nwp.ukv\n"
+        r"  Value error, NWP provider 'ukv': all values in 'accum_channels' "
+        r"should be present in 'channels'\. "
+        r"Extra values found: {'invalid_channel'}.*"
     )
-    with pytest.raises(ValueError, match=expected_error):
+    with pytest.raises(ValidationError, match=expected_error):
         _ = Configuration(**invalid_config.model_dump())
--- a/tests/config/test_config.py
+++ b/tests/config/test_config.py
@@ -135,7 +135,7 @@ def test_accum_channels_validation(test_config_filename):
 
     # Verify exact error message
     expected_error = (
-        r"input_data.nwp.ukv\n"
+        rf"input_data.nwp.{nwp_name}\n"
         fr"  Value error, NWP provider '{nwp_name}': all values in 'accum_channels' "
         r"should be present in 'channels'\. "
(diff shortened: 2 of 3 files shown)


ocf_data_sampler/config/model.py
```diff
@@ -211,6 +211,17 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
...
 
+    @field_validator("accum_channels")
+    def validate_accum_channels_subset(cls, v: list[str], values: ValidationInfo) -> list[str]:
+        """Validate accum_channels is subset of channels."""
+        channels = values.data.get("channels", [])
+        invalid_channels = set(v) - set(channels)
+        if invalid_channels:
+            raise ValueError(
+                f"accum_channels contains channels not present in 'channels': {invalid_channels}",
+            )
```
AUdaltsova, line 222: Good error message! I'd add the provider here maybe so it's super clear if that's easy to do?

tests/config/test_config.py
```diff
@@ -120,3 +120,39 @@ def test_inconsistent_dropout_use(test_config_filename):
...
+    new_config_dict["input_data"]["nwp"][nwp_name]["accum_channels"] = [target_channel]
+
+    norm_constants = new_config_dict["input_data"]["nwp"][nwp_name]["normalisation_constants"]
+    norm_constants[f"diff_{target_channel}"] = {
+        "mean": 0.0,  # Example values
+        "std": 1.0,
+    }
+
+    _ = Configuration(**new_config_dict) #should validate
+
```
AUdaltsova, line 144: I don't think this test is needed. I would change the test config to have an accum_channel instead

tests/config/test_config.py
```diff
@@ -120,3 +120,39 @@ def test_inconsistent_dropout_use(test_config_filename):
...
+def test_invalid_accum_channels(test_config_filename):
+    """Test accum_channels with non-existent channel raises error."""
+    configuration = load_yaml_configuration(test_config_filename)
+
+    nwp_name = next(iter(configuration.input_data.nwp.root.keys()))
```
AUdaltsova, line 149: Nice! Good that the provider is not hard-coded

AUdaltsova (changes requested):
Hi @RupeshMangalam21, great work & thanks for doing this, left a couple comments.

ocf_data_sampler/config/model.py
```diff
@@ -211,6 +211,17 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
         " the maximum forecast horizon of the NWP and the requested forecast length.",
     )
 
+    @field_validator("accum_channels")
+    def validate_accum_channels_subset(cls, v: list[str], values: ValidationInfo) -> list[str]:
+        """Validate accum_channels is subset of channels."""
+        channels = values.data.get("channels", [])
```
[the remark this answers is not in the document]
RupeshMangalam21: Good catch! You're right - since 'channels' is mandatory, we shouldn't need the empty list fallback. I've switched to a model_validator which is cleaner and matches our other validation patterns.

ocf_data_sampler/config/model.py
RupeshMangalam21: Great suggestion! I've updated the error message to include the provider name upfront - this will make debugging much easier when working with multiple NWP sources.

tests/config/test_config.py
RupeshMangalam21: You make a good point - it's cleaner to test against real-world configs. I'll add a proper accum_channel to our test YAML, include its diff_ normalization constants and keep just the invalid case test for explicit validation checking.

ocf_data_sampler/config/model.py
```diff
@@ -212,16 +212,16 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
...
+    def validate_accum_channels_subset(self) -> "NWP":
         """Validate accum_channels is subset of channels."""
-        channels = values.data.get("channels", [])
-        invalid_channels = set(v) - set(channels)
+        invalid_channels = set(self.accum_channels) - set(self.channels)
         if invalid_channels:
             raise ValueError(
-                f"accum_channels contains channels not present in 'channels': {invalid_channels}",
+                f"NWP provider '{self.provider}': accum_channels contains "
+                f"invalid channels: {invalid_channels}",
```
AUdaltsova, line 222:
```suggestion
                f"NWP provider '{self.provider}': all values in 'accum_channels' should "
                f"be present in 'channels'. Extra values found: {invalid_channels}",
```

I'm nitpicking but I think this is clearer, as technically the channels are not invalid, they should just be present in channels also, and this tells you how to fix it.

Great job adding the provider in! Looks good now

tests/config/test_config.py
```diff
@@ -122,37 +122,27 @@ def test_inconsistent_dropout_use(test_config_filename):
...
+    config = load_yaml_configuration(test_config_filename)
+    nwp_name, nwp_config = next(iter(config.input_data.nwp.root.items()))
+
+    # Test valid case with normalization constants
+    valid_config = config.model_copy(deep=True)
+    valid_nwp = valid_config.input_data.nwp.root[nwp_name]
+    valid_channel = valid_nwp.channels[0]
+    valid_nwp.accum_channels = [valid_channel]
+    valid_nwp.normalisation_constants[f"diff_{valid_channel}"] = {"mean": 0.0, "std": 1.0}
+    _ = Configuration(**valid_config.model_dump())  # Should pass
```
AUdaltsova, line 136: Sorry, I think I wasn't clear before. When I said to add accum channels to the config, I meant to the actual config file that the tests use, I think we have two of those, just check which one `test_config_filename` points to. And then you won't need this bit, because `load_yaml_configuration` does all the validation under the hood when it returns a `Configuration` object, so if it works the checks pass.

tests/config/test_config.py
```diff
@@ -122,37 +122,27 @@ def test_inconsistent_dropout_use(test_config_filename):
...
+
+    # Test invalid channel
+    invalid_config = config.model_copy(deep=True)
+    invalid_nwp = invalid_config.input_data.nwp.root[nwp_name]
+    invalid_nwp.accum_channels = ["invalid_channel"]
+    with pytest.raises(ValueError) as err:
```
AUdaltsova, line 142: To be super pedantic here I would also match the text to the expected error message (I think this is what you had here before?), this way you can be sure the ValueError it's catching is what you expect and not something else failing unexpectedly

RupeshMangalam21: Good call! I switched to using match= with the exact error message. Now it checks the full error text, including the provider and invalid channels. Thanks for pushing for precision here!

tests/config/test_config.py
RupeshMangalam21: Got it! I ditched the code hacks and updated the actual test YAML file with proper accum_channels and diff_ values. Now validation happens naturally during parsing. Thanks for steering me toward the right fix!

tests/test_data/configs/test_config.yaml
```diff
@@ -19,6 +19,7 @@ input_data:
       time_resolution_minutes: 60
       channels:
         - t
+      accum_channels: ["t"]
```
AUdaltsova, line 22:
```suggestion
      accum_channels:
        - t
```

tests/config/test_config.py
```diff
@@ -120,3 +120,25 @@ def test_inconsistent_dropout_use(test_config_filename):
...
+    nwp_name, _ = next(iter(config.input_data.nwp.root.items()))
+
+    # Test invalid channel scenario
+    invalid_config = config.model_copy(deep=True)
+    invalid_nwp = invalid_config.input_data.nwp.root[nwp_name]
+    invalid_nwp.accum_channels = ["invalid_channel"]
+
+    # Verify exact error message
+    expected_error = (
+        r"input_data.nwp.ukv\n"
```
AUdaltsova, line 138:
```suggestion
        rf"input_data.nwp.{nwp_name}\n"
```
Will something like this work? It seems strange to hardcode the source when it's not hardcoded in the test itself (which in my opinion is the way to do it btw - we do hardcode it in other places, but I think this is better practice)

tests/config/test_config.py
```diff
@@ -120,3 +120,25 @@ def test_inconsistent_dropout_use(test_config_filename):
...
+        r"  Value error, NWP provider 'ukv': all values in 'accum_channels' "
```
AUdaltsova, line 139:
```suggestion
        fr"  Value error, NWP provider '{nwp_name}': all values in 'accum_channels' "
```

AUdaltsova:
Hi @RupeshMangalam21, it's been a while & I think the tiny change to the test is the only thing left to do - are you happy for me to just commit the suggestions I left and merge?

AUdaltsova:
Great work on this by the way, thanks for taking this on! Looks very good.

AUdaltsova:
@RupeshMangalam21 no worries, thanks for the great work on this!

AUdaltsova (approved):

Issue #228, opened 2025-04-08
add anon=True into utils, when opening zarr, probably add something in configuration


peterdudfield:
It would be nice if ocf-data-sampler could open public s3 data

We `anon=True` into utils, when opening zarr, probably add something in configuration

For example opening `s3://ocf-open-data-pvnet/data/gfs/v3/2023.zarr`

peterdudfield:
Could this be done by setting credientials?

peterdudfield:
Does `d = xr.open_zarr('s3://ocf-open-data-pvnet/data/gfs/v3/2023.zarr')`

RupeshMangalam21:
Hi, @peterdudfield I would like to take up this issue!

dfulu:
> Does `d = xr.open_zarr('s3://ocf-open-data-pvnet/data/gfs/v3/2023.zarr')` work?

@peterdudfield I've done a quick test on this. It does not work if the user has not set their AWS credentials. It does work if they have set their credentials. 

A user can set credentials by:
- Installing the AWS client (can be done with `pip install awscli`)
- Creating an access key within their AWS account
- login by running `aws configure` in the terminal and saving the credentials

Then they can open the dataset

Else it can be opened without credentials via:

```
ds = xr.open_zarr("s3://ocf-open-data-pvnet/data/gfs/v3/2023.zarr", storage_options={'anon': True})
```

Or fsspec can be configured to use anon for s3 via:

```
import fsspec
fsspec.config.conf['s3'] = {'anon': True}

# dataset will now be opened anonymously by default
ds = xr.open_zarr("s3://ocf-open-data-pvnet/data/gfs/v3/2023.zarr")
```

peterdudfield:
oh thanks @dfulu 

@siddharth7113 could you check out the fsspec solution?

siddharth7113:
> > Does `d = xr.open_zarr('s3://ocf-open-data-pvnet/data/gfs/v3/2023.zarr')` work?
> 
> [@peterdudfield](https://github.com/peterdudfield) I've done a quick test on this. It does not work if the user has not set their AWS credentials. It does work if they have set their credentials.
> 
> A user can set credentials by:
> 
> * Installing the AWS client (can be done with `pip install awscli`)
> * Creating an access key within their AWS account
> * login by running `aws configure` in the terminal and saving the credentials
> 
> Then they can open the dataset
> 
> Else it can be opened without credentials via:
> 
> ```
[...]

PR #253, opened 2025-05-09
add public and pass in nwp config to opening nwp

peterdudfield:
# Pull Request

## Description

- Add option to load NWP from public source
- refactored the nwp load, so nwp configruation is passed through instead

#228 

## How Has This Been Tested?

- CI tests

## Checklist:

### DIFF
--- a/ocf_data_sampler/config/model.py
+++ b/ocf_data_sampler/config/model.py
@@ -211,6 +211,7 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
         " used to construct an example. If set to None, then the max staleness is set according to"
         " the maximum forecast horizon of the NWP and the requested forecast length.",
     )
+    public: bool = Field(False, description="Whether the NWP data is public or private")
 
     @field_validator("provider")
     def validate_provider(cls, v: str) -> str:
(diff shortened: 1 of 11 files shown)


siddharth7113:
Sure

dfulu:
@peterdudfield we have been purposely avoiding passing configs into these basic functions. This has been part of our mob programming sessions on data sampler and I'd like to keep it that way. I think it makes for more reusable code if we can open the data locally using the functions in this library without first having to construct a config object

peterdudfield:
> @peterdudfield we have been purposely avoiding passing configs into these basic functions. This has been part of our mob programming sessions on data sampler and I'd like to keep it that way. I think it makes for more reusable code if we can open the data locally using the functions in this library without first having to construct a config object

Fair enough, there's defiantely pros and cons of passing configs into these samples, as you say. 

For this particular problem, would you prefer passing through a `public` bool? Perhaps it only has to go through the nwp gfs

dfulu:
> For this particular problem, would you prefer passing through a `public` bool? Perhaps it only has to go through the nwp gfs

Yeh I think a bool adding a boolean param would be a good way to go

dfulu (approved):
Looks good, just one linting suggestion

PR #260, opened 2025-05-15
remove stored index pairs

dfulu:
# Pull Request

## Description

Currently in the UK regional dataset we have an attribute `index_pairs` which is an array of all of the possible (t0_index, location_index) pairs. This can be a very large array when training on a few years of data. I created a version of the dataset from 2019-2022 and this this array was 300MB! When we wrap the dataset in a pytorch DataLoader and use multiple workers this creates a copy of this index array in each worker. So for a typical 32 workers this can be 10GB of index pairs

Storing this index array was a design pattern copied across from `ocf_datapipes` and we actually don't need it here.

This PR removes the `index_pairs` array and calculates the values the values that would be in each index of `index_pairs` on the fly

## Checklist:

### DIFF
--- a/ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
+++ b/ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
@@ -1,6 +1,5 @@
 """Torch dataset for UK PVNet."""
 
-import numpy as np
 import pandas as pd
 import xarray as xr
 from torch.utils.data import Dataset
--- a/ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
+++ b/ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
@@ -257,22 +257,12 @@ class PVNetUKRegionalDataset(AbstractPVNetUKDataset):
         # Construct a lookup for locations - useful for users to construct sample by GSP ID
         location_lookup = {loc.id: loc for loc in self.locations}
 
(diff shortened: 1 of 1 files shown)


ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
```diff
@@ -290,7 +279,13 @@ def _get_sample(self, t0: pd.Timestamp, location: Location) -> NumpySample:
     @override
     def __getitem__(self, idx: int) -> NumpySample:
         # Get the coordinates of the sample
-        t_index, loc_index = self.index_pairs[idx]
+
+        if idx >= len(self):
+            raise ValueError(f"Index {idx} out of range for dataset of length {len(self)}")
+
+        t_index = idx % len(self.valid_t0_times)
+        loc_index = idx // len(self.valid_t0_times)
```
Sukh-P, line 287: This is great, really elegant! It took me a second to understand why it works, could be useful to have a comment above which explains that it works because there are t0_location pairs with `len(self.valid_t0_times)` of pairs for each location

Sukh-P (approved):

ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
dfulu: Yeh, sure. I had to do some notebook work to figure this out so definitely worth a comment

Issue #254, opened 2025-05-12
Use xarray-tensorstore for ~30x speed-up


dfulu:
I've done some initial speed tests which show that moving to xarray tensorstore could speed up our sample creation by 10s of times! This may be fast enough for us to no longer need to pre-save samples to train on - instead we could stream samples directly from the zarr. 

The speed up could be from 3-4s per sample down to 120ms per sample. I've added an [example notebook on a temporary branch](https://github.com/openclimatefix/ocf-data-sampler/blob/tensorstore_test/notebooks/datasampler_tensorstore.ipynb) to show my working for this.

Pre-saving samples is a massive overhead to developing models and doing experiments. It also costs us a lot in computing budget since we need to store these pre-saved samples which can be 20TB to train the day-ahead summation model in the UK. Also having batches fast enough to stream could make our backtests trivially fast and even allow us to do cross-validation for the backtests!

There are some considerations for this change:

  1. Loss of some lazy operations: There are some operations in xarray which can be done lazily via dask. For example using the `xarray.concat()` function to merge datasets is done lazily when using the dask backend but with `xarray-tensorstore` objects causes it to eagerly load all of the data. There is an example work around for this issue in the [notebook](https://github.com/openclimatefix/ocf-data-sampler/blob/tensorstore_test/notebooks/datasampler_tensorstore.ipynb) **There may be more cases of this different behaviour**. 
  This means we will need to be careful to avoid these gotchas. This may increase the burden of writing code, but it comes with such a big performance difference that I think it is worth it. It would also simplify our code in places since we would no longer need to work with pre-saved samples. 

 2. `xarray-tensorstore` is really a stub of a library. Its only 300 lines of code and simply wraps tensorstore and xarray together. We cannot be sure that it will be supported. However, with it being only 300 lines of code we could in-house it if we need to or take on the maintenance ourselves. In the [xarray-tensorstore readme](https://github.com/google/xarray-tensorstore?tab=readme-ov-file#xarray-tensorstore) it says that the creator (who is at google and is one of the core `xarray` devs) has found it to be much faster in some cases that `xarray`- so hopefully they will be incentivised to keep it up to date. In the `xarray-tensorstore` readme there is a warning that the xarray API itself might change to allow different backend to be incorporated more easily.
 
 3. Will it work with `torch.DataLoader`? We had some problems in getting the dask multiprocessing and the torch Dataloader multiprocessing. We should check if tensorstore can work together with DataLoader

PR #262, opened 2025-05-20
Xarray tensorstore backend

dfulu:
# Pull Request

## Description

This experimental PR adds xarray-tensorstore as a backend for the satellite data and the UKV and ECMWF NWP data. This leads to a significant speed up

Part of #254

## Checklist:

### DIFF
--- a/ocf_data_sampler/load/nwp/providers/utils.py
+++ b/ocf_data_sampler/load/nwp/providers/utils.py
@@ -2,9 +2,12 @@
 
 import xarray as xr
 
+from ocf_data_sampler.load.open_tensorstore_zarrs import open_zarr, open_zarrs
 
 def open_zarr_paths(
-    zarr_path: str | list[str], time_dim: str = "init_time", public: bool = False,
+    zarr_path: str | list[str], 
+    time_dim: str = "init_time", 
+    public: bool = False,
 ) -> xr.Dataset:
     """Opens the NWP data.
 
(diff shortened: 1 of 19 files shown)


ocf_data_sampler/load/nwp/providers/utils.py
```diff
@@ -41,3 +66,12 @@ def open_zarr_paths(
             **general_kwargs,
         )
     return ds
+
+
+def _tensostore_open_zarr_paths(zarr_path: str | list[str], time_dim: str) -> xr.Dataset:
+    if isinstance(zarr_path, list | tuple):
```
Sukh-P, line 72:
```suggestion
    if isinstance(zarr_path, list | tuple) or "*" in str(zarr_path):
```
Do we also want to support "*" wildcards in case there are many files and we don't want to list them all?

dfulu: Yeh sure, I can add this

ocf_data_sampler/load/open_tensorstore_zarrs.py
```diff
@@ -0,0 +1,92 @@
...
+    if context is None:
+        context = ts.Context()
+
+    ds = xr.open_mfdataset(
+        paths,
+        concat_dim=concat_dim,
+        combine="nested",
+        mask_and_scale=mask_and_scale,
+        decode_timedelta=True,
+    )
```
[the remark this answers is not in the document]
dfulu: Yeh that's a good idea, but I think the implementation need s to be slightly different than this. I'll add

Sukh-P (approved):
Great work on adding this in 🚀 ! Just a thought these changes might warrant a minor version change

ocf_data_sampler/load/open_tensorstore_zarrs.py
```diff
@@ -0,0 +1,92 @@
...
+        write: Whether to open the stores for writing.
+    """
+    arrays_list = []
+    for path in paths:
+        specs = {k: _zarr_spec_from_path(os.path.join(path, k)) for k in data_vars}
+        array_futures = {
+          k: ts.open(spec, read=True, write=write, context=context)
+          for k, spec in specs.items()
+        }
+        arrays_list.append({k: v.result() for k, v in array_futures.items()})
```
shoyer, line 40: Ideally you might call `result()` in a separate loop, so all the arrays can be opened simultaneously.

Issue #246, opened 2025-04-25
more tests


peterdudfield:
It would be great to have the following new tests
- [x] #247
- [x] Add test for site dataset (check we site config)
- [x] Ensure tests functions cover site cases (might will only cover gsp ones at the moment)
- [x] Check test using PVnet config

dfulu:
@peterdudfield could this be closed now?

PR #264, opened 2025-05-23
More tests - site focus

felix-e-h-p:
Related to https://github.com/openclimatefix/ocf-data-sampler/issues/246

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,7 @@ import dask.array
 import numpy as np
 import pandas as pd
 import pytest
+import hashlib
 import xarray as xr
 
 from ocf_data_sampler.config import load_yaml_configuration, save_yaml_configuration
@@ -39,6 +40,9 @@ uk_sat_area_string = """msg_seviri_rss_3km:
 def test_config_filename():
     return f"{_top_test_directory}/test_data/configs/test_config.yaml"
 
+@pytest.fixture(scope="session")
+def site_test_config_path():
+    return f"{_top_test_directory}/test_data/configs/site_test_config.yaml"
 
 @pytest.fixture(scope="session")
 def config_filename():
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -322,66 +326,87 @@ def ds_uk_gsp():
...
+            generation_data = np.empty((len(times), 0), dtype=np.float32)
+            capacity_kwp_data = np.empty((len(times), 0), dtype=np.float32)
+
+        coords = (("time_utc", times), ("site_id", site_ids))
+        da_cap = xr.DataArray(capacity_kwp_data, coords=coords)
+        da_gen = xr.DataArray(generation_data, coords=coords)
+
+        meta_df = pd.DataFrame()
+        meta_df["site_id"] = site_ids
+        if num_sites > 0 :
```
peterdudfield, line 385: can we assume `num_sites >0` so we dont need this if statement

tests/conftest.py
```diff
@@ -322,66 +326,87 @@ def ds_uk_gsp():
...
+        if num_sites == 0:
+            capacity_kwp_1d = np.array([])
+            longitude = np.array([])
+            latitude = np.array([])
+        elif num_sites <= len(base_capacity_kwp_1d):
+            capacity_kwp_1d = base_capacity_kwp_1d[:num_sites]
+            longitude = base_longitude[:num_sites]
+            latitude = base_latitude[:num_sites]
+        else:
+            factor = (num_sites // len(base_capacity_kwp_1d)) + 1
```
peterdudfield, line 366: do we need this? Whats the maximum number of sites we need? Might be better to keep it simple

tests/conftest.py
```diff
@@ -411,24 +436,99 @@ def pvnet_config_filename(
...
+):
+    config = load_yaml_configuration(config_filename)
+
+    config.input_data.site = default_data_site_model
+
+    if hasattr(config.input_data, "gsp"):
+        config.input_data.gsp = None
+
+    if not hasattr(config.input_data, "solar_position") or config.input_data.solar_position is None:
+        config.input_data.solar_position = SolarPosition(
```
peterdudfield, line 488: why not put these just in the test yaml file

tests/conftest.py
felix-e-h-p: Not entirely sure - suppose keeping it as simple as possible would just be:

capacity_kwp_1d = base_capacity_kwp_1d[:num_sites]
longitude = base_longitude[:num_sites]
latitude = base_latitude[:num_sites]

Have removed the first if statement and updated, please let me know if you feel better to be updated to the three lines above

tests/conftest.py
```diff
@@ -322,66 +326,74 @@ def ds_uk_gsp():
...
+        site_interval_end_minutes: int = 60,
+        site_time_resolution_minutes: int = 30,
+    ) -> Site:
+
+        param_tuple = (num_sites, start_time_str, end_time_str, time_freq,
+                       site_interval_start_minutes, site_interval_end_minutes,
+                       site_time_resolution_minutes)
+        param_key = hashlib.sha256(str(param_tuple).encode()).hexdigest()
+
+        if param_key in created_files:
```
peterdudfield, line 347: I would strip out `created_files` for the moment, until we need it

tests/conftest.py
peterdudfield: yea I would go simple like you have suggested. 
unless you there is already a case for the `num_sites > len(base_capacity_kwp_1d)` but i dont think there is

tests/conftest.py
```diff
@@ -411,24 +436,99 @@ def pvnet_config_filename(
...
+    return data_sites()
+
 
 @pytest.fixture()
 def sites_dataset(site_config_filename):
     return SitesDataset(site_config_filename)
+
+
+@pytest.fixture()
+def site_config_derived_from_pvnet_base(
```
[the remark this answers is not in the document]
peterdudfield: what does `PVNet derived config` mean? 

In general I would try to make sure ocf-data-sample doesnt need anything to do with PVNet.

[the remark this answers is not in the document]
peterdudfield: Would it be better to just make a yaml file for the configuration below? Or dont we have one already?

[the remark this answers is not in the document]
felix-e-h-p: Ah sorry - so with the pvnet_test_config.yaml already pre-included - can integrate the below configuration into that

Issue #258, opened 2025-05-14
Update testing coverage for tests/load/


felix-e-h-p:
I.e. for test_load_sites, to further include: Type verification / checking and potential missing value checks.

I feel the inclusion of this within this specific testing subdirectory is fundamentally more robust - whether involving migration of other similar testing or just to be added as additional.

PR #278, opened 2025-06-06
Test load update

felix-e-h-p:
Relating directly to https://github.com/openclimatefix/ocf-data-sampler/issues/258

Type and value assertions added for the four load testing files

### DIFF
--- a/ocf_data_sampler/load/nwp/nwp.py
+++ b/ocf_data_sampler/load/nwp/nwp.py
@@ -10,6 +10,69 @@ from ocf_data_sampler.load.nwp.providers.icon import open_icon_eu
 from ocf_data_sampler.load.nwp.providers.ukv import open_ukv
 
 
+def _validate_nwp_data(data_array: xr.DataArray, provider: str) -> None:
+    """Validates the structure and data types of a loaded NWP DataArray.
+
+    This helper function is extracted to keep the main `open_nwp` function clean.
+
+    Args:
+        data_array: The xarray.DataArray to validate.
+        provider: The NWP provider name.
+
+    Raises:
+        TypeError: If the data or any coordinate has an unexpected dtype.
+        ValueError: If a required coordinate is missing.
+    """
+    if not np.issubdtype(data_array.dtype, np.number):
+        raise TypeError(f"NWP data for {provider} should be numeric, not {data_array.dtype}")
+
+    base_expected_dtypes = {
+        "init_time_utc": np.datetime64,
+        "step": np.timedelta64,
+    }
+
+    provider_specific_dtypes = {
+        "ukv": {
+            "channel": np.str_,
+            "x_osgb": np.floating,
+            "y_osgb": np.floating,
+        },
+        "ecmwf": {
+            "channel": np.str_,
+            "latitude": np.floating,
+            "longitude": np.floating,
+        },
+        "icon-eu": {
+            "channel": np.object_,
+            "latitude": np.floating,
+            "longitude": np.floating,
+        },
+        "cloudcasting": {
+            "channel": np.str_,
+            "x_geostationary": np.floating,
+            "y_geostationary": np.floating,
+        },
+        "gfs": {
+            "channel": np.str_,
+            "latitude": np.floating,
+            "longitude": np.floating,
+        },
+    }
+
+    expected_dtypes = {**base_expected_dtypes, **provider_specific_dtypes.get(provider, {})}
+
+    for coord, expected_dtype in expected_dtypes.items():
+        if coord not in data_array.coords:
+            raise ValueError(f"Coordinate '{coord}' missing for provider '{provider}'")
+        if not np.issubdtype(data_array.coords[coord].dtype, expected_dtype):
+            actual_dtype = data_array.coords[coord].dtype
+            err_msg = (
+                f"'{coord}' for {provider} should be {expected_dtype.__name__}, "
+                f"not {actual_dtype}"
+            )
+            raise TypeError(err_msg)
+
+
 def open_nwp(
     zarr_path: str | list[str],
     provider: str,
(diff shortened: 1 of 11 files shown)


tests/load/test_load_gsp.py
```diff
@@ -29,4 +30,24 @@ def test_open_gsp(uk_gsp_zarr_path):
...
     assert da.shape == (49, 318)
 
+    assert np.issubdtype(da.dtype, np.floating)
 
+    assert np.issubdtype(da.coords["time_utc"].dtype, np.datetime64)
+    assert np.issubdtype(da.coords["gsp_id"].dtype, np.integer)
+    assert np.issubdtype(da.coords["nominal_capacity_mwp"].dtype, np.floating)
+    assert np.issubdtype(da.coords["effective_capacity_mwp"].dtype, np.floating)
+    assert np.issubdtype(da.coords["x_osgb"].dtype, np.floating)
+    assert np.issubdtype(da.coords["y_osgb"].dtype, np.floating)
```
Sukh-P, line 40: I am wondering if these assertions/checks would be better of in the load functions themselves and then the tests updated to check that bad data is caught correctly

tests/load/test_load_gsp.py
```diff
@@ -29,4 +30,24 @@ def test_open_gsp(uk_gsp_zarr_path):
...
+
+    assert not da.isnull().any()
```
Sukh-P, line 42: If you check if anything is null here are the checks for nulls on the coordinates not redundant?

ocf_data_sampler/load/nwp/nwp.py
```diff
@@ -9,39 +10,104 @@
...
+            "longitude": np.integer,
+        }
+    elif provider == "icon-eu":
+        expected_dtypes = {
+            "init_time_utc": np.datetime64,
+            "step": np.timedelta64,
+            "channel": np.object_,
+            "latitude": np.floating,
+            "longitude": np.floating,
+        }
```
Sukh-P, line 82: If you left out checking this one, all the rest are the same in terms of type checking

Sukh-P (changes requested):
Thanks for updating this and the work you have put in already, think it could do with a few changes still, left some comments, thanks

ocf_data_sampler/load/nwp/nwp.py
```diff
@@ -9,39 +10,103 @@
...
+            "channel": np.str_,
+            "x_osgb": np.floating,
+            "y_osgb": np.floating,
+        },
+        "ecmwf": {
+            "channel": np.str_,
+            "latitude": np.floating,
+            "longitude": np.floating,
+        },
+        "icon-eu": {
```
Sukh-P, line 45: I would add `mo_global` to this list and also just have them all just follow the same expected datatypes, so keep the channel as a string for all, this gets rid a lot of customisation as they will al follow the same validation data types

tests/conftest.py
```diff
@@ -179,8 +179,8 @@ def ds_nwp_ecmwf():
     init_times = pd.date_range(start="2023-01-01 00:00", freq="6h", periods=24 * 7)
     steps = pd.timedelta_range("0h", "14h", freq="1h")
 
-    lons = np.arange(-12, 3)
-    lats = np.arange(48, 60)
+    lons = np.arange(-12, 3, dtype=np.float32)
```
Sukh-P, line 182: Seem to have changed things in two different ways to get floats here, maybe just do 1?

tests/load/test_load_nwp.py
```diff
@@ -34,4 +37,140 @@ def test_load_cloudcasting(nwp_cloudcasting_zarr_path):
...
+
+
+def test_load_ukv_bad_dtype(bad_ukv_zarr):
+    """Test validation fails for UKV with bad dtype by reading from a temp file."""
+    with pytest.raises(TypeError, match="'step' for ukv should be timedelta64"):
+        open_nwp(zarr_path=bad_ukv_zarr, provider="ukv")
+
+
+@pytest.fixture
+def bad_ecmwf_zarr(tmp_path) -> str:
```
Sukh-P, line 77: I would just have a set of tests for one of the providers, since all the rest of the validations are the same or pretty similar I think that's enough coverage

ocf_data_sampler/load/nwp/nwp.py
```diff
@@ -9,23 +10,91 @@
...
+        provider: The NWP provider name.
+
+    Raises:
+        TypeError: If the data or any coordinate has an unexpected dtype.
+        ValueError: If a required coordinate is missing.
+    """
+    if not np.issubdtype(data_array.dtype, np.number):
+        raise TypeError(f"NWP data for {provider} should be numeric, not {data_array.dtype}")
+
+    base_expected_dtypes = {
```
Sukh-P, line 29: Do we need base and provider specific anymore as they are all the same, can we not have one set for all of them and reduce duplicated code?

tests/conftest.py
Sukh-P: This hasn't been addressed or answered

ocf_data_sampler/load/nwp/nwp.py
```diff
@@ -34,14 +85,14 @@ def open_nwp(zarr_path: str | list[str], provider: str, public: bool = False) ->
         _open_nwp = open_icon_eu
     elif provider == "gfs":
         _open_nwp = open_gfs
-
-        # GFS has a public/private flag
```
Sukh-P, line 38: Should leave this comment in, it's useful

tests/conftest.py
felix-e-h-p: Thought perhaps more robust to do at stages where float64 is defaulted to, also to perhaps be more efficient in that sense. Suppose just controlling the datatype at point of creation right. Have now as a single conversion if preferred.

Sukh-P (approved):

PR #290, opened 2025-07-16
added a check for site/gsp data with pydantic

### LINKED ISSUE #286: gsp/site data in the configuration file
## description
the model primarily needs either gsp or site data to work, it would be good to prompt the user to add one of those incase that field is left empty

zaryab-ali:
# Pull Request

## Description
a check was added to to see if either gsp or site data is provided

Fixes #286 

## How Has This Been Tested?
it was tested by running the save_samples script from pvnet with different configuration in example_configuration, 
- [x] Yes

## Checklist:

### DIFF
--- a/tests/config/test_save.py
+++ b/tests/config/test_save.py
@@ -7,7 +7,6 @@ from ocf_data_sampler.config import load_yaml_configuration, save_yaml_configura
 
 def test_save_yaml_configuration_basic(tmp_path, test_config_gsp_path):
     """Save an empty configuration object"""
-    #config = Configuration(**load_yaml_configuration(test_config_gsp_path).model_dump())
     config = load_yaml_configuration(test_config_gsp_path)
 
     filepath = f"{tmp_path}/config.yaml"


zaryab-ali:
@Sukh-P could you have a look at this please
i added a check that throws an error if both site and gsp are missing or commented
i know you guys need set gsp to null for inference but i think that edge case is already handled,  if gsp is not commented but zarr_path is set to null, pydantic throws an error

ocf_data_sampler/config/model.py
```diff
@@ -334,9 +334,18 @@ class InputData(Base):
...
 
+    @model_validator(mode="after")
+    def check_site_or_gsp(self) -> "InputData":
+        if self.site is None and self.gsp is None:
+            raise ValueError(
+                "You must provide either `site` or `gsp` in the `input_data`"
+            )
+
+        return self
+
```
Sukh-P, line 345: Nice!

ocf_data_sampler/config/model.py
```diff
@@ -334,9 +334,18 @@ class InputData(Base):
...
 
 class Configuration(Base):
     """Configuration model for the dataset."""
 
     general: General = General()
-    input_data: InputData = InputData()
+    input_data: InputData = Field(default_factory=InputData)
```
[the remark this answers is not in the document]
zaryab-ali: so when you run the code without this change, pydantic checks for data before it is loaded and an error is thrown that originates at the import statement, because the data is None at that time,

Sukh-P:
@zaryab-ali there are tests still failing which you can see [here](https://github.com/openclimatefix/ocf-data-sampler/actions/runs/16438648588/job/46458763197?pr=290#step:8:598) that's because some preexisting tests don't include either site or GSP so you will need to add some simple dummy config parameters in there to fix this

Sukh-P:
> @Sukh-P sorry about that, i thought those tests were supposed to run as they were most of the tests are using a fake config file [here](https://github.com/openclimatefix/ocf-data-sampler/blob/main/tests/test_data/configs/test_config.yaml), should i use it as well using `load_yaml_configuration`, since it does contain gsp data

Hi Zaryab, so it's good to check which tests are failing [here](https://github.com/openclimatefix/ocf-data-sampler/actions/runs/16438648588/job/46553620162?pr=290#step:8:598) there it lists the 3 failing tests in the logs, from it I can see two of the tests aren't passing because just the default values for configuration are used and in this case both GSP and site are `null` so the validation fails, so it would be good just to add another config file [here](https://github.com/openclimatefix/ocf-data-sampler/tree/main/tests/test_data/configs) with just some dummy site values for example and then use those in the two failing tests, in the [conftest file](https://github.com/openclimatefix/ocf-data-sampler/blob/main/tests/conftest.py#L44) you can see how the config yaml files are converted into python fixtures which can be used in the tests. 

The 3rd test that is failing is due to it not using a test config file, it could use the test config that the other test in this file uses https://github.com/openclimatefix/ocf-data-sampler/blob/main/tests/config/test_save.py#L8

zaryab-ali:
@Sukh-P i think this should be good to go all the checks passed [here](https://github.com/zaryab-ali/ocf-data-sampler/actions/runs/17218751359)
had to add gsp to a yaml where it was previously set to null, but i dont this this will be an issue as the test cases are passing

tests/test_data/configs/site_test_config.yaml
```diff
@@ -40,7 +40,14 @@ input_data:
 
   site: null
 
-  gsp: null
+  gsp:
```
Sukh-P, line 43: It makes less sense to change gsp to non null values here, would be better to change the site values to non null instead I think

zaryab-ali:
@Sukh-P changes made, should be alright now

tests/test_data/configs/gsp_test_config.yaml
```diff
@@ -0,0 +1,12 @@
...
+  name: example
+
+input_data:
+  gsp:
+    zarr_path: tests/data/gsp/test.zarr
+    interval_start_minutes: -60
+    interval_end_minutes: 120
+    time_resolution_minutes: 30
+    dropout_timedeltas_minutes: [-30]
+    dropout_fraction: 0.1
```
Sukh-P, line 12: These don't need to be here because they have defaults so they can be removed and that way it tests the defaults more

tests/test_data/configs/gsp_test_config.yaml
```diff
@@ -0,0 +1,12 @@
+general:
+  description: test example configuration
+  name: example
+
+input_data:
+  gsp:
+    zarr_path: tests/data/gsp/test.zarr
```
Sukh-P, line 7: This can be just an empty string ""

tests/config/test_config.py
```diff
@@ -4,17 +4,18 @@
...
 
-def test_extra_field_error():
+def test_extra_field_error(test_config_gsp_path):
     """
     Check an extra parameters in config causes error
     """
-
-    configuration = Configuration()
+    config = load_yaml_configuration(test_config_gsp_path)
+    configuration = Configuration(**config.model_dump())
```
Sukh-P, line 18: See comment above on loading config

Sukh-P (changes requested):
Thanks Zaryab for the work on this, it feels almost there just needs a bit of tidying up

Sukh-P (approved):
Looks good, thanks for all the updates to this, a useful additional check for new users!

PR #292, opened 2025-07-18
feat: Add Ice Chunk support for high-performance cloud data access

Dakshbir:
## PR Description:

This Pull Request introduces a major new capability to ocf-data-sampler: high-performance satellite data loading directly from cloud storage using the icechunk library and optimized GCS streaming.

This work addresses the bottleneck of downloading large Zarr datasets by enabling a "cloud-native" approach. It provides a robust, version-controlled, and highly paerformant alternative to the existing plain Zarr loader, with benchmarks demonstrating throughput on par with or exceeding 14,000 MB/s.

### Ice Chunk Integration
Adds two new data loaders for satellite data:
- An optimized loader for standard cloud Zarr files
- A "true Ice Chunk" loader that leverages the icechunk library for versioning and transactional reads

### Configuration Model Update
The Satellite configuration model has been extended to support Ice Chunk-specific parameters (`icechunk_path`, `use_true_icechunk`, `icechunk_commit`, etc.).

### Data Conversion Pipeline
A new script is provided to convert existing OCF Zarr datasets into a high-performance Ice Chunk format with a compatible data structure.

### Comprehensive Benchmarking Suite
New scripts and utilities have been added to rigorously benchmark and compare the performance of different data loading strategies.

## File-by-File Breakdown

Here is a description of each new or modified file included in this PR:

#### [`[ocf_data_sampler/config/model.py]`](ocf_data_sampler/config/model.py)
- **Purpose:** Extends the Pydantic configuration model to support Ice Chunk
- **Details:** Adds fields like `icechunk_path`, `use_true_icechunk`, and optimization parameters to the Satellite model, allowing users to configure the new data loaders through YAML files

#### [`[ocf_data_sampler/load/icechunk_optimized.py]`](ocf_data_sampler/load/icechunk_optimized.py)
- **Purpose:** A high-performance loader for standard cloud-hosted Zarr files
- **Details:** Implements best practices for GCS streaming (e.g., gcsfs tuning, Dask configuration) to maximize throughput for Zarr datasets that are not managed by the icechunk library

#### [`[ocf_data_sampler/load/icechunk_optimized_with_ics.py]`](ocf_data_sampler/load/icechunk_optimized_with_ics.py)
- **Purpose:** The primary loader for "true" Ice Chunk repositories
- **Details:** This loader uses the icechunk library to interact with version-controlled datasets, allowing for transactional reads from specific branches or commits. It also handles the conversion of different Zarr data layouts into the format expected by the sampler

#### [`[ocf_data_sampler/load/load_dataset.py]`](ocf_data_sampler/load/load_dataset.py)
- **Purpose:** The central dispatcher for data loading
- **Details:** The logic in `get_dataset_dict` has been updated to act as a router. Based on the provided configuration (`zarr_path` vs. `icechunk_path`), it intelligently selects the appropriate loader (traditional, optimized cloud, or true Ice Chunk) for satellite data

#### [`[ocf_data_sampler/torch_datasets/utils/benchmark.py]`](ocf_data_sampler/torch_datasets/utils/benchmark.py)
- **Purpose:** A new utility module to run and aggregate performance benchmarks
- **Details:** Provides the `OCFDataSamplerBenchmark` class, which programmatically runs loading tests, calculates average throughput, and provides a summary assessment

#### [`[scripts/benchmark_cli.py]`](scripts/benchmark_cli.py) 
- **Purpose:** A command-line interface for the benchmarking utility
- **Details:** Allows users to easily run a performance benchmark against any data source by simply providing a configuration file

#### [`[scripts/full_dataset_icechunk_conversion.py]`](scripts/full_dataset_icechunk_conversion.py)
- **Purpose:** A crucial tool for preparing data for the new pipeline
- **Details:** This script converts an existing OCF Zarr dataset (with its custom ocf_blosc2 codec) into a high-performance Ice Chunk repository. It correctly restructures the data with separate variables per channel and removes incompatible codecs, which is essential for achieving maximum read performance

#### [`[scripts/production_benchmark_comparison.py]`](scripts/production_benchmark_comparison.py)
- **Purpose:** A high-level script for direct performance comparison
- **Details:** This script runs an "apples-to-apples" benchmark comparing the existing plain Zarr loader against the new true Ice Chunk loader, using their respective optimal configurations. It provides a clear, final recommendation based on the results

#### [`[production_icechunk_2024-02_config.yaml]`](production_icechunk_2024-02_config.yaml)
- **Purpose:** An example of a production-ready configuration for the Ice Chunk loader
- **Details:** Demonstrates how to configure the sampler to use a specific Ice Chunk repository, commit ID, and all 11 satellite channels

#### [`[test_plain_zarr_clean.yaml]`](test_plain_zarr_clean.yaml)
- **Purpose:** The baseline configuration for the existing plain Zarr loader
- **Details:** Used in comparison benchmarks to measure the performance of the current production data access method

## Final Benchmark Results

After fixing the data format to enable a true apples-to-apples comparison, the new Ice Chunk loader demonstrates exceptional performance, on par with and sometimes exceeding the existing plain Zarr implementation.

```
Production Comparison Results
========================================
Performance Comparison:
  Plain Zarr: 9261.93 MB/s
  Ice Chunk: 14256.93 MB/s
  Ice Chunk is 1.54x FASTER

Recommendation: Ice Chunk recommended for performance
```
This confirms that the new data access pattern is a viable and high-performance solution ready for production use.

## How to Test This PR
To verify the new functionality and reproduce the benchmark results, please follow these steps.

### Prerequisites:

Ensure all new dependencies are installed (e.g., icechunk).
Google Cloud authentication must be configured locally to access the GCS bucket (gsoc-dakshbir).
The source dataset gs://gsoc-dakshbir/2024-02_nonhrv.zarr must be available for the comparison benchmark.

**Step 1: Run the Main Production Benchmark**
This is the primary test. It runs a direct "apples-to-apples" comparison between the existing plain Zarr loader and the new, optimized Ice Chunk loader using the provided configuration files.

python scripts/production_benchmark_comparison.py

- **What it does:** The script warms up the GCS connection and then runs multiple timed samples for both the test_plain_zarr_clean.yaml and production_icechunk_2024-02_config.yaml configurations.

- **Expected Outcome:** The final output will be a summary table directly comparing the throughput of both methods. The results should be similar to those claimed in the PR description, showing a significant performance increase with the Ice Chunk loader.

### Production Comparison Results
========================================
Performance Comparison:
  Plain Zarr: ~9,200 MB/s
  Ice Chunk: ~14,000 MB/s
  Ice Chunk is ~1.5x FASTER

Recommendation: Ice Chunk recommended for performance

(Note: Absolute throughput may vary slightly based on network conditions, but the relative performance gain should be consistent.)

**Step 2: (Optional) Run an Individual Benchmark with the CLI Tool**
This demonstrates the flexibility of the new benchmark_cli.py utility for testing any configuration.

python scripts/benchmark_cli.py --config production_icechunk_2024-02_config.yaml --samples 3
python scripts/benchmark_cli.py --config test_plain_zarr_clean.yaml --samples 3

- **What it does**: Runs 3 benchmark samples using only the specified configuration.
- **Expected Outcome:** The script will print the average throughput for the icechunk_optimized method, which should be in the multi-GB/s range, confirming the high performance of the new loader independently.

**Step 3: (Optional) Verify the Data Conversion Pipeline**
This step validates the script that converts a standard OCF Zarr dataset into the high-performance Ice Chunk format. The PR already includes a pre-converted dataset specified in production_icechunk_2024-02_config.yaml, so this is for verification purposes.

python scripts/full_dataset_icechunk_conversion.py

- **What it does:** This script reads the 2024-02_nonhrv.zarr dataset, restructures it to have one variable per channel, removes the incompatible ocf_blosc2 codec, and writes it to a new version-controlled Ice Chunk repository in GCS.
- **Expected Outcome**: The script will print progress as it processes batches. Upon completion, it will output the GCS path of the newly created repository, a unique commit ID, and generate a new production_icechunk_..._config.yaml file in your local directory.

### DIFF
--- a/ocf_data_sampler/config/model.py
+++ b/ocf_data_sampler/config/model.py
@@ -1,13 +1,17 @@
 """Configuration model for the dataset.
 
+
 Absolute or relative zarr filepath(s).
 Prefix with a protocol like s3:// to read from alternative filesystems.
 """
 
+
 from collections.abc import Iterator
+
 
 from pydantic import BaseModel, Field, RootModel, field_validator, model_validator
 from typing_extensions import override
+
 
 NWP_PROVIDERS = [
     "ukv",
@@ -18,17 +22,24 @@ NWP_PROVIDERS = [
 ]
 
 
+
+
 class Base(BaseModel):
     """Pydantic Base model where no extras can be added."""
+
 
     class Config:
         """Config class."""
 
+
         extra = "forbid"  # forbid use of extra kwargs
+
+
 
 
 class General(Base):
     """General pydantic model."""
+
 
     name: str = Field("example", description="The name of this configuration file")
     description: str = Field(
@@ -37,8 +48,11 @@ class General(Base):
     )
 
 
+
+
 class TimeWindowMixin(Base):
     """Mixin class, to add interval start, end and resolution minutes."""
+
 
     time_resolution_minutes: int = Field(
         ...,
@@ -46,15 +60,18 @@ class TimeWindowMixin(Base):
         description="The temporal resolution of the data in minutes",
     )
 
+
     interval_start_minutes: int = Field(
         ...,
         description="Data interval starts at `t0 + interval_start_minutes`",
     )
 
+
     interval_end_minutes: int = Field(
         ...,
         description="Data interval ends at `t0 + interval_end_minutes`",
     )
+
 
     @model_validator(mode="after")
     def validate_intervals(self) -> "TimeWindowMixin":
@@ -79,14 +96,18 @@ class TimeWindowMixin(Base):
         return self
 
 
+
+
 class DropoutMixin(Base):
     """Mixin class, to add dropout minutes."""
+
 
     dropout_timedeltas_minutes: list[int] = Field(
         default=[],
         description="List of possible minutes before t0 where data availability may start. Must be "
         "negative or zero.",
     )
+
 
     dropout_fraction: float = Field(
         default=0,
@@ -95,6 +116,7 @@ class DropoutMixin(Base):
         le=1,
     )
 
+
     @field_validator("dropout_timedeltas_minutes")
     def dropout_timedeltas_minutes_negative(cls, v: list[int]) -> list[int]:
         """Validate 'dropout_timedeltas_minutes'."""
@@ -102,6 +124,7 @@ class DropoutMixin(Base):
             if m > 0:
                 raise ValueError("Dropout timedeltas must be negative")
         return v
+
 
     @model_validator(mode="after")
     def dropout_instructions_consistent(self) -> "DropoutMixin":
@@ -115,14 +138,18 @@ class DropoutMixin(Base):
         return self
 
 
+
+
 class SpatialWindowMixin(Base):
     """Mixin class, to add path and image size."""
+
 
     image_size_pixels_height: int = Field(
         ...,
         ge=0,
         description="The number of pixels of the height of the region of interest",
     )
+
 
     image_size_pixels_width: int = Field(
         ...,
@@ -131,15 +158,20 @@ class SpatialWindowMixin(Base):
     )
 
 
+
+
 class NormalisationValues(Base):
     """Normalisation mean and standard deviation."""
     mean: float = Field(..., description="Mean value for normalization")
     std: float = Field(..., gt=0, description="Standard deviation (must be positive)")
 
 
+
+
 class NormalisationConstantsMixin(Base):
     """Normalisation constants for multiple channels."""
     normalisation_constants: dict[str, NormalisationValues]
+
 
     @property
     def channel_means(self) -> dict[str, float]:
@@ -148,6 +180,8 @@ class NormalisationConstantsMixin(Base):
             channel: norm_values.mean
             for channel, norm_values in self.normalisation_constants.items()
         }
+
+
 
 
     @property
@@ -159,19 +193,96 @@ class NormalisationConstantsMixin(Base):
         }
 
 
-class Satellite(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConstantsMixin):
-    """Satellite configuration model."""
 
-    zarr_path: str | tuple[str] | list[str] = Field(
-        ...,
+
+class Satellite(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConstantsMixin):
+    """Satellite configuration model with Ice Chunk support."""
+
+
+    zarr_path: str | tuple[str] | list[str] | None = Field(
+        None,
         description="Absolute or relative zarr filepath(s). Prefix with a protocol like s3:// "
         "to read from alternative filesystems.",
     )
+
+
+    # NEW: Ice Chunk support
+    icechunk_path: str | None = Field(
+        None,
+        description="Remote Ice Chunk dataset path for cloud-native streaming. "
+        "When provided, this will be used instead of zarr_path with 103+ MB/s optimizations.",
+    )
+
+
+    bucket_name: str = Field(
+        "gsoc-dakshbir",
+        description="GCS bucket name for Ice Chunk datasets",
+    )
+
+
+    # NEW: True Ice Chunk settings
+    use_true_icechunk: bool = Field(
+        False,
+        description="Whether to use the true Ice Chunk API for versioning and transactions.",
+    )
+
+
+    icechunk_branch: str | None = Field(
+        "main",
+        description="The Ice Chunk branch to use (e.g., 'main').",
+    )
+
+
+    icechunk_commit: str | None = Field(
+        None,
+        description="The specific Ice Chunk commit ID to use. If provided, it overrides the branch.",
+    )
+
+
+    # NEW: Optimization parameters (from your research)
+    use_optimized_streaming: bool = Field(
+        True,
+        description="Enable 103+ MB/s streaming optimizations for Ice Chunk",
+    )
+
+
+    optimal_time_steps: int = Field(
+        6,
+        description="Optimal time steps per chunk for streaming. "
+        "Default is 6, which is optimal for 11-channel satellite data.",
+    )
+
+
+    optimal_block_size_mb: int = Field(
+        64,
+        description="Optimal GCS block size in MB (default: 64 from optimization research)",
+    )
+
 
     channels: list[str] = Field(
         ...,
         description="the satellite channels that are used",
     )
+
+
+    @model_validator(mode="after")
+    def validate_data_source(self) -> "Satellite":
+        """Validate that either zarr_path or icechunk_path is provided."""
+        if not self.zarr_path and not self.icechunk_path:
+            raise ValueError(
+                "Either 'zarr_path' or 'icechunk_path' must be provided for satellite data"
+            )
+       
+        if self.zarr_path and self.icechunk_path:
+            # If both provided, prefer icechunk_path and warn
+            import warnings
+            warnings.warn(
+                "Both zarr_path and icechunk_path provided. Using icechunk_path for optimized streaming.",
+                UserWarning
+            )
+       
+        return self
+
 
     @model_validator(mode="after")
     def check_all_channel_have_normalisation_constants(self) -> "Satellite":
@@ -186,8 +297,11 @@ class Satellite(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, Normalisation
         return self
 
 
+
+
 class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConstantsMixin):
     """NWP configuration model."""
+
 
     zarr_path: str | tuple[str] | list[str] = Field(
         ...,
@@ -195,14 +309,18 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
         "to read from alternative filesystems.",
     )
 
+
     channels: list[str] = Field(
         ...,
         description="the channels used in the nwp data",
     )
 
+
     provider: str = Field(..., description="The provider of the NWP data")
 
+
     accum_channels: list[str] = Field([], description="the nwp channels which need to be diffed")
+
 
     max_staleness_minutes: int | None = Field(
         None,
@@ -210,6 +328,7 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
         " used to construct an example. If set to None, then the max staleness is set according to"
         " the maximum forecast horizon of the NWP and the requested forecast length.",
     )
+
 
     @field_validator("provider")
     def validate_provider(cls, v: str) -> str:
@@ -219,6 +338,8 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
         return v
 
 
+
+
     @model_validator(mode="after")
     def check_all_channel_have_normalisation_constants(self) -> "NWP":
         """Check that all the channels have normalisation constants."""
@@ -226,12 +347,14 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
         non_accum_channels = [c for c in self.channels if c not in self.accum_channels]
         accum_channel_names = [f"diff_{c}" for c in self.accum_channels]
 
+
         missing_norm_values = set(non_accum_channels) - set(normalisation_channels)
         if len(missing_norm_values)>0:
             raise ValueError(
                 "Normalsation constants must be provided for all channels. Missing values for "
                 f"channels: {missing_norm_values}",
             )
+
 
         missing_norm_values = set(accum_channel_names) - set(normalisation_channels)
         if len(missing_norm_values)>0:
@@ -244,38 +367,50 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
         return self
 
 
+
+
 class MultiNWP(RootModel):
     """Configuration for multiple NWPs."""
 
+
     root: dict[str, NWP]
+
 
     @override
     def __getattr__(self, item: str) -> NWP:
         return self.root[item]
 
+
     @override
     def __getitem__(self, item: str) -> NWP:
         return self.root[item]
+
 
     @override
     def __len__(self) -> int:
         return len(self.root)
 
+
     @override
     def __iter__(self) -> Iterator:
         return iter(self.root)
 
+
     def keys(self) -> Iterator[str]:
         """Returns dictionary-like keys."""
         return self.root.keys()
+
 
     def items(self) -> Iterator[tuple[str, NWP]]:
         """Returns dictionary-like items."""
         return self.root.items()
 
 
+
+
 class GSP(TimeWindowMixin, DropoutMixin):
     """GSP configuration model."""
+
 
     zarr_path: str = Field(
         ...,
@@ -284,8 +419,11 @@ class GSP(TimeWindowMixin, DropoutMixin):
     )
 
 
+
+
 class Site(TimeWindowMixin, DropoutMixin):
     """Site configuration model."""
+
 
     file_path: str = Field(
         ...,
@@ -296,16 +434,22 @@ class Site(TimeWindowMixin, DropoutMixin):
         description="The CSV files describing power system",
     )
 
+
     # TODO validate the netcdf for sites
     # TODO validate the csv for metadata
+
+
 
 
 class SolarPosition(TimeWindowMixin):
     """Solar position configuration model."""
 
 
+
+
 class InputData(Base):
     """Input data model."""
+
 
     satellite: Satellite | None = None
     nwp: MultiNWP | None = None
@@ -314,8 +458,13 @@ class InputData(Base):
     solar_position: SolarPosition | None = None
 
 
+
+
 class Configuration(Base):
     """Configuration model for the dataset."""
 
+
     general: General = General()
     input_data: InputData = InputData()
+
+
--- a/ocf_data_sampler/load/load_dataset.py
+++ b/ocf_data_sampler/load/load_dataset.py
@@ -1,52 +1,95 @@
 """Loads all data sources."""
 
+import logging
 import xarray as xr
 
 from ocf_data_sampler.config import InputData
 from ocf_data_sampler.load import open_gsp, open_nwp, open_sat_data, open_site
 
+logger = logging.getLogger(__name__)
 
-def get_dataset_dict(input_config: InputData) -> dict[str, dict[xr.DataArray] | xr.DataArray]:
-    """Construct dictionary of all of the input data sources.
 
-    Args:
-        input_config: InputData configuration object
-    """
+def get_dataset_dict(
+    input_config: InputData,
+    gsp_ids: list[int] | None = None,
+) -> dict[str, dict[xr.DataArray] | xr.DataArray]:
+    """Construct dictionary of all of the input data sources."""
+    
     datasets_dict = {}
 
     # Load GSP data unless the path is None
     if input_config.gsp and input_config.gsp.zarr_path:
-        da_gsp = open_gsp(zarr_path=input_config.gsp.zarr_path).compute()
+        da_gsp = open_gsp(
+            zarr_path=input_config.gsp.zarr_path,
+            boundaries_version=input_config.gsp.boundaries_version,
+            public=input_config.gsp.public,
+        ).compute()
 
-        # Remove national GSP
-        datasets_dict["gsp"] = da_gsp.sel(gsp_id=slice(1, None))
+        if gsp_ids is None:
+            da_gsp = da_gsp.sel(gsp_id=slice(1, None))
+        else:
+            da_gsp = da_gsp.sel(gsp_id=gsp_ids)
+
+        datasets_dict["gsp"] = da_gsp
 
     # Load NWP data if in config
     if input_config.nwp:
         datasets_dict["nwp"] = {}
         for nwp_source, nwp_config in input_config.nwp.items():
-            da_nwp = open_nwp(nwp_config.zarr_path, provider=nwp_config.provider)
-
+            da_nwp = open_nwp(
+                zarr_path=nwp_config.zarr_path,
+                provider=nwp_config.provider,
+                public=nwp_config.public,
+            )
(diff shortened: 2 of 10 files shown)


devsjc (changes requested):
Firstly, thanks for putting this together Dakshbir! I can see there's been a lot of work that's gone into this, and I know there's more still behind the scenes trial-and-erroring the optimal values and picking up the problem.

There's quite a lot going on in this PR, so I'm going to review it in stages.  I'll suggest several changes over these stages, but please don't take it as a slight on your hard work! It's more getting the code lined up with the organisational structure and layout, so it's as easy as possible for the active maintainers to understand it. 

For this first review stage, I'm ignoring the scripts and the benchmark functions for now and just focussing on the core logic. 

The main changes that Ive suggested are a consolidation or removal of a lot of the new configuration options - I suspect the fewer changes to the config file structure, the better for the ML team. I think we can assume a lot of the values are wanted to be the optimal defaults you've found without allowing explicit configuring - rather leave them as inputs to the functions so they can be changed in code if necessary.

This consolidation into the `zarr_path` requires a bit of new logic to be added in, which is similar to what you have in the `load_dataset` function, but I'm suggeating a regex based approach, and moving those branch conditionals into the `open_sat_data` function.

ocf_data_sampler/load/satellite.py
```diff
@@ -42,35 +60,160 @@ def get_single_sat_data(zarr_path: str) -> xr.Dataset:
 
     return ds
```
devsjc, line 61: [1] I'd be tempted to put the parsing logic in here actually, because then it could extend the existing checks that are occurring. That way all the decision branches are consolidated in one place, and it makes it a bit easier to reason about the flows. Something like:

```python
match _parse_icechunk_path(zarr_path):
    case {"protocol": None, "bucket": _, "prefix": _, "sha1": z}:
        # Raise error if trying to use commit syntax for local file")
        ...
    case {"protocol": "gs", "bucket": _, "prefix": y, "sha1": _} if "*" in y:
        # Raise an error if a wildcard is used in a GCP path
        ...
    case {"protocol": p, "bucket": b, "prefix": y, "sha1": z} if ".icechunk" in y and p is not None:
        # Icechunk logic goes here
        ds = _open_sat_data_icechunk(p, b, y, z) # Also has the benefit of simplifying the signature of this function
    case {"protocol": _, "bucket": x, "prefix": y, "sha1": None}:
        # Existing single zarr logic path here
        ds = xr.open_dataset(zarr_path, engine="zarr", chunks="auto")
    case {"protocol": None, "bucket": x, "prefix": y, "sha1": None} if "*" in y:
        # Handle multi-zarr dataset for as before
        ds = xr.open_mfdataset(...
    case _:
        # Raise error on unhandled path
        ...
``` 

Note this would require updating the regex in _parse_icechunk_path to allow optional protocol and wildcard symbols in the prefix as follows:

```python
pattern = r"^(?:(?P<protocol>[\w]{2,6}):\/\/)?(?P<bucket>[\w\/-]+)\/(?P<prefix>[\w*.\/-]+?)(?:@(?P<sha1>[\w]+))?$"
```

ocf_data_sampler/load/satellite.py
```diff
@@ -42,35 +60,160 @@ def get_single_sat_data(zarr_path: str) -> xr.Dataset:
...
+    components = match.groupdict()
+    # Convert None to None (no change needed, but explicit)
+    return components
 
-    Args:
-      zarr_path: Cloud URL or local path pattern, or list of these. If GCS URL,
-                 it must start with 'gs://'
-    """
-    # Open the data
+def _open_sat_data_icechunk(
```
devsjc, line 75: If [1] is followed, the signature of this function can become a more useful 
```
(protocol: str | None, bucket: str, prefix: str, sha1: str | None)
```

ocf_data_sampler/load/satellite.py
```diff
@@ -42,35 +60,160 @@ def get_single_sat_data(zarr_path: str) -> xr.Dataset:
...
+) -> xr.DataArray:
+    """Open satellite data from an Ice Chunk repository with optimized settings."""
+    logger.info(f"Opening satellite data from Ice Chunk repository: {zarr_path}")
+    _setup_optimal_environment()
+
+    # Parse the path using improved regex
+    path_components = _parse_icechunk_path(zarr_path)
+    bucket = path_components["bucket"]
+    prefix = path_components["prefix"]
+    commit_sha = path_components.get("sha1")
```
devsjc, line 86: [1] would have this logic handled so this could go

ocf_data_sampler/load/satellite.py
```diff
@@ -42,35 +60,160 @@ def get_single_sat_data(zarr_path: str) -> xr.Dataset:
...
+            pass
+        
+        # Method 2: If version doesn't work, try accessing via the main branch 
+        # (since the commit should be the latest on main)
+        if not session:
+            try:
+                session = repo.readonly_session(branch="main")
+                logger.info("Opened main branch (commit should be latest on main)")
+            except Exception as e:
+                raise ValueError(f"Could not open session for commit {commit_sha}: {e}")
```
devsjc, line 129: I think remove this - if someone specifies a commit, then it should error if that commit doesn't exist instead of opening the "main" branch instead.

ocf_data_sampler/load/satellite.py
```diff
@@ -42,35 +60,160 @@ def get_single_sat_data(zarr_path: str) -> xr.Dataset:
...
+    
+    if main_path.endswith((".icechunk", ".ic")):
+        if isinstance(zarr_path, (list, tuple)):
+            raise NotImplementedError(
+                "Loading from multiple Ice Chunk repositories is not currently supported."
+            )
+        return _open_sat_data_icechunk(zarr_path, channels)
+    else:
+        logger.info(f"Opening satellite data from standard Zarr: {zarr_path}")
+        return _open_sat_data_zarr(zarr_path, channels)
```
devsjc, line 219: Apologies, as I think I previously said to put the logic in here; but I now think it is better suited in `_get_single_sat_data` as mentioned in [1] - that way the logic in the `open_sat_data` funtion can remain the same as it was and all the conditional opening logic can be in a single place in `_get_single_sat_data`.

ocf_data_sampler/load/satellite.py
```diff
@@ -42,35 +60,160 @@ def get_single_sat_data(zarr_path: str) -> xr.Dataset:
...
     ds = ds.rename(
         {
             "variable": "channel",
             "time": "time_utc",
         },
     )
-
     check_time_unique_increasing(ds.time_utc)
     ds = make_spatial_coords_increasing(ds, x_coord="x_geostationary", y_coord="y_geostationary")
     ds = ds.transpose("time_utc", "channel", "x_geostationary", "y_geostationary")
```
devsjc, line 183: This should be renamed back to just open_sat_data as the logic splitting between icechunks and regular zarrs gets moved by [1]

devsjc (changes requested):
Thanks for making my requested changes!

I'm now going to revise them a bit in this next set, by suggesting that

- All the conditional logic joins the existing conditionals in `get_single_sat_data` in a new `match case`
- `_open_sat_data_icechunk` changes its signature to reflect the nature of the computation
- `open_sat_data` returns to as-was

I know I previously said to put the logic elsewhere to what I'm saying now, so apologies for being a bit confusing - but as we chip away at the code pathways it becomes easier to spot where to put things sensibly!

ocf_data_sampler/load/satellite.py
Dakshbir: gotcha

ocf_data_sampler/load/satellite.py
```diff
@@ -1,59 +1,230 @@
...
+                "protocol": None,
+                "bucket": None, 
+                "prefix": path,
+                "sha1": None
+            }
+        raise ValueError(f"Invalid path format: {path}")
+    
+    components = match.groupdict()
+    return components
+
```
devsjc, line 61: [1] Be good to put some tests in the test files that this parses a few of example paths as expected. Also, apologies for being annoying, but I've decided the errors raised in the new match block should actually become part of this parsing function, so they can be tested against more easily too. 

So I'd put a 

```python
if components["protocol"] is None and components["sha1"] is not None:
	raise ValueError("Commit syntax (@commit) not supported for local files")
	
if components["protocol"] == "gs" and components["prefix"] is not None and "*" in components["prefix"]:
    raise ValueError("Wildcard (*) paths are not supported for GCP (gs://) URLs")
```

and remove those checks from the `match` block.

ocf_data_sampler/load/satellite.py
```diff
@@ -1,59 +1,230 @@
...
-        },
+            "distributed.worker.memory.target": 0.7,
+            "array.chunk-size": "512MB",
+            "distributed.comm.compression": None,
+            "distributed.worker.threads": OPTIMAL_THREADS,
+        }
     )
+    logger.debug("Optimization environment configured successfully")
+
+def _parse_icechunk_path(path: str) -> dict:
```
devsjc, line 43: Since this is just parsing the zarr path now, the function name should change to reflect it isn't icechunk-specific.

ocf_data_sampler/load/satellite.py
```diff
@@ -1,59 +1,230 @@
...
+    
+    # Sol's requested match/case pattern for path routing
+    match path_info:
+        case {"protocol": None, "bucket": _, "prefix": _, "sha1": sha1} if sha1 is not None:
+            # Raise error if trying to use commit syntax for local file
+            raise ValueError("Commit syntax (@commit) not supported for local files")
+        
+        case {"protocol": "gs", "bucket": _, "prefix": prefix, "sha1": _} if "*" in prefix:
+            # Raise an error if a wildcard is used in a GCP path
+            raise ValueError("Wildcard (*) paths are not supported for GCP (gs://) URLs")
```
devsjc, line 148: These can be removed as per [1] (sorry!)

ocf_data_sampler/load/satellite.py
```diff
@@ -1,59 +1,230 @@
...
+        
+        case {"protocol": protocol, "bucket": bucket, "prefix": prefix, "sha1": sha1} if ".icechunk" in prefix and protocol is not None:
+            # Ice Chunk logic goes here - Sol's requested signature
+            ds = _open_sat_data_icechunk(protocol, bucket, prefix, sha1)
+        
+        case {"protocol": _, "bucket": _, "prefix": _, "sha1": None}:
+            # Existing single zarr logic path here - Sol's specification
+            ds = xr.open_dataset(zarr_path, engine="zarr", chunks="auto")
```
devsjc, line 156: Now we've rebased, it looks like this should be `ds = open_zarr(zarr_path)` with the open_zarr imported from `xarray_tensorstore` library

ocf_data_sampler/load/satellite.py
```diff
@@ -1,59 +1,230 @@
...
+            # Handle multi-zarr dataset for local files
+            ds = xr.open_mfdataset(
+                zarr_path,
+                engine="zarr",
+                concat_dim="time",
+                combine="nested",
+                chunks="auto",
+                join="override",
+            )
+            check_time_unique_increasing(ds.time)
```
devsjc, line 168: This can go because you take care of multiple zarrs in the `open_sat_data` function

devsjc (changes requested):

ocf_data_sampler/load/satellite.py
Dakshbir: The satellite zarr data which i am working on uses the custom codec ocf-blosc2 and hence it fails with ds = open_zarr(zarr_path).....
But since you mentioned that you guys aren't using the codec now so I will update the code to what you wanted....
So, hopefully it should work fine with the satellite data which isn't compressed using the custom codec

ocf_data_sampler/load/satellite.py
```diff
@@ -9,51 +16,203 @@
...
+        ds_list = []
+        for path in zarr_path:
+            ds_list.append(get_single_sat_data(path))
+        
+        ds = xr.combine_nested(
+            ds_list,
+            concat_dim="time",
+            combine_attrs="override",
+            join="override",
+        )
```
devsjc, line 182: Because icechunks are single folders for the full store, I think we can assume that we won't ever want to open more than one at once for a single data source - and as such this conditional branch can return to the original one, with `open_zarrs` imported from `.open_tensorstore_zarrs`:

```
if isinstance(zarr_path, list | tuple):
        ds = open_zarrs(zarr_path, concat_dim="time")
```

ocf_data_sampler/load/satellite.py
```diff
@@ -9,51 +16,203 @@
...
+            # Ice Chunk logic goes here - Sol's requested signature
+            ds = _open_sat_data_icechunk(protocol, bucket, prefix, sha1)
+        
+        case {"protocol": _, "bucket": _, "prefix": _, "sha1": None}:
+            # Existing single zarr logic path here - Sol's specification
+            ds = open_zarr(zarr_path)
+        
+        case _:
+            # Raise error on unhandled path
+            raise ValueError(f"Unhandled path format: {zarr_path}")
```
devsjc, line 161: I'd just move this logic into the `else` of the `open_sat_data` function directly and remove this extra function altogether, it doesn't do anything that warrants testing on its own.

ocf_data_sampler/load/satellite.py
```diff
@@ -9,51 +16,203 @@
...
+    
+    return ds
+
+def open_sat_data(zarr_path: str | list[str], channels: Optional[List[str]] = None) -> xr.DataArray:
```
devsjc, line 165: Also, just for easier diffing, can you return this function to be the first one declared in the file? The new ones can follow on after.

devsjc (changes requested):

ocf_data_sampler/load/satellite.py
Dakshbir: Okay sure will put a comment in that case

ocf_data_sampler/config/model.py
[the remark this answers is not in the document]
Dakshbir: Yep got it...so i will rollback it to the original file

devsjc (changes requested):
Just one more code suggestion for the main logic - thanks for all the changes!

Can you also move ~the benchmarks into the `scripts` folder, and~ the tests into the `tests` folder?

test_loading.py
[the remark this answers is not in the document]
Dakshbir: yep...its not relevant now....will remove this

ocf_data_sampler/load/satellite.py
```diff
@@ -1,59 +1,198 @@
...
-    # Validate data types directly loading function
-    if not np.issubdtype(data_array.dtype, np.number):
-        raise TypeError(f"Satellite data should be numeric, not {data_array.dtype}")
+def _open_sat_data_icechunk(
+    protocol: str | None, bucket: str, prefix: str, sha1: str | None
+) -> xr.Dataset:
+    """Open satellite data from an Ice Chunk repository with optimized settings."""
+    
+    # Handle local icechunk paths
+    if protocol == "local":
```
devsjc, line 134: Change this to `if protocol == None`

ocf_data_sampler/load/satellite.py
```diff
@@ -1,59 +1,198 @@
...
+        # Remove @commit from path if present for repository opening
+        base_path = prefix.split('@')[0] if '@' in prefix else prefix
+        logger.info(f"Opening local Ice Chunk repository: {base_path}")
+        
+        storage = icechunk.local_filesystem_storage(base_path)
```
devsjc, line 139: Also, you shouldn't need this split, as the commit is already captured seperately to the path in the `sha1` variable. So it should be as straightforward as 

```python
storage = icechunk.local_filesystem_storage(prefix)
```

ocf_data_sampler/load/satellite.py
```diff
@@ -1,59 +1,198 @@
...
+        try:
+            repo = icechunk.Repository.open(storage)
+        except Exception as e:
+            logger.error(f"Failed to open local icechunk repository at {base_path}")
+            raise e
```
devsjc, line 144: You also might as well do the try/catch storage open only once at the end of the storage logic branches, so the logic skeleton for this function ends up looking something like this:

```python
# Get storage according to protocol
if protocol == None:
    storage = icechunk.local_filesystem_storage(prefix)
elif protocol == "gs":
    _setup_optimal_environment() # I think this only applies for GCS right?
    storage = icechunk.gcs_storage(bucket=bucket...)
else:
    raise ValueError(f"Unsupported protocol: {protocol}")
 
 # Get repo from storage
 try:
     repo = icechunk.Repository.open(storage)
 except Exception as e:
     ...
     
 # Get session from repo according to commit, if present
 if sha1 is not None:
     session = # get session from sha1 commit id
 else:
     session = # get session from main branch
     
 # Get ds from session
 ...
 ```

File: ocf_data_sampler/load/satellite.py [hunk 18]
```diff
@@ -1,59 +1,195 @@
...
+        path_info = _parse_zarr_path(zarr_path)
+        
+        # Sol's requested match/case pattern for path routing
+        match path_info:
+            # Updated case to handle local icechunk paths correctly
+            case {"protocol": None, "bucket": bucket, "prefix": prefix, "sha1": sha1} if ".icechunk" in prefix:
+                # Pass the original zarr_path for local icechunk paths
+                ds = _open_sat_data_icechunk("local", "", zarr_path, sha1)
+
+            case {"protocol": protocol, "bucket": bucket, "prefix": prefix, "sha1": sha1} if ".icechunk" in prefix and protocol is not None:
```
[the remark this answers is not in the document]
devsjc: Ah you're right - but in that case, there's no ned for both:

```python
case {"protocol": _, "bucket": bucket, "prefix": prefix, "sha1": sha1} if ".icechunk" in prefix:
    ...
```

Then you just have to change the logic check for the local opening in the `_open_sat_data_icechunk` function to look for `None` instead of `"local"` (as commented below)

devsjc (changes requested):

devsjc (changes requested):

ocf_data_sampler/load/load_dataset.py
```diff
@@ -1,10 +1,12 @@
 """Loads all data sources."""
 
+import logging
 import xarray as xr
 
 from ocf_data_sampler.config import InputData
 from ocf_data_sampler.load import open_gsp, open_nwp, open_sat_data, open_site
 
+logger = logging.getLogger(__name__)
```
felix-e-h-p, line 9: Not sure if needed - please remove

ocf_data_sampler/load/load_dataset.py
```diff
@@ -25,7 +27,7 @@ def get_dataset_dict(
             zarr_path=input_config.gsp.zarr_path,
             boundaries_version=input_config.gsp.boundaries_version,
             public=input_config.gsp.public,
-        )
+        ).compute()
```
felix-e-h-p, line 30: Should be removed - could cause potential memory related issues

ocf_data_sampler/load/satellite.py
```diff
@@ -1,59 +1,183 @@
...
+    except Exception as e:
+        logger.error(f"Failed to open Ice Chunk repository at {protocol or 'local'}://{bucket or ''}/{prefix}")
+        raise e
 
-    for coord, expected_dtype in coord_dtypes.items():
-        if not np.issubdtype(data_array.coords[coord].dtype, expected_dtype):
-            dtype = data_array.coords[coord].dtype
-            raise TypeError(f"{coord} should be {expected_dtype.__name__}, not {dtype}")
+    # Get session from repo according to commit, if present
+    if sha1:
```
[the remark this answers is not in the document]
felix-e-h-p: Potentially like:

```python
try:
    if sha1:
        session = repo.readonly_session(commit=sha1)
    else:
        session = repo.readonly_session("main")

except Exception as e:
    target = sha1 or "main"
    raise ValueError(f"Failed to open session for '{target}': {e}") from e
```

tests/conftest.py
```diff
@@ -99,6 +99,62 @@ def sat_zarr_path(session_tmp_path):
...
+    
+    # Create icechunk store using the correct function name
+    icechunk_path = session_tmp_path / "bucket" / "test_sat.icechunk"
+    os.makedirs(icechunk_path.parent, exist_ok=True)
+    
+    # Use local_filesystem_storage instead of filesystem_storage
+    storage = icechunk.local_filesystem_storage(str(icechunk_path))
+    repo = icechunk.Repository.create(storage)
+    session = repo.writable_session("main")
+    store = session.store  # Remove the parentheses - it's a property, not a method
```
devsjc, line 151:
```suggestion
```

devsjc (changes requested):
Just a few suggestions removing superfluos comments!

tests/conftest.py
Dakshbir: Should i delete this entire line? or just the comment?

ocf_data_sampler/config/model.py
```diff
@@ -368,4 +368,4 @@ class Configuration(Base):
     """Configuration model for the dataset."""
 
     general: General = General()
-    input_data: InputData = InputData()
+    input_data: InputData = InputData()
```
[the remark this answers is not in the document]
Dakshbir: you mean to say i should change this to:-
input_data: InputData = Field(default_factory=InputData)

??

ocf_data_sampler/load/satellite.py
```diff
@@ -1,59 +1,186 @@
...
 )
 
-from .open_tensorstore_zarrs import open_zarrs
-
+logger = logging.getLogger(__name__)
 
-def open_sat_data(zarr_path: str | list[str]) -> xr.DataArray:
-    """Lazily opens the zarr store and validates data types.
+OPTIMAL_BLOCK_SIZE_MB = 64
+OPTIMAL_THREADS = 2
```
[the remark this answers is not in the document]
devsjc: I see you've moved it, but can you remove it from here as well?

ocf_data_sampler/config/model.py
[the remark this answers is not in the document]
devsjc: No, I think just add a carriage return at the end of the line - or just revert the change using vscode git! I can't see a functional difference and this is the only change in this file!

devsjc, line 371:
```suggestion
    input_data: InputData = InputData()

```

dfulu:
This was very useful to explore, but since this is out of date with the current library and has gone stale I'm going to close it for now

PR #296, opened 2025-08-01
Site dataset numpysample

Sukh-P:
# Pull Request

## Description

This PR includes changes to the Site Dataset to get training to work with streaming samples and also align the site pipeline more with the UK regional one, changes include:

- Changing the format that a sample is returned by the torch Site Dataset, it now returns a numpy sample
- Removing all the code which was previously used to transform a dictionary of xarraydatasets into a single xr Dataset
- Changing how the SiteSample class works to now save as a .pt file instead of a netcdf and changing the loading
- Changing tests to reflect the changes above 

## Checklist:

### DIFF
--- a/ocf_data_sampler/torch_datasets/datasets/__init__.py
+++ b/ocf_data_sampler/torch_datasets/datasets/__init__.py
@@ -1,6 +1,5 @@
 from .pvnet_uk import PVNetUKRegionalDataset, PVNetUKConcurrentDataset
 
 from .site import (
-    convert_netcdf_to_numpy_sample,
     SitesDataset
 )
--- a/ocf_data_sampler/torch_datasets/datasets/site.py
+++ b/ocf_data_sampler/torch_datasets/datasets/site.py
@@ -58,6 +58,98 @@ def get_locations(site_xr: xr.Dataset) -> list[Location]:
 
     return locations
 
+def process_and_combine_datasets(
+    dataset_dict: dict,
+    config: Configuration,
+    t0: pd.Timestamp,
+) -> NumpySample:
+    """Normalise and convert data to numpy arrays.
+
+    Args:
+        dataset_dict: Dictionary of xarray datasets
+        config: Configuration object
+        t0: init-time for sample
+    """
+    numpy_modalities = []
+
+    if "nwp" in dataset_dict:
+        nwp_numpy_modalities = {}
+
+        for nwp_key, da_nwp in dataset_dict["nwp"].items():
+
+            # Standardise and convert to NumpyBatch
+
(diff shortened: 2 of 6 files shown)


ocf_data_sampler/torch_datasets/datasets/__init__.py
```diff
@@ -1,6 +1,5 @@
 from .pvnet_uk import PVNetUKRegionalDataset, PVNetUKConcurrentDataset
 
 from .site import (
-    convert_netcdf_to_numpy_sample,
     SitesDataset
```
dfulu, line 4: I'd probably just move this import onto a single line

ocf_data_sampler/torch_datasets/datasets/site.py
```diff
@@ -58,6 +58,98 @@ def get_locations(site_xr: xr.Dataset) -> list[Location]:
...
+        )
+
+        # add datetime features
+        datetimes = pd.DatetimeIndex(da_sites.time_utc.values)
+        datetime_features = encode_datetimes(datetimes=datetimes)
+
+        numpy_modalities.append(datetime_features)
+
+    # Only add solar position if explicitly configured
+    has_solar_config = (
```
dfulu, line 125: I think this could be simpler. The config always has a `solar_position` attribute, [in the config model](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/config/model.py#L364) it defaults to `None`. So the `hasattr(config.input_data, "solar_position")` could go since it is always true, and in that case we probably don't need to assign this to a variable because `if config.input_data.solar_position is not None:` is clean enough

ocf_data_sampler/torch_datasets/datasets/site.py
```diff
@@ -58,6 +58,98 @@ def get_locations(site_xr: xr.Dataset) -> list[Location]:
...
+        datetimes = pd.date_range(
+            t0 + minutes(solar_config.interval_start_minutes),
+            t0 + minutes(solar_config.interval_end_minutes),
+            freq=minutes(solar_config.time_resolution_minutes),
+        )
+
+
+        # Calculate solar positions and add to modalities
+        numpy_modalities.append(
+            make_sun_position_numpy_sample(datetimes, da_sites.longitude.values,
```
dfulu, line 143: My style preference for this kind thing where it doesn't fit on a single line is:

```
x.append(
    f(
        arg1,
        arg2,
        arg3,
    )
)
```

But I don't feel too strongly

Sukh-P:
Just putting it here as a reminder that this probably warrants a minor version bump when merged in

Issue #299, opened 2025-08-06
Speed-up: Remove overuse of `channel_dict_to_dataarray()`


dfulu:
When creating each sample, we take the normalisation coords from the data config and turn them into a xarray object via the `channel_dict_to_dataarray()` function. This is wasteful because we create the exact same few xarray objects when generating each sample. For efficiency, we should should only create these once. Based on some simple profiling, this could speed up our samper by around 5%

PR #305, opened 2025-08-08
Extract the norm values once and store them

dfulu:
# Pull Request

## Description

Previously we were creating an xarray DataArray of the normalisation coefficients each time we created a sample. This was inefficient. This PR changes the behaviour to create the DataArrays once and store them

Completes #299

## Checklist:


tests/torch_datasets/test_pvnet_uk.py
```diff
@@ -1,72 +1,18 @@
...
 from ocf_data_sampler.config.model import SolarPosition
-from ocf_data_sampler.select.location import Location
 from ocf_data_sampler.torch_datasets.datasets.pvnet_uk import (
     PVNetUKConcurrentDataset,
     PVNetUKRegionalDataset,
     compute,
 )
 
 
-def test_process_and_combine_datasets(pvnet_config_filename):
```
peterdudfield, line 18: how come you have removed a test?

dfulu: Because the method it was testing has been moved to an instance method instead of a static method. It now can't be tested by itself. Plus that method is already run in other tests, so it doesn't really need its own

peterdudfield: soudns good

PR #308, opened 2025-08-11
Add logging

### LINKED ISSUE #244: Add Logging for Dataset Generation in Test Fixtures
Hey, 
In the test fixtures for [conftest.py](https://github.com/openclimatefix/ocf-data-sampler/blob/main/tests/conftest.py) I noticed that we’re generating a number of synthetic datasets (like ds_nwp_ukv) but there aren’t any log messages associated with them. I think adding some basic logging here would really help, especially when we're trying to debug test failures—either locally or in CI.
A small enhancement like this :

```
import logging

def ds_nwp_ukv():
    logger.info("Creating dummy UKV NWP dataset")
    ...
```
In this way logs can give us clear idea about what's happening behind the scene .

Why ?

- Helps pinpoint where a test might be breaking or hanging
- Gives better visibility when multiple datasets are being generated
- Makes debugging in CI pipelines much smoother
- It’s a clean, low-effort change with practical benefits

ram-from-tvl:
Fixes #244


dfulu:
@ram-from-tvl thanks for your work. However I'd like to remind you of our [community guidelines](https://github.com/openclimatefix#github-contributions) which say that contributors should wait to be assigned to issues before working on them. 

We don't intend to incorporate the issue that you'd solved here. So I'm going to close this PR and that issue. 

Thanks for your work, hopefully we'll see you in other issues

PR #309, opened 2025-08-11
Refactor xarray-tensorstore modules

dfulu:
# Pull Request

## Description

This cleans up the xarray tensorstore modules to import what we can from `xarray-tensorstore` and rewrite only what we need. 

TODO:

 - [ ] Add some tests, the previous PR silently broke the open_zarrs function since it wasn't used in the tests

## Checklist:

### DIFF
--- a/tests/load/test_open_xarray_tensorstore.py
+++ b/tests/load/test_open_xarray_tensorstore.py
@@ -8,17 +8,15 @@ from ocf_data_sampler.load.open_xarray_tensorstore import open_zarr, open_zarrs
 @pytest.fixture(scope="module")
 def concatable_nwp_like_data(ds_nwp_ecmwf):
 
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
     "pyaml_env",
     "pyresample",
     "h5netcdf",
-    "tensorstore",
+    "xarray-tensorstore",
```
peterdudfield, line 38: Im guessing we are now importing some bits and overwriting some bits? Trying to keep the code down?

peterdudfield: If so, is it worth pinning this to a version, incase the code gets changed, especialyl if we are import private functions

peterdudfield (approved):
Ive approved, but one comment I think should be addressed

PR #315, opened 2025-08-14
(Speed-up) Assume spatial coords are sorted

dfulu:
# Pull Request

## Description

Doing mins and maxes across the spatial coords is slow. In the load functions we make sure that the spatial coordinates are sorted in increasing order using [this function](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/load/utils.py#L15). We can make use of this to index the right values instead of doing mins and maxes

## Checklist:

### DIFF
--- a/ocf_data_sampler/select/select_spatial_slice.py
+++ b/ocf_data_sampler/select/select_spatial_slice.py
@@ -25,13 +25,13 @@ def _get_pixel_index_location(da: xr.DataArray, location: Location) -> tuple[int
     x, y = location.in_coord_system(target_coords)
 
     # Check that requested point lies within the data
-    if not (da[x_dim].min() < x < da[x_dim].max()):
+    if not (da[x_dim].values[0] < x < da[x_dim].values[-1]):
         raise ValueError(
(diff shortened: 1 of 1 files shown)


peterdudfield (commented):
Should we put a check when the dataset is initialed / loaded to make sure coordiantes are strictly ascending. This means we check it once, but we dont use min and max in every sample

dfulu:
> Should we put a check when the dataset is initialed / loaded to make sure coordiantes are strictly ascending. This means we check it once, but we dont use min and max in every sample

My feeling is that since we run the `make_spatial_coords_increasing()` function on all datasets that's enough. Like it checks the coordinates and flips them if they aren't increasing

peterdudfield:
> > Should we put a check when the dataset is initialed / loaded to make sure coordiantes are strictly ascending. This means we check it once, but we dont use min and max in every sample
> 
> My feeling is that since we run the `make_spatial_coords_increasing()` function on all datasets that's enough. Like it checks the coordinates and flips them if they aren't increasing

Perhaps just double checfk they are in each load function? Then i think its great

PR #319, opened 2025-08-15
(Speed-up) NWP time slicing

dfulu:
# Pull Request

## Description

This speeds up the NWP time slicing. 

The important speed considerations of this are:
- Avoiding performing the diff before all the data has been loaded. Taking the diff is non-lazy and we want to load all the data asynchronously at one location in the code.
- Avoid using `xarray.concat` - this is very slow
- Avoid using `xarray.diff()` - this is also slow. We opt to use `np.diff()` instead.
- Avoid using the 2D (init-time, step) slicing where possible. This is much slower than slicing individual specific steps from a single init time. I also think the vast majority of our use-cases use just a single init-time.
- Avoid assigning new coordinates i.e. the channels are no longer renamed because `ds["channel"]=...` is very slow

I have split out taking the diff of the channel from the time selection function. 

Though there should be no backward compatibility issues, but I think this changes enough that should be a minor bump since this does change a few things

## Checklist:

### DIFF
--- a/ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
+++ b/ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py
@@ -22,12 +22,12 @@ from ocf_data_sampler.select import Location, fill_time_periods
 from ocf_data_sampler.torch_datasets.utils import (
     add_alterate_coordinate_projections,
     config_normalization_values_to_dicts,
+    diff_nwp_data,
     fill_nans_in_arrays,
     find_valid_time_periods,
     merge_dicts,
     slice_datasets_by_space,
     slice_datasets_by_time,
-    diff_nwp_data,
 )
 from ocf_data_sampler.utils import minutes, tensorstore_compute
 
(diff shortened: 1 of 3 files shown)


ocf_data_sampler/torch_datasets/utils/time_slice_for_dataset.py
```diff
@@ -28,15 +28,20 @@ def slice_datasets_by_time(
         for nwp_key, da_nwp in datasets_dict["nwp"].items():
             nwp_config = config.input_data.nwp[nwp_key]
 
+            # Add a buffer if we need to diff some of the channels in time
+            if len(nwp_config.accum_channels)>0:
+                diff_buffer = minutes(nwp_config.interval_end_minutes)
+            else:
+                diff_buffer = minutes(0)
+
```
peterdudfield, line 36: im not sure if `minutes` was a slow call, but you could do
```
if len(nwp_config.accum_channels)>0:
    interval_end = minutes(nwp_config.interval_end_minutes + X)
else:
    interval_end = minutes(nwp_config.interval_end_minutes)
```
and then use interval_end below

peterdudfield: also should the diff_buffer be `minutes(nwp_config.interval_end_minutes)` or should it just be one extra timestamp

dfulu: Yes it should thanks!

peterdudfield: looks good

peterdudfield (approved):

PR #331, opened 2025-08-22
Convert index to int for site getitem

AUdaltsova:
# Pull Request

## Description
This adds casting the index in `_getitem_` as `int` because it is currently supplied as a tensor, which breaks pandas indexing. Judging by [this line](https://github.com/openclimatefix/ocf-data-sampler/blob/21e76fba7f2a64d6224d7a8be6443951d38e5aaa/ocf_data_sampler/torch_datasets/datasets/pvnet_uk.py#L272C9-L272C23) we've had the same issue in the GSP pipeline, so just adding this for site.

## How Has This Been Tested?

- [X] Yes

Running training on prem

## Checklist:

### DIFF
--- a/ocf_data_sampler/torch_datasets/datasets/site.py
+++ b/ocf_data_sampler/torch_datasets/datasets/site.py
@@ -261,9 +261,9 @@ class SitesDataset(PickleCacheMixin, Dataset):
 
     @override
     def __getitem__(self, idx: int) -> dict:
-        
+
         idx = int(idx)
-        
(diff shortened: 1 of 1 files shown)


AUdaltsova:
@peterdudfield yeah as far as I could find, when iterating over a dataset torch issues indices as tensors, not ints, which pandas can't use for indexing. There might be a different way around it, and I'm not sure why it wasn't an issue before, we can investigate further.

PR #333, opened 2025-08-23
Fixes #330 by ensuring site generation data is cast to float on initialization.

### LINKED ISSUE #330: site generattion of dtype int is not allowed
## Describe the bug
I'm using new site data which comes as dtype int and failing the load check for site. Unless there is a reason that we require a float and not an int for site data, this should be extended to allow int data.

Satvik24511:
# Pull Request

## Description

This PR fixes issue #330, where site generation of dtype int was not allowed. This was being caused because of an assert statement which ensured the data values stay between 0 and 1 and are float values.

Fixes #
Made the data convert to float before being initialised in class. Also changed the assert statement to ensure that the values are float and >= 0.

## Checklist:

### DIFF
--- a/ocf_data_sampler/torch_datasets/sample/site.py
+++ b/ocf_data_sampler/torch_datasets/sample/site.py
@@ -1,9 +1,11 @@
 """PVNet Site sample implementation for netCDF data handling and conversion."""
 
 import torch
+import numpy as np
 from typing_extensions import override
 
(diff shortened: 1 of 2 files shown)


AUdaltsova:
Hi @Satvik24511, thanks for jumping in but as per the issue comment, this is not something we're looking to add currently. I'll close this for now.

PR #338, opened 2025-08-28
accommodate coord mismatch for satellite

### LINKED ISSUE #337: Silent error when combining satellite data with slightly different coordinates
## Describe the bug
Our satellite data archive for UK currently has two parts: v1 (2017-2022) and v2 (2023-2024). The data in these two folders has slightly different crops, e.g.:
```
>>> sat_v1.x_geostationary
<xarray.DataArray 'x_geostationary' (x_geostationary: 614)> Size: 5kB
array([   15002.015829,    12001.612663,     9001.209497, ..., -1818244.318485,
       -1821244.721651, -1824245.124817], shape=(614,))
>>> sat_v2.x_geostationary
<xarray.DataArray 'x_geostationary' (x_geostationary: 614)> Size: 5kB
array([   15002.107544,    12001.704468,     9001.301392, ..., -1818244.171997,
       -1821244.575073, -1824244.978149], shape=(614,))
```
NB: there is a difference in the decimal parts for every coordinate value, even though they are very close (the difference is less than 0.15 in geostationary for all points, which is less than 0.005% of the spacial resolution). The crops are also slightly different, though almost imperceptibly so when eyeballing:

Image

Image

Currently, [we use `xr.concat`](https://github.com/openclimatefix/ocf-data-sampler/blob/b70f9efa8c9a733168784e0a179997bb28024d3f/ocf_data_sampler/load/open_xarray_tensorstore.py#L146C5-L152C6) to join datasets consisting of different files. `xr.concat` merges the coordinates by default, so when supplied the data from both v1 and v2 folders ends up creating something like this:

Image

Which is also half the crop you would expect because the amount of coordinate points has doubled. **This happens silently**. The best you will get is a performance warning that  the chunk size is being increased.

Previously, the satellite loader circumvented this [by using `xr.combine_nested(...., join="override")`](https://github.com/openclimatefix/ocf-data-sampler/blame/60652e4bfa6946c3c2dbe243e960147ab97a63c9/ocf_data_sampler/load/satellite.py#L63-L68) which [replaces the indexes of all objects by that of the first object if they are of the same size](https://docs.xarray.dev/en/v2024.10.0/generated/xarray.combine_nested.html). I think this is a good way of solving this specific problem (and the fact that we had that in makes me think that this is a recurring issue and sat data just does this from time to time), but right now we use the same opening method across all data sources, and putting `join="override"` in for everything seems risky to me. 

## Suggested solutions
It seems like we need some way of checking for coordinate match explicitly before concatenation, and at least throw a warning when that's not the case instead of letting `xr.concat` perform an outer join silently. One way to do this would be to put a `xr.concat(..., join="exact")` under a try/except clause, which means that we basically get current behavior if coordinates are aligned, and can get it to through a warning and do a `xr.concat(... join="override")` if not. We can also condition it on only being satellite/within a certain deviation etc if we wanted to, and fail hard in other cases.

AUdaltsova:
# Pull Request

## Description

Some of our satellite data has slightly different coordinates which we want to override when concatenating into a single dataset (see #337). This PR:
- Adds try/except to try concatenating with exact coordinate match, then throw warning and concatenate with coordinate override
- Adds a `data_source` str parameter to open_zarrs for warning context. It might be a bit excessive but I think is very useful for the warning, so it's not just an "Uh oh! Some of your data is not right, don't know what though!"

Fixes #337 

## How Has This Been Tested?
Set off training with the problematic satellite data and normal ECMWF data. Also tried putting in two files from different ECMWF archives to check that it throw the right warning.

- [X] Yes

## Checklist:

### DIFF
--- a/ocf_data_sampler/load/nwp/providers/utils.py
+++ b/ocf_data_sampler/load/nwp/providers/utils.py
@@ -75,7 +75,7 @@ def _tensostore_open_zarr_paths(zarr_path: str | list[str], time_dim: str) -> xr
         zarr_path = sorted(glob(zarr_path))
 
     if isinstance(zarr_path, list | tuple):
-        ds = open_zarrs(zarr_path, concat_dim=time_dim, data_source="NWP").sortby(time_dim)
+        ds = open_zarrs(zarr_path, concat_dim=time_dim, data_source="nwp").sortby(time_dim)
     else:
         ds = open_zarr(zarr_path)
     return ds


ocf_data_sampler/load/nwp/providers/utils.py
```diff
@@ -75,7 +75,7 @@ def _tensostore_open_zarr_paths(zarr_path: str | list[str], time_dim: str) -> xr
         zarr_path = sorted(glob(zarr_path))
 
     if isinstance(zarr_path, list | tuple):
-        ds = open_zarrs(zarr_path, concat_dim=time_dim).sortby(time_dim)
+        ds = open_zarrs(zarr_path, concat_dim=time_dim, data_source="NWP").sortby(time_dim)
```
Sukh-P, line 78:
```suggestion
        ds = open_zarrs(zarr_path, concat_dim=time_dim, data_source="nwp").sortby(time_dim)
```
Can ignore, just thought might be good to keep strings in lowercase since uppercase has an implied meaning of global variable and stuff in python

Sukh-P (approved):
Nice work adding in this warning in and changing the default behaviour!

PR #339, opened 2025-08-29
Change site loader

Sukh-P:
# Pull Request

## Description

Three changes are including in this PR:

- The main change is the relaxation of one of the checks to allow `NaN` values in the site generation data, this is due to the fact if you create a xarray dataset of generation data from different sites which have some non-overlapping time periods of data availability the resulting generation data variable will have `NaN` values which is fine as this is handled in our sampling logic [here](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/torch_datasets/datasets/site.py#L228-L231) which handles in each site separately and then combines all available samples
- The next biggest change is following suit from a change in the gsp loader to load the site generation data into memory rather than have this still be lazy loaded, this speeds up data sampling and comes at a small extra memory cost as the site data is generally small. 
- Lastly there is a small change to allow either int or float for capacity values for site metadata as either would work 

## How has this been tested?
- Created a site generation dataset with non overlapping time periods (so with `NaN` values in the data variable) between sites and created samples with these changes and updated unit tests for other change. 

## Checklist:

### DIFF
--- a/ocf_data_sampler/load/site.py
+++ b/ocf_data_sampler/load/site.py
@@ -31,8 +31,8 @@ def open_site(generation_file_path: str, metadata_file_path: str) -> xr.DataArra
         capacity_kwp=("site_id", metadata_df["capacity_kwp"].values),
     )
 
-    # Sanity checks
-    # Allow NaNs in generation_kw as can have non overlapping time periods for sites
+    # Sanity checks, to prevent inf or negative values
+    # Note NaNs are allowed in generation_kw as can have non overlapping time periods for sites
     if np.isinf(generation_ds.generation_kw.values).all():
         raise ValueError("generation_kw contains infinite (+/- inf) values")
     if not (generation_ds.capacity_kwp.values > 0).all():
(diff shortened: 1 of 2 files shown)


peterdudfield (approved):
One small comment, but otherwise this looks great

ocf_data_sampler/load/site.py
```diff
@@ -31,8 +31,8 @@ def open_site(generation_file_path: str, metadata_file_path: str) -> xr.DataArra
         capacity_kwp=("site_id", metadata_df["capacity_kwp"].values),
     )
 
-    # Sanity checks
-    # Allow NaNs in generation_kw as can have non overlapping time periods for sites
+    # Sanity checks, to prevent inf or negative values
```
peterdudfield, line 34: does it prevent negative values? or does it just prevent inf or negative inf values

Sorry to be very critical on this

Sukh-P: The comment is for both checks, the second check prevents negative values

peterdudfield: ah i see

PR #341, opened 2025-09-05
site loader update for variable capacity

### LINKED ISSUE #295: Variable site capacity
### This issue will address two new improvements:
1. Changing the data loading pipeline for sites to expect a single file that contains all generation, capacity and corresponding metadata about a site. This differs from the current implementation which would expect a generation file and metadata file. The reason for this, is to reduce the number of files needed when creating site data to train, and to simplify/reduce the code needed to handle cases where there is a variable capacity vs time.
2. Handling variable capacity. I created an initial solution to this in this PR #239, but to account for some of the feedback to merge the generation and metadata file and to reduce the downstream code, it may be better to work off the main branch rather than this existing PR. (also as a few changes have come along since i started the PR).

## Desired implementation:
When training is kicked off for sites, validation is done on the combined metadata/generation file to check it is in the expected format. It could be best for the `open_site` function (https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/load/site.py) to expect a capacity value for every generation value in the dataset. (even if the capacity stays the same across time). This should mean we can keep the code a bit cleaner than my implementation in #239. For example we would no longer need to add a parameter to the config and use the parameter downstream in https://github.com/openclimatefix/ocf-data-sampler/pull/239/files#diff-b40689b299c299259884e6f50e9df2f3460799a45f3de4496aa76d2d4b39f515 as i did in my implementation.
The `open_site` function should return the whole `generation_ds` dataset, not just extract the generation as it currently does https://github.com/openclimatefix/ocf-data-sampler/commit/cf769585c83ae4d9874666d14006b90e1c49d24e#diff-c5d31e277f06a22594c11e50b6a1e61191deab35e6926c00be073ac69431101d, this commit introduce returning the it as `site_da = generation_ds.generation_kw`.

The sites data then gets processed downstream in the `_get_sampler` function here: https://github.com/openclimatefix/ocf-data-sampler/blob/e87ec372e311b60891488ed06273271a0b871391/ocf_data_sampler/torch_datasets/datasets/site.py#L174. Important to ensure that the correct capacity data is preserved here as I came across some interesting behaviour previously in how xarray will drop coordinates and dimensions which are not apart of the data variable of the dataset when doing this processing. (having a capacity which is data variable which is extracted to a coordainte at this stage could be a possible solution for this). 

## Tests:
It would be useful to test the following use cases:
- Several sites with and without variable capacity
- 1 site with and without variable capacity

## Some things to note a keep in mind:
- Moving from a dataarray to a dataset as we have two variables in our dataset (generation and capacity).
- Backwards compatibility? 
- It might be cleaner to handle some of the data validation outside of the function for cleaner code.
- A fix for solar cords has been implimented in https://github.com/openclimatefix/ocf-data-sampler/commit/f593355070be94df5f2fb2c8c99d43e573c13b4d, which I was working on in #239. 
- Sample streaming

### Other potentially relevant issues
-  #178 
-  #46

AUdaltsova:
# Pull Request

## Description

Small change to the site loader to get the capacity from the generation .nc file (in line with the GSP loader), which makes variable capacity possible. The older site files contain static capacity in the metadata .csv, so added an if/else clause that checks whether capacity is provided in the .nc and loads it from the .csv if not.

For testing:
- Made `create_site_data`'s parametrisation in conftest accessible (to be able to create different files for site data with and without variable capacity)
- Updated site loading test and added variable capacity test to explicitly check that the capacity from the expected file is used in both cases.

Fixes #295 

## How Has This Been Tested?

Created some samples with this change and checked manually that normalisation has been performed correctly

- [X] Yes

## Checklist:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -271,7 +271,10 @@ def create_site_data(
     if variable_capacity:
         ds = xr.Dataset({
             "capacity_kwp": xr.DataArray(
-                np.tile(rng.uniform(1,100,1)*base["capacity_kwp"], (len(times), 1)).astype(np.float32), coords=coords,
+                np.tile(
+                    rng.uniform(1,100,1)*base["capacity_kwp"],
+                    (len(times), 1)).astype(np.float32),
+                coords=coords,
             ),
             "generation_kw": xr.DataArray(
                 rng.uniform(0, 200, (len(times), num_sites)).astype(np.float32), coords=coords,


tests/load/test_load_sites.py
```diff
@@ -10,11 +10,29 @@
...
+    site_model = default_data_site_model(variable_capacity=True)
+    da = open_site(site_model.file_path, site_model.metadata_file_path)
+
+    assert isinstance(da, xr.DataArray)
+    assert da.dims == ("time_utc", "site_id")
+    assert "capacity_kwp" in da.coords
+    assert (da.capacity_kwp.data == 2).all()
     assert "latitude" in da.coords
     assert "longitude" in da.coords
     assert da.shape == (49, 10)
```
peterdudfield, line 38: is it possible to make a test with variable capacity, and check that not all `da.capacity_kwp.data` are the same

ocf_data_sampler/load/site.py
```diff
@@ -25,11 +25,20 @@ def open_site(generation_file_path: str, metadata_file_path: str) -> xr.DataArra
...
-        longitude=("site_id", metadata_df["longitude"].values),
-        capacity_kwp=("site_id", metadata_df["capacity_kwp"].values),
-    )
+    # Check if variable capacity was passed with the generation data
+    # If not assign static capacity from metadata
+    if hasattr(generation_ds,"capacity_kwp"):
+        generation_ds = generation_ds.assign_coords(
+            latitude=(metadata_df.latitude.to_xarray()),
+            longitude=(metadata_df.longitude.to_xarray()),
+            capacity_kwp=generation_ds.capacity_kwp,
```
Sukh-P, line 34: What does the genenration_ds look like when capacity_kwp is in it? Are there already capacity_kwp coordaintes? If so is line 34 here redundant?

tests/conftest.py
```diff
@@ -336,74 +336,80 @@ def ds_uk_gsp():
...
-        interval_end_minutes=site_interval_end_minutes,
-        time_resolution_minutes=site_time_resolution_minutes,
-    )
-    return site_model
-
-
 @pytest.fixture(scope="session")
-def data_sites(session_tmp_path):
-    return create_site_data(tmp_path_base=session_tmp_path)
+def default_data_site_model(session_tmp_path):
```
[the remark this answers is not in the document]
peterdudfield: You can pull that function pack outside (with the if and else), then you can make two fixtures
1.  as before
```
def data_sites(session_tmp_path):
    return create_site_data(tmp_path_base=session_tmp_path)
```
2.
```
def data_sites_variable_capacity(session_tmp_path):
    return create_site_data(tmp_path_base=session_tmp_path, variable_capacity=True)
```
or something like this

ocf_data_sampler/load/site.py
Sukh-P: Sounds good, thanks

tests/load/test_load_sites.py
```diff
@@ -10,11 +10,31 @@
...
+
+    assert len(np.unique(da.coords["site_id"])) == da.shape[1]
+
+
+def test_open_site_variable_capacity(default_data_site_model_variable_capacity):
+    """Test the site data loader with valid data."""
+    site_model = default_data_site_model_variable_capacity
+    da = open_site(site_model.file_path, site_model.metadata_file_path)
+    generation_ds = xr.open_dataset(site_model.file_path)
+    capacity = generation_ds.capacity_kwp
```
AUdaltsova, line 32:
```suggestion
```

tests/load/test_load_sites.py
```diff
@@ -10,11 +10,31 @@
...
+
+    assert isinstance(da, xr.DataArray)
+    assert da.dims == ("time_utc", "site_id")
+    assert "capacity_kwp" in da.coords
+    assert (da.capacity_kwp.data == generation_ds['capacity_kwp']).all()
```
AUdaltsova, line 37:
```suggestion
    assert (da.capacity_kwp.data == generation_ds["capacity_kwp"]).all()
```

Issue #327, opened 2025-08-21
Add t0 to batch


peterdudfield:
Add t0 to batch into both site and GSP batch

This helps for debugging. Its a small amonut of data, but does add a small amount

PR #342, opened 2025-09-08
add t0 to sample

peterdudfield:
# Pull Request

## Description

- add t0 into batch on site and pvnet datasets
- added a dataloader test

#327 

## How Has This Been Tested?

- [ ] Ci tests
- [x] added to tests

## Checklist:

### DIFF
--- a/tests/torch_datasets/test_pvnet_uk.py
+++ b/tests/torch_datasets/test_pvnet_uk.py
@@ -280,10 +280,11 @@ def test_pvnet_uk_regional_dataset_batch_size_2(pvnet_config_filename):
 
     # Assertions for the raw batch
     assert isinstance(batch, dict), \
-        "Sample yielded by DataLoader with batch_size=None should be a dict"
+        "Sample yielded by DataLoader with batch_size=2 should be a dict"
 
     # Check for expected keys directly
-    required_keys = ["nwp", "satellite_actual", "gsp", "solar_azimuth", "solar_elevation", "gsp_id"]
+    required_keys = \
+        ["nwp", "satellite_actual", "gsp", "solar_azimuth", "solar_elevation", "gsp_id", "t0"]
     for key in required_keys:
         assert key in batch, f"Raw Sample: Expected key '{key}' not found"
 


peterdudfield:
> Could even be worth adding these in as part of our tests, would be an integration tests of sort for compatibility of the samples with PVNet once all the batch/tensor operations are done, although we don't add things to the samples that often so could be slight overkill

its easy to add to `stack_np_samples_into_batch`, but its a bit harder to add to `batch_to_tensor` and `copy_batch_to_device`. Do we have a test for `DataLoader` with a non-zero batch size, this might be a way to test this things. Perhaps this could over to a new issue? What do you think @dfulu and @Sukh-P

peterdudfield:
> > Could even be worth adding these in as part of our tests, would be an integration tests of sort for compatibility of the samples with PVNet once all the batch/tensor operations are done, although we don't add things to the samples that often so could be slight overkill
> 
> its easy to add to `stack_np_samples_into_batch`, but its a bit harder to add to `batch_to_tensor` and `copy_batch_to_device`. Do we have a test for `DataLoader` with a non-zero batch size, this might be a way to test this things. Perhaps this could over to a new issue? What do you think @dfulu and @Sukh-P

Actually fidderling a bit more, I think it might be worthing putting this in here. Have to be a bit careful with types and this t0 is a datetime, so I think I want to make sure

tests/torch_datasets/test_pvnet_uk.py
```diff
@@ -257,3 +259,47 @@ def test_pvnet_uk_regional_dataset_pickle(tmp_path, pvnet_config_filename):
...
+    batch = next(iter(dataloader))
+    batch = batch_to_tensor(batch)
+    batch = copy_batch_to_device(batch, torch.device("cpu"))
+
+    # Assertions for the raw batch
+    assert isinstance(batch, dict), \
+        "Sample yielded by DataLoader with batch_size=None should be a dict"
+
+    # Check for expected keys directly
+    required_keys = ["nwp", "satellite_actual", "gsp", "solar_azimuth", "solar_elevation", "gsp_id"]
```
Sukh-P, line 286: Would you want t0 in this check as well?

Sukh-P:
Looks good, thanks for adding in the additional test checking all the batch & tensor operations still work!

PR #350, opened 2025-09-18
Testing neater

felix-e-h-p:
Neatening tests - subdirectory per commit for clarity (or thereabouts)

Current focus of removing redundancies, repetitions, reduction of line count  etc

Feel conftest should be segregated and updated after this per testing directory etc - issue here: https://github.com/openclimatefix/ocf-data-sampler/issues/353

### DIFF
--- a/.github/workflows/pull_ci.yml
+++ b/.github/workflows/pull_ci.yml
@@ -16,3 +16,4 @@ jobs:
       containerfile: 'None'
       tests_folder: 'tests'
       tests_matrix: true
+      test_python_versions: '["3.11", "3.12"]'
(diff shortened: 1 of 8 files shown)


tests/config/test_config.py
```diff
@@ -17,125 +27,114 @@ def test_extra_field_error(test_config_gsp_path):
...
     ):
-        _ = Configuration(**configuration.model_dump())
+        _revalidate(configuration)
 
 
 def test_accum_channels_validation(test_config_filename):
     """Test accum_channels validation with required normalization constants."""
-    # Load valid config (implicitly tests valid case)
     config = load_yaml_configuration(test_config_filename)
     nwp_name, _ = next(iter(config.input_data.nwp.root.items()))
```
Sukh-P, line 139: Could use the new `_load_config_and_provider` function here

Issue #334, opened 2025-08-26
Join GSP and Site pipelines together


peterdudfield:
## Detailed Description
It would be great to join the Site and GSP site pipelines together

## Context
- good to reduce code
- GSP have variable capacity #295 

## Possible Implementation
- Need to reformat and define data that goes in site pipeline
- More details to come ....
- Tidy up confgiruation

PR #365, opened 2025-10-13
One pipeline with different data input formats

Sukh-P:
# Pull Request

## Description

Changes to combine the torch datasets for gsps and sites without changing data formats (1st step) 

Starts on solving https://github.com/openclimatefix/ocf-data-sampler/issues/334, more notes on further improvements [here](https://docs.google.com/document/d/14hVrDzRGGkCIoQ6LcjCJnMY7GQp3fchWtIPyg7Zk4rQ/edit?tab=t.0)

This is to merged into a development branch so we can get changes/reviews in smaller increments, current changes still supports gsp data or site data separately and changes logic based on whether site or gsp data is being supplied in the data config, more simplification to come from using one data schema in the next changes

### DIFF
--- a/ocf_data_sampler/torch_datasets/pvnet_dataset.py
+++ b/ocf_data_sampler/torch_datasets/pvnet_dataset.py
@@ -1,0 +1,522 @@
+"""Torch dataset for UK PVNet."""
+
+import os
+import pickle
+
+import pandas as pd
+import xarray as xr
+from torch.utils.data import Dataset
+from typing_extensions import override
+
+from ocf_data_sampler.config import Configuration, load_yaml_configuration
+from ocf_data_sampler.load.gsp import get_gsp_boundaries
+from ocf_data_sampler.load.load_dataset import get_dataset_dict
+from ocf_data_sampler.numpy_sample import (
+    convert_gsp_to_numpy_sample,
+    convert_nwp_to_numpy_sample,
+    convert_satellite_to_numpy_sample,
+    convert_site_to_numpy_sample,
+    encode_datetimes,
+    make_sun_position_numpy_sample,
+)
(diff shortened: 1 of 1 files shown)


ocf_data_sampler/torch_datasets/datasets/pvnet_dataset.py
```diff
@@ -175,17 +219,30 @@ def process_and_combine_datasets(
...
-        numpy_modalities.append(
-            {
-                GSPSampleKey.gsp_id: location.id,
-                GSPSampleKey.x_osgb: osgb_x,
-                GSPSampleKey.y_osgb: osgb_y,
-            },
-        )
+            numpy_modalities.append(
+                {
+                    GSPSampleKey.gsp_id: location.id,
```
Sukh-P: Will no longer has GSP/Site SampleKeys and would just have one so this will be simpler

ocf_data_sampler/torch_datasets/datasets/pvnet_dataset.py
```diff
@@ -272,43 +395,54 @@ def __getitem__(self, idx: int) -> NumpySample:
...
+class PVNetConcurrentDataset(AbstractPVNetDataset):
+    """A torch Dataset for creating concurrent PVNet location samples."""
 
     @override
     def __len__(self) -> int:
-        return len(self.valid_t0_times)
+        if self.config.input_data.gsp:
+            return len(self.valid_t0_times)
+        # site specific as t0s repeated for all sites
+        return len(self.valid_t0_times)//len(self.locations)
```
dfulu: This assumes that there are absolutely no NaNs in the site data right? Otherwise I think some t0 times would be filtered out here. For example, if one of the sites can't be run for even one init-time then the last init-time would be cut by this

PR #381, opened 2025-12-08
fixed type hints for coord_system parameter in the location class

prashantpandeygit:
# Pull Request

## Description
`coord_system` in `Location.__init__` and `Location.add_coord_system` was typed as `int`, but the codebase uses string. also,

while checking the tests for the repository i found the tests/select/test_location.py file uses coord_system='osgb' which is a string, whereas in the select/location.py the coord_system is typed as int in two functions, i think might cause issues. 

## How Has This Been Tested?
ran pytest again through the repo again, all the checks passed without any breaking

### DIFF
--- a/ocf_data_sampler/select/location.py
+++ b/ocf_data_sampler/select/location.py
@@ -7,7 +7,7 @@ allowed_coord_systems = {"osgb", "lon_lat", "geostationary"}
 class Location:
     """A spatial location."""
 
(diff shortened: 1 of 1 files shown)


prashantpandeygit:
hi @Sukh-P can you have a review at this? thanks!

prashantpandeygit:
pushed to PR https://github.com/openclimatefix/ocf-data-sampler/pull/382

PR #385, opened 2025-12-23
Add optional t0 embedding features

dfulu:
# Pull Request

## Context

Previously, I did an experiment which showed that giving the model a feature encoding whether t0 is HH:00 or HH:30 increased the model accuracy. The experiment was motivated by the fact the the NWP data is hourly and the model may want to learn to use the NWP differently at HH:00 and HH:00 t0 times. The experiment write up is [here](https://docs.google.com/document/d/1st4g9tphsqWOpCVuLBzT_HMFJCm0z3MrBJKBFsPA3o0/edit?usp=sharing)

## PR description

This PR adds the option to include embeddings of the t0 time. I've tried to generalised the functionality so that we can choose the periodicity of the embedding features. i.e. we can encode how far through each hour t0 is. For 3-hourly NWP steps we can also generate hour far through each 3-hour period t0 is. We can also use it to make time-of-day (by adding "24h" in `periods`) and day-of-year features (by adding "1y" in `periods`).

I've made it configurable so that we can choose a representation for each periodicity. For example, for time-of-day we likely want to sin-cos embed the times ("cyclic" in the code) since 11:59pm is close to 00:00am. But for time-past-the-hour features we likely likely want linear encodings where `HH:00 -> 0` and `HH:59 -> ~1`. This time-past-the-hour feature is supposed to inform the network about how to interpolate the NWP data.We shift the NWP inputs by a full hour between HH:59 (HH+1):00. So in this way HH:00 is maximally dissimilar HH:59 - so we would want our time feature to reflect this.

## Extra notes

I did a couple of extra semi-related clean-ups in this PR that I couldn't resist. I've noted them

## Future usage

I think it's likely that this function could replace the usage of `encode_datetimes()` entirely. Our PVNet models are trained to predict for fixed set of horizons, so if we already have the time-of-day and day-of-year features of t0, there is no additional information in knowing the other forecast datetimes. Once we (or the model) know t0 we already know all forecast datetimes.

I think we should try to move to sunset the `encode_datetimes()` function but I've left it in for now. We'll need to do a little testing and I don't want to break backwards compatibility.

## How Has This Been Tested?

I've added tests for the exact numeric values which `get_t0_embedding()` produced and checked that it works inside the `PVNetDataset` class.

## Checklist:

### DIFF
--- a/ocf_data_sampler/numpy_sample/datetime_features.py
+++ b/ocf_data_sampler/numpy_sample/datetime_features.py
@@ -62,7 +62,7 @@ def get_t0_embedding(
         if embedding=="cyclic":
             radians = 2 * np.pi * frac
             features.extend([np.sin(radians), np.cos(radians)])
-        
+
         elif embedding=="linear":
             features.append(frac)
 


ocf_data_sampler/torch_datasets/pvnet_dataset.py
```diff
@@ -243,30 +244,37 @@ def process_and_combine_datasets(
...
 
-            numpy_modalities.append(datetime_features)
+        if self.config.input_data.t0_embedding is not None:
+            emb_conf = self.config.input_data.t0_embedding
+            numpy_modalities.append(get_t0_embedding(t0, emb_conf.periods, emb_conf.embeddings))
 
-            # Only add solar position if explicitly configured
-            if self.config.input_data.solar_position is not None:
-                solar_config = self.config.input_data.solar_position
+        # Only add solar position if explicitly configured
```
dfulu: This is another unrelated clean up I spotted along the way. Adding the solar features should not depend on the generation data being there

ocf_data_sampler/numpy_sample/datetime_features.py
```diff
@@ -27,3 +29,52 @@ def encode_datetimes(datetimes: pd.DatetimeIndex) -> NumpySample:
...
+
+    if len(periods)!=len(embeddings):
+        raise ValueError("`periods` and `embeddings` must be the same length")
+
+    for period_str, embedding in zip(periods, embeddings, strict=True):
+
+        if period_str.endswith("h"):
+            period_hours = int(period_str.removesuffix("h"))
+            if not (1<=period_hours<=24):
+                raise ValueError("The period in hours must be in interval [1, 24]")
```
Sukh-P: Similar comment here to the one above about this validation check already seems covered by the pydantic validation so not sure it's adding much here

ocf_data_sampler/numpy_sample/datetime_features.py
```diff
@@ -27,3 +29,52 @@ def encode_datetimes(datetimes: pd.DatetimeIndex) -> NumpySample:
...
+        periods: List of periods to encode (e.g., "1h", "Nh", "1y", "Ny")
+        embeddings: How to represent each of these periods. Either "cyclic" or "linear". When cyclic
+            the period is sin-cos embedded, else it is 0-1 scaled as fraction through the period.
+            Note that using "cyclic" adds 2 elements to the output vector to embed a period whilst
+            "linear" adds only 1 element.
+    """
+    features = []
+
+    if len(periods)!=len(embeddings):
+        raise ValueError("`periods` and `embeddings` must be the same length")
```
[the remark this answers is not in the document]
dfulu: I did think about that, and I'd be more than happy if the code was simplified here. I'll remove them here

PR #397, opened 2026-02-24
Speedup: Add optional LightDataArray backend and migrate datetimes to np.datetime64

dfulu:
# Pull Request

## Description

There are two main changes in this PR which combine the speed up the data-sampler.

1. Removing/reducing the use of pandas DatetimeIndex class and migrating to use numpy datetime64 everywhere
2. Replacing xarray DataArray with a custom class which mimics a limited portion of xr.DataArray's functionality and so is faster

These changes were motivated by running code profiling with `pyinstrument`

## 1. Getting rid of pandas timestamps

When profiling the code I found that we were spending a silly amount of the time doing things with timestamps. This is just because the pandas `Timestamp`, `DatetimeIndex` and `date_range()` are slow. I have moved us to use numpy datetime64 objects everywhere to handle datetimes. The pandas timestamp stuff is built on top of numpy datetime64 anyway so we've just removed a layer of overhead. Unfortunately this costs us in simplicity of code. The pandas Timestamps have those nice `.hour`, `.day_of_year`, `.ceil()` attributes and methods which numpy does not have. I have added a new file `time_utils.py` with a bunch of helper functions to make working with the numpy datetime64 objects easier and to replace some of the utility we were getting from pandas.

I've added tests for these utility functions which check them against the pandas equivalents.

## 2. Replacing xarray DataArray 

This is the biggest part of the speedup. In profiling we were spending a lot of time on xarray's internal methods which we don't really need. I've added a new class `LightDataArray` which replaces the xarray `DataArray` in the code path under `_get_sample()`. This means that we can use xarray for the initialisation of the torch Dataset (where the speed of xarray doesn't matter) but then we use the new class when we are slicing out samples. We do not need all of xarray's functionality when slicing these samples, and so we can do it faster with the custom class. 

I have based the custom class on the xarray DataArray class. For example I've created `isel()`, `.sel()`, and `.load()` methods and it has a `.values` property. This means we can get away with less code changes here and also means it should be pretty trivial if we want to switch back to xarray. In the current code I have left it as a option to use the new `LightDataArray` class when calling `_get_sample()` and it works pulling from `xr.DataArray`s or `LightDataArray`s. This has the benefit of making it easy to test. 

The `LightDataArray` object has a `from_xarray()` method which instantiates a `LightDataArray` from an `xr.DataArray`. The new class also has a `to_xarray()` method which instantiates an `xr.DataArray` from a `LightDataArray` instance. This makes it easy to test that the methods of `LightDataArray` match the xarray methods. For example, we can run something like

```python
# Create a LightDataArray from the xarray DataArray
lda = LightDataArray.from_xarray(da)

# Run some xarray method
da_Result = da.sel(...)

# Run the same method on the LightArray
lda_result = lda.sel(...)

# Convert the LDA back to xarray and check the results are the same
lda_result.to_xarray()).equals(da_result)
```

So we use the xarray behaviour as the expected behaviour for our tests

## Checklist:

### DIFF
--- a/ocf_data_sampler/torch_datasets/pvnet_dataset.py
+++ b/ocf_data_sampler/torch_datasets/pvnet_dataset.py
@@ -190,9 +190,11 @@ class AbstractPVNetDataset(PickleCacheMixin, Dataset):
 
         # Assign config and input data to self
         self.config = config
-        self.datasets_dict = datasets_dict
-        self.use_xarray = use_xarray
-        self.light_datasets_dict = xarray_to_lightarray_dict(datasets_dict)
+
+        if use_xarray:
+            self.datasets_dict = datasets_dict
+        else:
+            self.datasets_dict = xarray_to_lightarray_dict(datasets_dict)
 
         # Assign t0 idx value
         self.t0_idx = (
@@ -406,9 +408,7 @@ class PVNetDataset(AbstractPVNetDataset):
             t0: init-time for sample
             location: location for sample
         """
-        input_dict = self.datasets_dict if self.use_xarray else self.light_datasets_dict
-
-        sample_dict = slice_datasets_by_space(input_dict, location, self.config)
+        sample_dict = slice_datasets_by_space(self.datasets_dict, location, self.config)
         sample_dict = slice_datasets_by_time(sample_dict, t0, self.config)
         sample_dict = load_data_dict(sample_dict)
         sample_dict = diff_nwp_data(sample_dict, self.config)
@@ -486,10 +486,8 @@ class PVNetConcurrentDataset(AbstractPVNetDataset):
         Args:
             t0: init-time for sample
         """
-        input_dict = self.datasets_dict if self.use_xarray else self.light_datasets_dict
-
         # Slice by time then load to avoid loading the data multiple times from disk
-        sample_dict = slice_datasets_by_time(input_dict, t0, self.config)
+        sample_dict = slice_datasets_by_time(self.datasets_dict, t0, self.config)
         sample_dict = load_data_dict(sample_dict)
         sample_dict = diff_nwp_data(sample_dict, self.config)
 


ocf_data_sampler/torch_datasets/pvnet_dataset.py
```diff
@@ -206,7 +228,7 @@ def process_and_combine_datasets(
                 channel_means = self.means_dict["nwp"][nwp_key]
                 channel_stds = self.stds_dict["nwp"][nwp_key]
 
-                da_nwp = (da_nwp - channel_means) / channel_stds
+                da_nwp.data = (da_nwp.data - channel_means) / channel_stds
```
dfulu: Replace the xarray `.data` is faster than the original line and this also works with the LightDataArray

ocf_data_sampler/time_utils.py
```diff
@@ -0,0 +1,155 @@
...
+
+def get_posix_timestamp(
+    datetimes: np.datetime64 | NDArray[np.datetime64],
+) -> np.float64 | NDArray[np.float64]:
+    """Get the POSIX timestamp (seconds since 1970-01-01) for the given datetimes.
+
+    Args:
+        datetimes: the datetimes to get POSIX timestamps for
+    """
+    return datetimes.astype("datetime64[ns]").astype(np.float64) * 1e-9
```
dfulu: To match the `pd.Timestamp.timestamp()` method, this needs to use float64 to avoid losing accuracy

ocf_data_sampler/numpy_sample/generation.py
```diff
@@ -24,9 +24,15 @@ def convert_generation_to_numpy_sample(da: xr.DataArray, t0_idx: int | None = No
         da: Xarray DataArray containing generation data
         t0_idx: Index of the t0 timestamp in the time dimension of the generation data
     """
+    generation_values = da.sel(gen_param="generation_mw").values
```
dfulu: I moved the normalisation in here instead of the main body of the `.process_and_combine_datasets()` method. See comment on that method for the reasoning

PR #399, opened 2026-02-25
Switch to only use `.isel()` indexing

dfulu:
# Pull Request

## Description

The current `LightDataArray` class has a `.sel()` method which is much narrower in compared to the `xr.DataArray.sel()` method. The xarray version can use indexers which are slices, scalars, and list-like values and even  handle multi-dimensional indexing. It can also hande different data-types within these different indexer classes. For example a string scalar `.sel(channel="VIS008")`, a float slice `.sel(latitude=slice(54.0, 55.1)` or even cast strings to datetimes `.sel(time_utc=["2026-02-25 15:00"])`. 

This is a lot of functionality which I don't think we should support, because it would be too much code to maintain and we don't need it. But the fact that we have this `LightDataArray.sel()` implies that functionality which matches xarray will exist. So this method seems risky since we will make a mistake with it at some point. And even writing a set of if-else clauses to guard against indexers which `LightDataArray.sel()` doesn't support would likely be bulky and have a series of edge cases we will miss.

In this PR, I remove the `LightDataArray.sel()` method and instead switch us to use `.isel()` methods everywhere. This feels safer and easier to support in the long run.

This work includes:
- Switch from `.sel()` to `.isel()` everywhere
- Verifying that all coords are in the expected ascending order - this is something that was a risk before anyway
- Removing the multi-dimensional indexing from the NWP time slicing. We almost never do this anyway. Since our NWP delays are 3-6 hours and we never use a history of > 3 hours this was a dusty code-path.
- Update the calculation of contiguous NWP periods to reflect the above
- Update tests

## Checklist:

### DIFF
--- a/ocf_data_sampler/select/__init__.py
+++ b/ocf_data_sampler/select/__init__.py
@@ -4,3 +4,4 @@ from .find_contiguous_time_periods import (
     intersection_of_multiple_dataframes_of_periods,
 )
 from .location import Location
+from .get_indices_in_sorted_unique import get_indices_in_sorted_unique
--- a/ocf_data_sampler/select/get_indices_in_sorted_unique.py
+++ b/ocf_data_sampler/select/get_indices_in_sorted_unique.py
@@ -1,0 +1,26 @@
+"""Helper function to find index positions."""
+
+from typing import TypeVar
+
+import numpy as np
+from numpy.typing import NDArray
+
+T = TypeVar("T", bound=np.generic)
+
+
+def get_indices_in_sorted_unique(
+    values: NDArray[T],
+    query: T | NDArray[T],
+) -> int | NDArray[np.intp]:
+    """Get the integer index(es) of a value(s) within a sorted array.
+
+    Args:
+        values: Sorted 1D array with unique elements.
+        query: Scalar or array of elements to find in `values`.
+    """
+    indices = np.searchsorted(values, query, side="left")
+
+    if (indices >= values.size).any() or (values[indices] != query).any():
+        raise ValueError(f"Not all values in {query} exist in array {values}")
+
+    return indices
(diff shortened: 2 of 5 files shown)


ocf_data_sampler/select/find_contiguous_time_periods.py
```diff
@@ -90,15 +92,15 @@ def trim_contiguous_time_periods(
 
 
 def find_contiguous_t0_periods(
-    datetimes: pd.DatetimeIndex,
+    datetimes: NDArray[np.datetime64],
```
dfulu: Similarly this dtype wasn't updated before

ocf_data_sampler/select/find_contiguous_time_periods.py
```diff
@@ -133,63 +135,78 @@ def find_contiguous_t0_periods(
 
 
 def find_contiguous_t0_periods_nwp(
-    init_times: np.datetime64,
+    init_times: NDArray[np.datetime64],
```
dfulu: I really missed a lot of these in the previous PR

ocf_data_sampler/load/utils.py
```diff
@@ -1,15 +1,13 @@
...
 
-def check_time_unique_increasing(datetimes: xr.DataArray) -> None:
-    """Check that the time dimension is unique and increasing."""
-    time = pd.DatetimeIndex(datetimes)
-    if not (time.is_unique and time.is_monotonic_increasing):
-        raise ValueError("Time dimension must be unique and monotonically increasing")
+def assert_values_increasing(values: np.ndarray, label: str = "values") -> None:
+    """Assert the values are unique and monotonically increasing."""
+    if not (values[:-1] < values[1:]).all():
+        raise ValueError(f"{label} must be increasing")
```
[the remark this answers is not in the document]
AUdaltsova: Thank you!

AUdaltsova (approved):
Thanks for doing this!

PR #404, opened 2026-03-02
Improving `open_nwp` function and dev-dependency change in `pyproject.toml`

siddharth7113:
# Pull Request

## Description
While going through the code base I found in `nwp.py` the function `open_nwp` has an `if-else` block which requires changing everytime new provider is added, I modifed logic outside of the function to update in a private dictionaries, instead of adding `if-else`.

~While installing `dev` dependency, I observed that these were grouped under `[dependency-groups]` , I changed it to `[project.optional-dependencies]` to make the `uv pip install -e .[dev]` work. (If this was intentional, I can revert the changes.)~ (Reverted due to CI breaking)

Fixes #

## How Has This Been Tested?
All tests passes, except one unrelated test being skipped `test_load_gfs`

I haven't run `black` or `ruff` since on running `black` , it is reformatting more than 50 files, there is no defined `pre-commit`

- [x] Yes

- [ ] Yes

## Checklist:

- [] I have made corresponding changes to the documentation

### DIFF
--- a/ocf_data_sampler/load/nwp/nwp.py
+++ b/ocf_data_sampler/load/nwp/nwp.py
@@ -12,7 +12,7 @@ from ocf_data_sampler.load.nwp.providers.gfs import open_gfs
 from ocf_data_sampler.load.nwp.providers.icon import open_icon_eu
 from ocf_data_sampler.load.nwp.providers.ukv import open_ukv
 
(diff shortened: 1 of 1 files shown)


siddharth7113:
I have reverted `pyproject.toml` due to CI breaking, but would highly recommend to make pyproject have an optional dependency group as dev to make `pip install -e .[dev]` work.

PR #406, opened 2026-03-23
Add optional clip

dfulu:
# Pull Request

## Description

This PR introduces an optional clip for the NWP and satellite data which is controlled from the data config. The clip is applied before the data is standardised so the clip values are more physically interpretable. Clipping before applying the normalisation means the clip range is not independent of the mean and std values used.

The clip is optional and backwards compatible since it defaults to no clip.

Also in this PR (sorry):
 - Updates to allow this to work with the new satellite dataset format
 - Remove an unused config file in the tests

## How Has This Been Tested?

I've run the data-sampler locally to make sure the results look sensible. Basically, I set up a config and whilst looking at ECMWF temperature I made sure the maxes and mins of the samples were as expected when I included/excluded the clipping

## Checklist:


felix-e-h-p (changes requested):
Looks great - noticed something that may potentially be a silent issue RE clip_min/max for NWP and Sat.

felix-e-h-p (approved):

PR #410, opened 2026-03-31
Pre-slice the input datasets spatially in the concurrent Dataset for efficiency

dfulu:
# Pull Request

## Description

The `PVNetConcurrentDataset`class loads the entire spatial crop before slicing out windows for each location. This is fine when the input dataset roughly matches the spatial extent of the spread of location. However, I've recently gained a use-case where the input dataset is spatially much larger than the spread of locations. e.g. the dataset covers europe but I'm only interested in the UK.  

This PR adds a spatial slice into the `__init__()` of `PVNetConcurrentDataset` so that the input datasets are reduced cover only the spatial area needed to create the window slices around all the locations. This reduces the amount of unnecessary data loaded for each sample when the input datasets are much wider than required.

Also:
- Removed references to padding since that was removed a while ago

## How Has This Been Tested?

I've run some checks locally and the `PVNetConcurrentDataset` class is already tested in:
```
tests/torch_datasets/test_pvnet_dataset.py::test_pvnet_concurrent_dataset
```

## Checklist:

### DIFF
--- a/ocf_data_sampler/select/select_spatial_slice.py
+++ b/ocf_data_sampler/select/select_spatial_slice.py
@@ -157,7 +157,6 @@ def select_spatial_slice_pixels_multiple(
     bottom_idx = int(idx_y_min - half_height)
     top_idx = int(idx_y_max + half_height)
 
-    # Padding checks
     slice_unavailable = (
         left_idx < 0
         or right_idx > data_width_pixels
--- a/ocf_data_sampler/select/select_spatial_slice.py
+++ b/ocf_data_sampler/select/select_spatial_slice.py
@@ -124,8 +124,8 @@ def select_spatial_slice_pixels_multiple(
         The selected DataArray slice.
 
     Raises:
-        ValueError: If the dimensions are not even or the slice is not allowed
-                    when padding is required.
+        ValueError: If the window dimensions are not even or the slice extends beyond the data
+            boundaries.
     """
     if (width_pixels % 2) != 0:
         raise ValueError("Width must be an even number")
--- a/ocf_data_sampler/select/select_spatial_slice.py
+++ b/ocf_data_sampler/select/select_spatial_slice.py
@@ -61,7 +61,7 @@ def select_spatial_slice_pixels(
         The selected DataArray slice.
 
     Raises:
-        ValueError: If the window dimensions are not even or the slice extends beyond the data 
+        ValueError: If the window dimensions are not even or the slice extends beyond the data
             boundaries.
     """
     if (width_pixels % 2) != 0:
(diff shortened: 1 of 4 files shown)


ocf_data_sampler/select/select_spatial_slice.py
```diff
@@ -104,7 +103,83 @@ def select_spatial_slice_pixels(
...
+        locations: List of locations of interest that will be covered by the returned slice
+        height_pixels: Height of the slice in pixels
+        width_pixels: Width of the slice in pixels
+
+    Returns:
+        The selected DataArray slice.
+
+    Raises:
+        ValueError: If the dimensions are not even or the slice is not allowed
+                    when padding is required.
```
felix-e-h-p: Think only padding reference remaining

ocf_data_sampler/torch_datasets/utils/spatial_slice_for_dataset.py
```diff
@@ -53,3 +57,45 @@ def slice_datasets_by_space(
...
+    locations: list[Location],
+    config: Configuration,
+) -> dict:
+    """Reduce the spatial extent of the datasets to only cover the locations.
+
+    Args:
+        datasets_dict: Dictionary of the input data sources
+        locations: List of locations to reduce to
+        config: Configuration object
+    """
```
felix-e-h-p, line 73: Could be not needed, but what about equivalent validation check same as in slice_datasets_by_space to go before sliced_datasets_dict = {}

felix-e-h-p (changes requested):
Looking all good, just a couple of comments.

Also, what about a test that fails without the buffer via select_spatial_slice_pixels_multiple?

ocf_data_sampler/torch_datasets/utils/spatial_slice_for_dataset.py
dfulu: Since this is an internal function, and the `datasets_dict` is also created internally from the library I don't think that check adds value

dfulu:
Thanks for the review @felix-e-h-p 

> Also, what about a test that fails without the buffer via select_spatial_slice_pixels_multiple?

I'm not sure I understand. Do you mean we'd use `select_spatial_slice_pixels_multiple()` with a window size other than 2 so the buffer isn't applied?

felix-e-h-p:
> Thanks for the review @felix-e-h-p
> 
> > Also, what about a test that fails without the buffer via select_spatial_slice_pixels_multiple?
> 
> I'm not sure I understand. Do you mean we'd use `select_spatial_slice_pixels_multiple()` with a window size other than 2 so the buffer isn't applied?

Ah was thinking more a test that uses window size of 2 and just shows that removing the buffer would cause a failure.

dfulu:
> Ah was thinking more a test that uses window size of 2 and just shows that removing the buffer would cause a failure.

To do that we'd probably need to make the buffer configurable, and then we'd simply be testing that the version where the buffer isn't the default value is incorrect. That doesn't feel like best practice to me

felix-e-h-p (approved):

PR #412, opened 2026-04-14
Make time filtering in PVNet datasets more flexible

dfulu:
# Pull Request

## Description

This pull request enhances the PVNet dataset classes by allowing users to filter data using multiple time periods rather than just a single start and end time. It introduces a new utility function `get_time_periods_mask()` for time period masking and updates both the dataset logic and associated tests to support and validate this more flexible approach.

This makes it easier to perform cross-validation with the dataset. However, note that this is a breaking change and will require corresponding changes in PVNet

## Checklist:


Sukh-P (approved):
Looks good, thanks for adding this in!

PR #413, opened 2026-04-14
Refactoring all providers into a single script

### LINKED ISSUE #147: Generic NWP loader
It would be good to minimise having to add ad-hoc data loaders for NWPs with different schemas as was this case [here in ocf_datapipes](https://github.com/openclimatefix/ocf_datapipes/tree/main/ocf_datapipes/load/nwp/providers), while there may be differences that mean different loaders may be appropriate e.g different coordinate systems, many of the NWPs can use a generic loader which would look similar to what we have for the [ECMWF loader currently](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/load/nwp/providers/ecmwf.py). 

Given this it would be good to rename the ECMWF loader to a more generic name, reference that [here](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/load/nwp/nwp.py) and for any NWPs that can be in this generic format to use this loader. Two NWPs which fall into this category are GFS and MetOffice Global ("gfs", "mo_global") so can be added to the same nwp.py file as NWPs which can use this loader.

siddharth7113:
# Pull Request

Fixes #147 

## Checklist:


ocf_data_sampler/load/nwp/providers/loaders.py
```diff
@@ -0,0 +1,108 @@
+"""NWP provider loaders.
+
+All providers follow the same shape:
+    open zarr -> normalise dim/coord names -> shared post-processing.
```
Sukh-P:
```suggestion
All providers follow the same pipeline:
    open zarr -> standardise dim/coord names -> shared post-processing.
```
Just a few suggestions on wording so it doesn't get confusing since shape often refers to the shape of the data and normalise to normalisation of data which we do elsewhere in this repo so avoiding those words in here to avoid confusion

ocf_data_sampler/load/nwp/providers/loaders.py
```diff
@@ -0,0 +1,108 @@
...
+    if isinstance(ds, xr.Dataset):
+        return get_xr_data_array_from_xr_dataset(ds)
+    return ds
+
+
+def open_ifs(zarr_path: str | list[str]) -> xr.DataArray:
+    """Opens ECMWF IFS / MetOffice Global NWP data."""
+    ds = open_zarr_paths(zarr_path, backend="tensorstore")
+    # LEGACY SUPPORT - older zarrs use "init_time"/"variable" dim names
+    ds = ds.rename({"init_time": "init_time_utc", "variable": "channel"})
```
Sukh-P: So `open_ifs` and `open_gdm` are almost identical apart from the renaming, they could just be one generic open standard_lat_long grid with some checks to see if  init_time or variable are in the ds and in each case if they are then rename them or use the rename map pattern which is in `open_cloudcasting`

Sukh-P: I think there is one issue actually which is when using open_zarr_paths you have to specify whether to use `init_time` or `init_time_utc` but we could have a try/except in there which tries init_time and if not then init_time_utc

Sukh-P:
@siddharth7113 sorry for being slow to review this but it seems like a solid start, left some suggestions and ideas to simplify a bit more, thanks

Sukh-P:
I think you'll need to pull some of the latest changes in the library into your branch as well

Sukh-P:
@siddharth7113 just an FYI we are just updated access to raise PRs against some repos in OCF so we've sent you an invite to a team within OCF's github which will give you the correct access

siddharth7113:
> @siddharth7113 just an FYI we are just updated access to raise PRs against some repos in OCF so we've sent you an invite to a team within OCF's github which will give you the correct access

Thanks Sukh, for the review as well as the invite, I'll make the changes soon.

ocf_data_sampler/load/nwp/providers/loaders.py
```diff
@@ -0,0 +1,100 @@
...
+        return get_xr_data_array_from_xr_dataset(ds)
+    return ds
+
+
+def open_standard_lat_long_grid(zarr_path: str | list[str]) -> xr.DataArray:
+    """Opens NWP data on a standard latitude/longitude grid.
+
+    Used by ECMWF IFS, MetOffice Global, and GDM (e.g. GenCast).
+    """
+    ds = open_zarr_paths(zarr_path, backend="tensorstore")
```
Sukh-P: I am wondering if a simpler/cleaner way to do this would be to add the time_dim as a parameter here, still defaulting to "init_time" and then just gencast (and fgn in the new commit that needs to be merged in) will need "init_time_utc" passed in, this way you can avoid the try/except errors in the open_zarr_paths function and it's a lot less code

Sukh-P:
@siddharth7113 I am really sorry I have taken ages to get back to you on this! I have been off work for a bit, I think this could do with a slight, tweak, please see my comment and also there are some merge conflicts now from a new commit I made, apologies, once these are addressed I think this will be good to go, thanks for all the work on this!

Sukh-P (changes requested):
A few small updates needed I think to streamline this change, thanks

siddharth7113:
> @siddharth7113 I am really sorry I have taken ages to get back to you on this! I have been off work for a bit, I think this could do with a slight, tweak, please see my comment and also there are some merge conflicts now from a new commit I made, apologies, once these are addressed I think this will be good to go, thanks for all the work on this!

Hey Sukh!
No worries, I have made the changes, I have added `gencast` and `fgn` via `functools.partial` in `nwp.py` to pass `time_dim="init_time_utc"`. Should I make it via thin wrapper or is this approach fine?

ocf_data_sampler/load/nwp/providers/loaders.py
```diff
@@ -0,0 +1,109 @@
...
+    return ds
+
+
+def open_standard_lat_long_grid(
+    zarr_path: str | list[str],
+    time_dim: str = "init_time",
+) -> xr.DataArray:
+    """Opens NWP data on a standard latitude/longitude grid.
+
+    Used by ECMWF IFS, MetOffice Global, GDM (e.g. GenCast), and FGN.
```
Sukh-P:
```suggestion
    Used by ECMWF IFS, MetOffice Global, GDM (e.g. GenCast & FGN).
```

Sukh-P:
> > @siddharth7113 I am really sorry I have taken ages to get back to you on this! I have been off work for a bit, I think this could do with a slight, tweak, please see my comment and also there are some merge conflicts now from a new commit I made, apologies, once these are addressed I think this will be good to go, thanks for all the work on this!
> 
> Hey Sukh! No worries, I have made the changes, I have added `gencast` and `fgn` via `functools.partial` in `nwp.py` to pass `time_dim="init_time_utc"`. Should I make it via thin wrapper or is this approach fine?

Yeah that implementation looks good, I think the only thing now I have in mind is whether the folder structure makes sense, I feel it could be flattened slightly now since the 'providers' folder made sense when there was lots of different provider files but now they are in one file I think getting rid of that folder and having the new loaders.py file and the utils file that was in there in the `ocf_data_sampler/load/nwp` directory makes more sense, I am going to ask the team what they think about this as well to get some more opinions before we merge this in, thanks again and I think we can wrap this up this in the next couple of days!

Sukh-P:
Hey @siddharth7113 yeah so if the directory change I mention in the last comment (removing the providers directory and moving the remaining files into the directory above) I think this should be good to go and get merged in

siddharth7113:
> Hey @siddharth7113 yeah so if the directory change I mention in the last comment (removing the providers directory and moving the remaining files into the directory above) I think this should be good to go and get merged in

Hey Sukh, I think this should be good now.

Sukh-P (approved):
Thanks for all the work on this, it looks good now and a nice reduction in repeated code! Well done!

Sukh-P:
Will merge this in now, thanks again @siddharth7113 for tackling this :)

PR #418, opened 2026-07-17
Fix LightDataArray.isel indexing and generation validation

dfulu:
# Pull Request

## Description

This PR fixes some of the edge cases in the `LightDataArray`'s `.isel()` method. We have not encountered any of these edge cases in our code yet, but this guards us more into the future.

In this PR
- There were a few latent bugs for when we used array-indices eg `ds.sel(time=[0,1,2])`. For example, using multiple arrays indexes on a 2D array `x` like `x[index_arrays_1, index_arrays_2]` returns a 1D array instead of a 2D array. This PR fixes these.
- We preserve the correct dimension order when integer, slice, and array indexers are used across non-adjacent dimensions. Numpy will transpose the array during slicing with some combinations of these, which would misallign our data and coordinates.
- We add support for using an array of boolean masks in `isel()`.
- We normalize TensorStore index origins after slicing so subsequent `isel()` calls continue to use relative positional indices. There was a latent bug if we did `da.isel(dim=slice(10, 20)).isel(dim=slice(0, 1))` which would cause an error using the tensorstore backend
- We add tests covering the above edge-cases to avoid regression
- We tighten indexer validation so unsupported dtypes and multidimensional index arrays fail with clear errors.
- We clean up the internals of `LightDataArray` to only have one data-structure for the coordinates. This also lets us create a `LightDataArray using a style more similar to xarray`. E.g. like the familiar form
   ```
   da = xr.DataArray(
      data=temperature,
      dims=["x", "y", "time"],
      coords=dict(
          lon=(["x", "y"], lon),
          lat=(["x", "y"], lat),
          time=(("time), time),
          reference_time=("reference_time", reference_time),
      ),
      attrs=dict(
          description="Ambient temperature.",
          units="degC",
      ),
  )
  ```
- We tighten up the validation on the generation data to make sure the expected generation and capacity variables exist in the data. We also use safer slicing for these variables for the case they are missing.

## Checklist:

### DIFF
--- a/src/ocf_data_sampler/common/lightarray.py
+++ b/src/ocf_data_sampler/common/lightarray.py
@@ -1,6 +1,6 @@
 """A lightweight DataArray-like class."""
 
-from typing import Any, TypedDict
+from typing import Any, TypedDict, overload
 
 import numpy as np
 import tensorstore as ts
@@ -134,9 +134,23 @@ def _normalise_index_origin(data: np.ndarray | ts.TensorStore) -> np.ndarray | t
     return np.asarray(data)
 
 
+@overload
+def _apply_axis_indexers(
+    data: np.ndarray,
+    axis_indexers: tuple[Indexer, ...],
+) -> np.ndarray: ...
+
+
+@overload
+def _apply_axis_indexers(
+    data: ts.TensorStore,
+    axis_indexers: tuple[Indexer, ...],
+) -> ts.TensorStore: ...
+
+
 def _apply_axis_indexers(
     data: np.ndarray | ts.TensorStore,
-    axis_indexers: tuple[Indexer, ...]
+    axis_indexers: tuple[Indexer, ...],
 ) -> np.ndarray | ts.TensorStore:
 
     if len(axis_indexers) != data.ndim:
--- a/src/ocf_data_sampler/common/lightarray.py
+++ b/src/ocf_data_sampler/common/lightarray.py
@@ -297,6 +297,10 @@ class LightDataArray:
             indexers: A dict with keys matching dimensions and values given by integers, slice
                 objects or arrays. `indexer` can be an integer, slice or array-like.
             **indexers_kwargs: The keyword arguments form of indexers.
+
+        Returns:
+            A new LightDataArray. Attributes are copied, while coordinate arrays may be views of
+            the original coordinate arrays rather than independent copies.
         """
         if indexers is not None:
             indexers_kwargs.update(indexers)
--- a/src/ocf_data_sampler/common/lightarray.py
+++ b/src/ocf_data_sampler/common/lightarray.py
@@ -3,9 +3,8 @@
 from typing import Any, TypedDict
 
 import numpy as np
+import tensorstore as ts
 import xarray as xr
-from tensorstore import Future as TensorStoreFuture
-from tensorstore import TensorStore
 from xarray_tensorstore import _TensorStoreAdapter
 
 from ocf_data_sampler.common.types import Indexer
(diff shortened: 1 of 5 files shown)


src/ocf_data_sampler/common/lightarray.py
```diff
@@ -14,41 +13,235 @@
 class LightDataArrayState(TypedDict):
     """Serialized state of a LightDataArray."""
 
-    _data: np.ndarray | TensorStore
+    _data: np.ndarray | ts.TensorStore
```
dfulu, line 16: We're importing and using more function from tensorstore now so I opted to not import the `TensorStore` object by name and just use the `ts` namespace for it

src/ocf_data_sampler/common/lightarray.py
```diff
@@ -14,41 +13,235 @@
...
     dims: tuple[str, ...]
-    coords: dict[str, np.ndarray]
+    coords: dict[str, tuple[tuple[str, ...], np.ndarray]]
     attrs: dict[Any, Any]
-    coord_dims: dict[str, tuple[str, ...]]
+
+
+def _validate_coords(
```
dfulu, line 22: We use this new function to do more validation on input coordinates and data of the `LightDataArray` object

src/ocf_data_sampler/common/lightarray.py
```diff
@@ -14,41 +13,235 @@
...
+
+    for i, dim in enumerate(dims):
+        if data.shape[i] != coords[dim][1].shape[0]:
+            raise ValueError(
+                f"Dimension '{dim}' has size {data.shape[i]} in the data but size "
+                f"{coords[dim][1].shape[0]} in the coordinates.",
+            )
+
+
+def _sanitise_indexer(dim: str, dim_size: int, indexer: Indexer) -> Indexer:
```
dfulu, line 82: This new function catches some wrong inputs to `.isel()` and makes the indexers a little more standard

src/ocf_data_sampler/common/lightarray.py
```diff
@@ -14,41 +13,235 @@
...
+    if not np.issubdtype(indexer.dtype, np.integer):
+        raise TypeError(
+            f"Indexer for dimension '{dim}' must have an integer or boolean dtype. "
+            f"Got dtype {indexer.dtype}.",
+        )
+
+    return indexer
+
+
+def _normalise_index_origin(data: np.ndarray | ts.TensorStore) -> np.ndarray | ts.TensorStore:
```
dfulu, line 130: This deals with an edge case in Tensorstore. Tensorstore's indexing works differently than numpy. If you do `x[slice(10, 20)][slice(0, 10)]` you get an error because tensorstore remembers that `x[slice(10, 20)]` had the indices 10-20, so the second slice fails. However `x[slice(10, 20)][slice(10, 20)]` would work in tensorstore but fail in numpy. 

This helper function is used to make tensorstore play nice with the style of indexing we use with numpy

src/ocf_data_sampler/common/lightarray.py
```diff
@@ -14,41 +13,235 @@
...
+    """Translate TensorStore domains to zero so subsequent indexing remains positional."""
+    if isinstance(data, ts.TensorStore):
+        return data[ts.d[:].translate_to[0]]
+    return np.asarray(data)
+
+
+@overload
+def _apply_axis_indexers(
```
dfulu, line 138: These overloads are for `mypy`

src/ocf_data_sampler/common/lightarray.py
```diff
@@ -14,41 +13,235 @@
...
+    data: ts.TensorStore,
+    axis_indexers: tuple[Indexer, ...],
+) -> ts.TensorStore: ...
+
+
+def _apply_axis_indexers(
```
dfulu, line 151: This is just a little helper function to be called from `isel()`. I split it out to keep `isel()` cleaner

src/ocf_data_sampler/common/lightarray.py
```diff
@@ -14,41 +13,235 @@
...
+    __slots__ = ["_data", "_future", "attrs", "coords", "dims"]
 
     def __init__(
         self,
-        data: np.ndarray | TensorStore,
+        data: np.ndarray | ts.TensorStore,
+        coords: dict[str, tuple[tuple[str, ...], np.ndarray]],
         dims: tuple[str, ...],
-        coords: dict[str, np.ndarray],
-        coord_dims: dict[str, tuple[str, ...]],
```
dfulu, line 34: `coord_dims` is gone since we move the coordinate data structure closer to the one used in xarray. That keeps all the relevant information in `coords`

src/ocf_data_sampler/common/lightarray.py
```diff
@@ -57,65 +250,40 @@ def data(self, value: np.ndarray | TensorStore) -> None:
             )
 
         self._data = value
-        self.future = None
+        self._future = None
```
dfulu, line 253: Implicitly make this private

src/ocf_data_sampler/common/lightarray.py
```diff
@@ -57,65 +250,40 @@ def data(self, value: np.ndarray | TensorStore) -> None:
...
             data = da.variable._data.array
         elif isinstance(da.variable._data, np.ndarray):
             data = da.variable._data
         else:
             raise ValueError(f"Data backend of type {type(da.variable._data)} not supported.")
 
-        coord_values: dict[str, np.ndarray] = {}
-        coord_dims: dict[str, tuple[str, ...]] = {}
-
-        for k, v in da.coords.items():
```
dfulu, line 77: This kind of validation was moved to the helper function

src/ocf_data_sampler/common/lightarray.py
```diff
@@ -57,65 +250,40 @@ def data(self, value: np.ndarray | TensorStore) -> None:
...
             attrs=da.attrs,
+            skip_validation=True,
         )
 
     def to_xarray(self) -> xr.DataArray:
-        """Convert to an Xarray DataArray.
-
-        Note this loads the data eagerly.
-        """
-        coords_dict: dict[str, Any] = {}
```
dfulu, line 103: This is no longer needed with the new `coords` structure

tests/common/test_lightarray.py
```diff
@@ -64,3 +66,54 @@ def test_isel(xr_gen, xr_ukv, xr_sat):
         sliced_fda = fda.isel(**indexer)
         sliced_xda = xda.isel(**indexer)
         assert sliced_fda.to_xarray().equals(sliced_xda)
+
+
+@pytest.mark.parametrize(
```
dfulu, line 71: This introduces some new test cases covering the shortcomings mentioned in the PR description

peterdudfield (approved):

PR #420, opened 2026-07-20
Fix helper function to get sample from PVNetDataset 

### LINKED ISSUE #419: fix helper function to get sample from PVNetDataset
## Describe the bug
Issue when using the helper function [PVNetDataset.get_sample](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/torch_datasets/pvnet_dataset.py) for a given time t0 and location_id. This is caused by [valid_sample_request](https://github.com/openclimatefix/ocf-data-sampler/blob/36accbc00e9963f6467645e12f0e5792922c1778/ocf_data_sampler/torch_datasets/pvnet_dataset.py#L426) which [checks t0 exists in the dataframe as an index, rather than t0 value](https://github.com/openclimatefix/ocf-data-sampler/blob/36accbc00e9963f6467645e12f0e5792922c1778/ocf_data_sampler/torch_datasets/pvnet_dataset.py#L441)

## To Reproduce
```python
import pandas as pd
from ocf_data_sampler.torch_datasets.pvnet_dataset import PVNetDataset
config_filename = 'example_config.yaml'
pvnet_dataset = PVNetDataset(config_filename)
# Choose t0 and location_id that definitely exists in dataset
t0 = pvnet_dataset.valid_t0_and_location_ids['t0'][0]
location_id = pvnet_dataset.valid_t0_and_location_ids['location_id'][0]
# The following line fails
sample = pvnet_dataset.get_sample(t0=t0, location_id=location_id)
```
Fails with: 
```python
ValueError: Input t0 time '2019-07-01 06:00:00' and location id '1' pair not in valid t0 and location pairs
```

## Expected behavior
Should not raise error - the time and location_id exists.

## Additional context
Only affects the helper function as rest of code uses [_get_sample](https://github.com/openclimatefix/ocf-data-sampler/blob/36accbc00e9963f6467645e12f0e5792922c1778/ocf_data_sampler/torch_datasets/pvnet_dataset.py#L379)

I think the simplest fix is we could change the line that looks for [`t0 in self.valid_t0_and_location_ids.index`](https://github.com/openclimatefix/ocf-data-sampler/blob/36accbc00e9963f6467645e12f0e5792922c1778/ocf_data_sampler/torch_datasets/pvnet_dataset.py#L441) to something like `t0 in self.valid_t0_and_location_ids['t0'].values`

lm2612:
# Pull Request

## Description

Fixes #419 
For datasets that use `complete_generation=False`, the helper function `get_sample` would call `validate_sample_request` to check if the time and location is valid. This checks the dataframe `valid_t0_and_location_ids`, but searched for `t0` (a pandas timestamp) in the index of the dataframe (integers). This would cause a ValueError even when valid `t0` and `location_id` were provided.

This PR changes the line in `validate_sample_request` to search in the the dataframe column`t0` instead, which now finds `t0` if it exists as a valid timestamp, and does not throw a ValueError.

## How Has This Been Tested?

Extended tests for pvnet datasets to check if we can use `get_sample`:
1. test_pvnet_dataset - where `complete_generation=True`. 
2. test_pvnet_dataset_sites - where `complete_generation=False`. 

Both use `t0` and `location_id` from the first valid time and location available and then check that we can retrieve a sample with these.

- [x] Yes

## Checklist:

### DIFF
--- a/ocf_data_sampler/torch_datasets/pvnet_dataset.py
+++ b/ocf_data_sampler/torch_datasets/pvnet_dataset.py
@@ -437,10 +437,9 @@ class PVNetDataset(AbstractPVNetDataset):
             if location_id not in self.location_lookup:
                 raise ValueError(f"Input location '{location_id}' not known")
         else:
+            t0_inds = self.valid_t0_and_location_ids["t0"]==t0
             if not (
-                t0 in self.valid_t0_and_location_ids["t0"].values
-                and location_id in self.valid_t0_and_location_ids[
-                    self.valid_t0_and_location_ids["t0"]==t0]["location_id"].values
+                location_id in self.valid_t0_and_location_ids[t0_inds]["location_id"].values
             ):
                 raise ValueError(
                     f"Input t0 time '{t0!s}' and location id '{location_id}' "
--- a/ocf_data_sampler/torch_datasets/pvnet_dataset.py
+++ b/ocf_data_sampler/torch_datasets/pvnet_dataset.py
@@ -437,9 +437,9 @@ class PVNetDataset(AbstractPVNetDataset):
             if location_id not in self.location_lookup:
                 raise ValueError(f"Input location '{location_id}' not known")
         else:
-            t0_inds = self.valid_t0_and_location_ids["t0"]==t0
+            t0_idxs = self.valid_t0_and_location_ids["t0"]==t0
             if not (
-                location_id in self.valid_t0_and_location_ids[t0_inds]["location_id"].values
+                location_id in self.valid_t0_and_location_ids[t0_idxs]["location_id"].values
             ):
                 raise ValueError(
                     f"Input t0 time '{t0!s}' and location id '{location_id}' "
--- a/ocf_data_sampler/torch_datasets/pvnet_dataset.py
+++ b/ocf_data_sampler/torch_datasets/pvnet_dataset.py
@@ -438,9 +438,7 @@ class PVNetDataset(AbstractPVNetDataset):
                 raise ValueError(f"Input location '{location_id}' not known")
         else:
             t0_idxs = self.valid_t0_and_location_ids["t0"]==t0
-            if not (
-                location_id in self.valid_t0_and_location_ids[t0_idxs]["location_id"].values
-            ):
+            if location_id not in self.valid_t0_and_location_ids[t0_idxs]["location_id"].values:
                 raise ValueError(
                     f"Input t0 time '{t0!s}' and location id '{location_id}' "
                     f"pair not in valid t0 and location pairs",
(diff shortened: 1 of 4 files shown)


ocf_data_sampler/torch_datasets/pvnet_dataset.py
```diff
@@ -438,8 +438,9 @@ def validate_sample_request(self, t0: pd.Timestamp, location_id: int) -> None:
                 raise ValueError(f"Input location '{location_id}' not known")
         else:
             if not (
-                t0 in self.valid_t0_and_location_ids.index
-                and self.valid_t0_and_location_ids.loc[t0, "location_id"] == location_id
+                t0 in self.valid_t0_and_location_ids["t0"].values
```
peterdudfield: potentially dont need this line now

ocf_data_sampler/torch_datasets/pvnet_dataset.py
```diff
@@ -438,8 +438,9 @@ def validate_sample_request(self, t0: pd.Timestamp, location_id: int) -> None:
...
+                and location_id in self.valid_t0_and_location_ids[
+                    self.valid_t0_and_location_ids["t0"]==t0]["location_id"].values
```
peterdudfield: make sure there is a test if `t0` no in `valid_t0_and_location_ids["t0"]=`

tests/torch_datasets/test_pvnet_dataset.py
```diff
@@ -109,6 +109,13 @@ def test_pvnet_dataset(pvnet_config_filename):
     # The config uses 3 periods each of which generates a sin and cos embedding
     assert sample["t0_embedding"].shape == (6,)
 
+    # Test helper function get_sample to retrieve sample by t0 and location_id
+    assert dataset.complete_generation
+    t0 = dataset.valid_t0_times[0]
+    location_id = next(iter(dataset.location_lookup))
+    sample = dataset.get_sample(t0=t0, location_id=location_id)
+    assert isinstance(sample, dict)
```
peterdudfield: Could make this a seperate test. There's quite a lot of step up, so if it takes longer than 0.1 seconds then keep how it is

ocf_data_sampler/torch_datasets/pvnet_dataset.py
peterdudfield: You could use `with pytest.raises(`

ref - https://docs.pytest.org/en/7.1.x/how-to/assert.html

ocf_data_sampler/torch_datasets/pvnet_dataset.py
```diff
@@ -437,9 +437,9 @@ def validate_sample_request(self, t0: pd.Timestamp, location_id: int) -> None:
             if location_id not in self.location_lookup:
                 raise ValueError(f"Input location '{location_id}' not known")
         else:
+            t0_inds = self.valid_t0_and_location_ids["t0"]==t0
```
peterdudfield: t0_idxs

peterdudfield: t0_indexs

peterdudfield (approved):

PR #422, opened 2026-07-21
Refactor NWP load module

dfulu:
# Pull Request

## Description

Blocked by #418 

This PR refactors the load NWP module to make it cleaner and more minimal. In doing this I refactored a few neighbouring files to share functionality and move some functions around to better locations.

Changes include
- Remove support for the GFS dataset in its current format
- Remove support for the ICON-EU dataset in its current format
- Remove support for UKV to have coords named `x` and `y`. They must explicitly be `x_osgb` and `y_osgb`
- Consolidate out xarray-tensortore functions for opening zarrs. There is now only one public function to open zarrs instead of 3
- Consolidate our list of NWP providers so we don't have to duplicate it. The same list was copied in 3 places
- Remove support for the `public` parameter in config and data loaders. We don't have a use for this, and it can be bypassed by the user setting their AWS/GCP/etc credentials globally
- Refactor the loading and validation functions in `load/nwp.py`
- Fix typos
- Change the expected dtypes of different coordinates and data to be the minimal we need to check. E.g. spatial coords can be floats or ints, data must be float
- Add more validation to the data and coordinate shapes
- Type-hinting improvements
- Rename `get_xr_data_array_from_xr_dataset() -> extract_single_data_array()` to be less cumbersome of a name
- Small updates satellite loader and tests for the new satellite dataset format

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,3 +141,9 @@ line-ending = "auto"
 docstring-code-format = true
 docstring-code-line-length = 100
 
+
+# --- TESTING CONFIGURATION --- #
+
+[tool.pytest.ini_options]
+testpaths = ["tests"]
+
--- a/src/ocf_data_sampler/common/xr_tensorstore.py
+++ b/src/ocf_data_sampler/common/xr_tensorstore.py
@@ -15,9 +15,10 @@ References:
 """
 
 import logging
-import os.path
+import os
 import re
-from typing import Any, cast
+from glob import glob, has_magic
+from typing import Any, TypeAlias, cast
 
 import tensorstore as ts
 import xarray as xr
@@ -29,6 +30,9 @@ from xarray_tensorstore import (
 )
 
 logger = logging.getLogger(__name__)
+
+ZarrPath: TypeAlias = str | os.PathLike[str]
+ZarrSource: TypeAlias = ZarrPath | list[ZarrPath] | tuple[ZarrPath, ...]
 
 
 def _zarr_spec_from_path(path: str, zarr_format: int) -> dict[str, Any]:
@@ -90,7 +94,34 @@ def _tensorstore_open_zarrs(
     return arrays
 
 
-def open_zarr(
+def open_zarr_paths(zarr_path: ZarrSource, concat_dim: str | None = None) -> xr.Dataset:
+    """Open one or more Zarr stores using TensorStore.
+
+    Args:
+        zarr_path: A path, local glob pattern, or sequence of paths.
+        concat_dim: Dimension along which multiple stores are concatenated.
+    """
+    if isinstance(zarr_path, str | os.PathLike):
+        path = os.fspath(zarr_path)
+        if not has_magic(path):
+            return _open_single_zarr(path)
+        paths = sorted(glob(path))
+    else:
+        paths = [os.fspath(path) for path in zarr_path]
+
+    if not paths:
+        raise ValueError(f"No Zarr stores found for {zarr_path!r}")
+
+    if len(paths) == 1:
+        return _open_single_zarr(paths[0])
+
+    if concat_dim is None:
+        raise ValueError("`concat_dim` must be specified when opening multiple Zarr stores")
+
+    return _open_and_concat_zarrs(paths, concat_dim)
+
+
+def _open_single_zarr(
     path: str,
     context: ts.Context | None = None,
     mask_and_scale: bool = True,
@@ -129,12 +160,11 @@ def open_zarr(
     return cast("xr.Dataset", ds.copy(data=new_data))
 
 
-def open_zarrs(
+def _open_and_concat_zarrs(
     paths: list[str],
     concat_dim: str,
     context: ts.Context | None = None,
     mask_and_scale: bool = True,
-    data_source: str = "unknown",
 ) -> xr.Dataset:
     """Open multiple zarrs with TensorStore.
 
@@ -143,7 +173,6 @@ def open_zarrs(
         concat_dim: Dimension along which to concatenate the data variables.
         context: TensorStore context.
         mask_and_scale: Whether to mask and scale the data.
-        data_source: Which data source is being opened. Used for warning context.
 
     Returns:
         Concatenated Dataset with all data variables opened via TensorStore.
@@ -166,9 +195,8 @@ def open_zarrs(
         )
     except ValueError:
         logger.warning(
-            f"Coordinate mismatch found in {data_source} input data. Opening with "
-            "`join='override'` to ignore coordinate mismatches. THIS MAY CAUSE UNEXPECTED "
-            "BEHAVIOUR!",
+            f"Coordinate mismatch found when opening paths {paths}. Opening with `join='override'` "
+            "to ignore coordinate mismatches. THIS MAY CAUSE UNEXPECTED BEHAVIOUR.",
         )
         ds = xr.concat(
             ds_list,
--- a/src/ocf_data_sampler/config/model.py
+++ b/src/ocf_data_sampler/config/model.py
@@ -10,16 +10,8 @@ from typing import Literal
 from pydantic import BaseModel, ConfigDict, Field, RootModel, field_validator, model_validator
 from typing_extensions import override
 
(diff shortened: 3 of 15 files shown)


src/ocf_data_sampler/common/xr_tensorstore.py
```diff
@@ -30,6 +31,9 @@
 
 logger = logging.getLogger(__name__)
 
+ZarrPath: TypeAlias = str | os.PathLike[str]
```
dfulu, line 34: Make these type aliases since we repeat them in a bunch of places

src/ocf_data_sampler/common/xr_tensorstore.py
```diff
@@ -90,7 +94,34 @@ def _tensorstore_open_zarrs(
     return arrays
 
 
-def open_zarr(
+def open_zarr_paths(zarr_path: ZarrSource, concat_dim: str | None = None) -> xr.Dataset:
```
dfulu, line 97: This is now the only public function we should be using to open zarrs with xarray-tensorstore

src/ocf_data_sampler/common/xr_tensorstore.py
```diff
@@ -166,9 +195,8 @@ def open_zarrs(
         )
     except ValueError:
         logger.warning(
-            f"Coordinate mismatch found in {data_source} input data. Opening with "
-            "`join='override'` to ignore coordinate mismatches. THIS MAY CAUSE UNEXPECTED "
-            "BEHAVIOUR!",
+            f"Coordinate mismatch found when opening paths {paths}. Opening with `join='override'` "
```
dfulu, line 198: Use the paths in this warning message. This allows us to see the same information without needing to have a `data_source` label parameter just for this message

src/ocf_data_sampler/config/model.py
```diff
@@ -10,16 +10,8 @@
 from pydantic import BaseModel, ConfigDict, Field, RootModel, field_validator, model_validator
 from typing_extensions import override
 
-NWP_PROVIDERS = [
```
dfulu, line 13: These are imported from `ocf_data_sampler.load.nwp` now

tests/load/test_conventions.py
```diff
@@ -0,0 +1,26 @@
...
+
+def test_validate_coords_rejects_multidimensional_coord():
+    ds = xr.Dataset(
+        coords={
+            "latitude": (
+                ("x", "y"),
+                np.array([[50.0, 51.0], [52.0, 53.0]]),
+            ),
+        },
+    )
```
peterdudfield, line 16: we dont every have this? I thought for UKV we did at one point?

src/ocf_data_sampler/config/model.py
```diff
@@ -327,8 +320,6 @@ class Generation(TimeWindowMixin, DropoutMixin):
         "to read from alternative filesystems.",
     )
 
-    public: bool = Field(False, description="Whether the NWP data is public or private")
```
dfulu, line 330: Only GFS was using this, we don't support it anymore

tests/common/test_xr_tensorstore.py
```diff
@@ -2,7 +2,7 @@
 import pytest
 import xarray as xr
 
-from ocf_data_sampler.common.xr_tensorstore import open_zarr, open_zarrs
+from ocf_data_sampler.common.xr_tensorstore import _open_and_concat_zarrs, _open_single_zarr
```
peterdudfield: should you test `open_zarr_paths` as well. Or just this. 

The _ functions could change, but we need to make sure the public function works correctly

src/ocf_data_sampler/load/conventions.py
```diff
@@ -31,15 +66,21 @@ def make_spatial_coords_increasing(ds: xr.Dataset, x_coord: str, y_coord: str) -
...
 
-def get_xr_data_array_from_xr_dataset(ds: xr.Dataset) -> xr.DataArray:
+def extract_single_data_array(ds: xr.Dataset, promote_attrs: bool = True) -> xr.DataArray:
     """Return underlying xr.DataArray from passed xr.Dataset.
 
     Checks only one variable is present and returns it as an xr.DataArray.
 
     Args:
         ds: xr.Dataset to extract xr.DataArray from
+        promote_attrs: Whether to promote dataset attributes to the data array
```
dfulu, line 76: This was made with the intention of pulling the `"area"` attribute from the satellite data, but can be used more widely

peterdudfield (approved):
I approved, but left a 2 comments

src/ocf_data_sampler/load/generation.py
```diff
@@ -4,51 +4,58 @@
 
 Dimensions: (time_utc, location_id)
 Data Variables:
-    generation_mw (time_utc, location_id): float32 representing the generation in MW
-    capacity_mwp (time_utc, location_id): float32 representing the capacity in MW peak
+    generation_mw (time_utc, location_id): The generation in MW
```
dfulu, line 7: This can be float32 or float64 so we remove this here. The types are cleanly stated in the code below

dfulu: The sme is true for all the types below

src/ocf_data_sampler/load/generation.py
```diff
@@ -4,51 +4,58 @@
...
+    # TODO: establish if xarray tensorstore could be used here
+    ds = xr.open_dataset(zarr_path, engine="zarr", chunks=None)
 
     if set(ds.data_vars) != {"generation_mw", "capacity_mwp"}:
         raise ValueError(
             f"Generation data should have variables 'generation_mw' and 'capacity_mwp', "
             f"but found {set(ds.data_vars)} instead."
         )
 
+    coord_dtypes = {
```
dfulu, line 43: These are now the expected dtypes of the coords

src/ocf_data_sampler/load/nwp.py
```diff
@@ -1,285 +1,79 @@
...
-    if isinstance(zarr_path, list | tuple):
-        ds = open_zarrs(zarr_path, concat_dim=time_dim, data_source="nwp").sortby(time_dim)
-    else:
-        ds = open_zarr(zarr_path)
-    return ds
-
-
-def _validate_nwp_data(data_array: xr.DataArray, provider: str) -> None:
-    """Validates the structure and data types of a loaded NWP DataArray.
+def open_nwp(zarr_path: ZarrSource, provider: str) -> xr.DataArray:
```
dfulu, line 31: This single function opens all NWPs now

src/ocf_data_sampler/load/nwp.py
```diff
@@ -1,285 +1,79 @@
...
     )
-    da = ds.to_array(dim="channel")
-    return _canonicalize_regular_grid_layout(da, x_coord="longitude", y_coord="latitude")
 
+    ds = make_spatial_coords_increasing(ds, x_coord=x_coord, y_coord=y_coord)
 
-def open_icon_eu(zarr_path: str | list[str]) -> xr.DataArray:
-    """Opens DWD ICON-EU data.
+    assert_values_unique_increasing(ds["init_time_utc"].values, "init_time_utc")
+    assert_values_unique_increasing(ds["step"].values, "step")
```
dfulu, line 60: This is extra validation

pyproject.toml
```diff
@@ -141,3 +141,9 @@ line-ending = "auto"
 docstring-code-format = true
 docstring-code-line-length = 100
 
+
+# --- TESTING CONFIGURATION --- #
+
+[tool.pytest.ini_options]
```
dfulu, line 147: This is just an extra to make sure pytest only looks at tests in the test directory rather than any local files which might have the word test in them

tests/load/test_conventions.py
dfulu: We did but never intentionally. There was a point where it existed in the zarrs but we were never using it for any ML stuff

tests/common/test_xr_tensorstore.py
dfulu: Yeh good point, I'll update those

Issue #402, opened 2026-02-27
The same dropout value is used for all NaNs


dfulu:
The implementation of the configurable dropout value is problematic or misleading. [In the config](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/config/model.py#L97-L100), it states that it is used to replace values that are dropped out, but in reality is it used to [replace any NaN values](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/torch_datasets/utils/merge_and_fill_utils.py#L29) in the data. This could be NaNs that are in the raw zarr - which is possible with satellite data at least. 

We should define what the desired behaviour is and make this NaN filling more clear

PR #423, opened 2026-07-21
Housekeeping

dfulu:
# Pull Request

## Description

This PR groups together a bunch of tiny housekeeping changes. eg spelling mistakes, internal variable rename etc

## Checklist:

### DIFF
--- a/src/ocf_data_sampler/common/indexing.py
+++ b/src/ocf_data_sampler/common/indexing.py
@@ -28,7 +28,5 @@ def get_indices_in_sorted_unique(
 
 def assert_values_unique_increasing(values: np.ndarray, label: str = "values") -> None:
     """Assert the values are unique and monotonically increasing."""
-    if len(np.unique(values))!=len(values):
-        raise ValueError(f"{label} must be unique")
     if not (values[:-1] < values[1:]).all():
-        raise ValueError(f"{label} must be increasing")
+        raise ValueError(f"{label} must be strictly increasing")
--- a/src/ocf_data_sampler/config/model.py
+++ b/src/ocf_data_sampler/config/model.py
@@ -87,10 +87,10 @@ class DropoutMixin(Base):
         "floats (probability that dropout of the corresponding timedelta is applied)",
     )
 
-    dropout_value: float = Field(
+    dropout_fill_value: float = Field(
         default=0.0,
-        description="The value to use for dropped out values. "
-        "Idea is to use -1, but to be backwards comptaible we've put the default as 0")
+        description="The value used to fill in dropped out data or any missing values."
+    )
 
     @field_validator("dropout_timedeltas_minutes")
     def dropout_timedeltas_minutes_negative(cls, v: list[int]) -> list[int]:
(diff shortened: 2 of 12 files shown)


src/ocf_data_sampler/common/indexing.py
```diff
@@ -28,7 +28,5 @@ def get_indices_in_sorted_unique(
 
 def assert_values_unique_increasing(values: np.ndarray, label: str = "values") -> None:
     """Assert the values are unique and monotonically increasing."""
-    if len(np.unique(values))!=len(values):
```
dfulu, line 31: This check was redundant alongside the check below. The check below covers both uniqueness and ordering in a single pass through the array

src/ocf_data_sampler/config/model.py
```diff
@@ -87,10 +87,10 @@ class DropoutMixin(Base):
         "floats (probability that dropout of the corresponding timedelta is applied)",
     )
 
-    dropout_value: float = Field(
+    dropout_fill_value: float = Field(
```
dfulu, line 90: Make this more explicit and improve the description. This is related to #402

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -43,11 +41,6 @@
 )
 from ocf_data_sampler.spatial import Location, convert_coordinates, find_coord_system
 
-# Ignore pydantic warning which doesn't cause an issue
```
dfulu, line 46: We don't actually need this filter or xarray setting anymore

src/ocf_data_sampler/features/time_encodings.py
dfulu, line 1: This is just a couple of extra error catches

src/ocf_data_sampler/select/time_slice.py
```diff
@@ -23,9 +24,9 @@ def select_time_slice(
         interval_end: The end of the interval with respect to t0
         time_resolution: Distance between neighbouring timestamps
     """
-    date_range = np.array([t0 + interval_start, t0 + interval_end])
```
dfulu, line 26: There is a function named `date_range()` which is imported into this module. This variable shouldn't have had this name

src/ocf_data_sampler/select/time_slice.py
```diff
@@ -36,8 +37,8 @@ def select_time_slice_nwp(
     interval_start: np.timedelta64,
     interval_end: np.timedelta64,
     time_resolution: np.timedelta64,
-    dropout_timedeltas: list[np.timedelta64] | None = None,
```
dfulu, line 39: These types were wrong

src/ocf_data_sampler/select/time_slice.py
```diff
@@ -85,6 +82,15 @@ def select_time_slice_nwp(
 
     # Find the most recent available init-time <= t0_available
     selected_init_time_index = np.searchsorted(all_init_times, t0_available, side="right") - 1
+
+    # If the selected init-time index is -1, this means that t0_available is before the first
+    # available init-time in the data
+    if selected_init_time_index == -1:
```
dfulu, line 88: This is an extra edge case catch. We've never fallen foul of this before since we calculate the available t0 times up top, but this is a cheap check and if this did come up it would be confusing to debug

peterdudfield (approved):

Issue #411, opened 2026-03-31
Spatial slice window may not include location


dfulu:
In the current spatial slicing the exact position of the target can be offset or even outside the slice.

In the code, we [calculate the index ](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/select/select_spatial_slice.py#L40) closest to the location. Then we [slice around it](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/select/select_spatial_slice.py#L73-L81). 

This leads to some unexpected behaviour where the slice may not include the location. See B in the diagram below where the closest grid point is to the left of the location.

This could be solved in a bunch of ways which make the location more central in the slice, but will break backwards compatibility

<img width="2508" height="2211" alt="Image" src="https://github.com/user-attachments/assets/1becd85e-c358-459a-a078-91ca3d060c81" />

Sukh-P:
Decision: Do this!
To check: see if there are any issues in PVNet in terms of supporting odd number of spatial slices

dfulu:
Completed in the dev branch

PR #424, opened 2026-07-21
Upgrade spatial window slicing

dfulu:
# Pull Request

## Description

We noted some shortcoming in our spatial window slicing in #411 where depending on the configuration the desired location would not be in the centre of the image. 

This PR upgrades our spatial slicing to:
- Fix the spatial slicing so the desired location is always in the centre
- Allow spatial slices with odd sized windows
- Refactor the surrounding code to make this all cleaner
- Add to the tests to lock in this behaviour

Solves #411

## Checklist:

### DIFF
--- a/src/ocf_data_sampler/select/spatial_slice.py
+++ b/src/ocf_data_sampler/select/spatial_slice.py
@@ -18,7 +18,9 @@ def _get_central_index(
     Args:
         values: The array of values to search.
         val: The value to find the closest index for.
-        method: Method to use for finding the index ("nearest" or "left").
+        method: Method to use for finding the index ("nearest" or "left"). If set to "nearest", the
+            index of the closest value will be returned. If set to "left", the index of the closest
+            value that is less than or equal to `val` will be returned.
 
     Returns:
         The index of the closest value.
@@ -32,17 +34,14 @@ def _get_central_index(
             f"{val} is not in the interval {values[0]}: {values[-1]}",
         )
 
-    def get_nearest_index(vals: np.ndarray, val: float) -> int:
-        """Get the index of the nearest value in vals to val."""
-        idx = np.searchsorted(vals, val, side="left") - 1
-        if idx < len(vals) - 1 and (abs(vals[idx+1] - val) < abs(vals[idx] - val)):
-            idx += 1
-        return idx
-
     if method == "left":
+        # Get the index of the closest value that is less than or equal to val
         index = np.searchsorted(values, val, side="right") - 1
     elif method == "nearest":
-        index = get_nearest_index(values, val)
+        # Get the index of the closest value to val
+        index = np.searchsorted(values, val, side="left") - 1
+        if index < len(values) - 1 and (abs(values[index+1] - val) < abs(values[index] - val)):
+            index += 1
     else:
         raise ValueError(f"Unknown method: {method}")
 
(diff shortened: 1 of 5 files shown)


tests/select/conftest.py
```diff
@@ -35,17 +35,3 @@ def da_sample():
         np.random.normal(size=(len(datetimes),)),
         coords={"time_utc": (["time_utc"], datetimes)},
     )
-
-
-@pytest.fixture(scope="module")
```
dfulu, line 40: This is only used in `test_spatial_slice.py` so I moved it in there

src/ocf_data_sampler/select/spatial_slice.py
```diff
@@ -1,46 +1,95 @@
...
+    method: Literal["nearest", "left"],
+) -> int:
+    """Find pixel index location closest to given value.
 
     Args:
-        da: The DataArray-like object.
-        location: The Location object representing the point of interest.
+        values: The array of values to search.
+        val: The value to find the closest index for.
+        method: Method to use for finding the index ("nearest" or "left").
```
peterdudfield: could you exaplin a bit more what nearest in left is here. Im unsre if left is the very left of the window, or if its just to the left of the location you want

src/ocf_data_sampler/select/spatial_slice.py
```diff
@@ -1,46 +1,95 @@
...
-    if not (y_vals[0] < y < y_vals[-1]):
+    if not (values[0] < val < values[-1]):
         raise ValueError(
-            f"{y} is not in the interval {y_vals[0]}: {y_vals[-1]}",
+            f"{val} is not in the interval {values[0]}: {values[-1]}",
         )
 
-    closest_x = np.argmin(np.abs(x_vals - x))
-    closest_y = np.argmin(np.abs(y_vals - y))
+    def get_nearest_index(vals: np.ndarray, val: float) -> int:
```
peterdudfield: do you need a seperate function for this, or just put eh logic in the code below.

src/ocf_data_sampler/select/spatial_slice.py
```diff
@@ -57,50 +106,32 @@ def select_spatial_slice_pixels(
...
+    y_values = da[y_dim].values
 
-    left_idx = int(center_idx_x - half_width)
-    right_idx = int(center_idx_x + half_width)
-    bottom_idx = int(center_idx_y - half_height)
-    top_idx = int(center_idx_y + half_height)
+    # If odd window size, get index of nearest pixel, else get index of left pixel to ensure the
+    # location is within the returned slice
+    x_method = "nearest" if (width_pixels % 2) == 1 else "left"
+    y_method = "nearest" if (height_pixels % 2) == 1 else "left"
```
peterdudfield, line 121: if left is the index just to theleft of the cetner, maybe call it `left_center` to help explain that

src/ocf_data_sampler/select/spatial_slice.py
dfulu: Just the index to the left of the point-of-interest. I'll add comments

src/ocf_data_sampler/select/spatial_slice.py
dfulu: I'll add to the documentation

peterdudfield (approved):

PR #426, opened 2026-07-22
Improve NWP dropout argument passing

dfulu:
# Pull Request

## Description

This PR has a few changes mostly related to the dropout used for NWPs

- Modify the `select_time_slice_nwp()` function so that it accepts a list of dropout probabilities linked to the dropout values. This matches the behaviour of the dropout function used for satellite and generation data. It also matches to what inputs already passes the pydantic checks in the the data config
- Refactor the `select_time_slice_nwp()` to make it easier to follow
- Add an additional validation check in the data config to make sure the dropout probability list (if configured with list instead of float) is the same length as the dropout times list
- Add a new method to sanity-check indexes requested from the dataset

## Checklist:

### DIFF
--- a/src/ocf_data_sampler/config/model.py
+++ b/src/ocf_data_sampler/config/model.py
@@ -124,12 +124,14 @@ class DropoutMixin(Base):
     @model_validator(mode="after")
     def dropout_instructions_consistent(self) -> "DropoutMixin":
         """Validator for dropout instructions."""
-        if isinstance(self.dropout_fraction, list):
-            if len(self.dropout_fraction) != len(self.dropout_timedeltas_minutes):
-                raise ValueError(
-                    "When `dropout_fraction` is a list, it must have the same length as "
-                    "`dropout_timedeltas_minutes`"
-                )
+        if (
+            isinstance(self.dropout_fraction, list)
+            and len(self.dropout_fraction) != len(self.dropout_timedeltas_minutes)
+        ):
+            raise ValueError(
+                "When `dropout_fraction` is a list, it must have the same length as "
+                "`dropout_timedeltas_minutes`"
+            )
 
         if self.dropout_fraction == 0:
             if self.dropout_timedeltas_minutes != []:
--- a/src/ocf_data_sampler/config/model.py
+++ b/src/ocf_data_sampler/config/model.py
@@ -124,6 +124,13 @@ class DropoutMixin(Base):
     @model_validator(mode="after")
     def dropout_instructions_consistent(self) -> "DropoutMixin":
         """Validator for dropout instructions."""
+        if isinstance(self.dropout_fraction, list):
+            if len(self.dropout_fraction) != len(self.dropout_timedeltas_minutes):
+                raise ValueError(
+                    "When `dropout_fraction` is a list, it must have the same length as "
+                    "`dropout_timedeltas_minutes`"
+                )
+
         if self.dropout_fraction == 0:
             if self.dropout_timedeltas_minutes != []:
                 raise ValueError("To use dropout timedeltas dropout fraction should be > 0")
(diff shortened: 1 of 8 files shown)


src/ocf_data_sampler/config/model.py
```diff
@@ -124,6 +124,15 @@ def dropout_fractions(cls, dropout_frac: float | list[float]) -> float | list[fl
     @model_validator(mode="after")
     def dropout_instructions_consistent(self) -> "DropoutMixin":
         """Validator for dropout instructions."""
+        if (
+            isinstance(self.dropout_fraction, list)
+            and len(self.dropout_fraction) != len(self.dropout_timedeltas_minutes)
```
dfulu, line 129: We didn't have this check before that these linked-lists were the same length

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -247,6 +247,21 @@ def __init__(
         self.clip_min_dict = clip_min_dict
         self.clip_max_dict = clip_max_dict
 
+    def _sanitise_index(self, idx: int) -> int:
```
dfulu, line 250: This replaces a check lower down and allows us to use negative indexing on the dataset. e.g. `pvnet_dataset[-1]` for the last item. Also previously we were raising a `ValueError` where an `IndexError` would be more appropriate

src/ocf_data_sampler/select/time_periods.py
```diff
@@ -167,12 +167,17 @@ def find_contiguous_t0_periods_nwp(
         raise ValueError("No init-times to use")
 
     if max_dropout < ZERO_TDELTA:
-        raise ValueError("The max dropout must be positive")
+        raise ValueError("The max dropout must be non-negative (zero or positive)")
```
dfulu, line 170: This was inaccurate

src/ocf_data_sampler/select/time_periods.py
```diff
@@ -167,12 +167,17 @@ def find_contiguous_t0_periods_nwp(
...
 
     if max_staleness is not None:
 
         if max_staleness < ZERO_TDELTA:
-            raise ValueError("The max staleness must be positive")
+            raise ValueError("The max staleness must be non-negative (zero or positive)")
+
+        if max_dropout > max_staleness:
```
dfulu, line 177: This requirement was stated in the docstrings but was never verified in the code

src/ocf_data_sampler/select/time_slice.py
```diff
@@ -47,8 +47,8 @@ def select_time_slice_nwp(
     interval_start: np.timedelta64,
     interval_end: np.timedelta64,
     time_resolution: np.timedelta64,
-    dropout_timedeltas: NDArray[np.timedelta64] | None = None,
-    dropout_frac: float = 0,
+    dropout_timedeltas: NDArray[np.timedelta64] | None,
+    dropout_frac: float | list[float],
```
dfulu, line 51: We allow the NWP to have a list of dropout probabilities to match to the dropout configuration used in the other data sources

src/ocf_data_sampler/select/time_slice.py
```diff
@@ -108,3 +95,48 @@ def select_time_slice_nwp(
     selected_step_indices = get_indices_in_sorted_unique(all_steps, required_steps)
 
     return da.isel(init_time_utc=selected_init_time_index, step=selected_step_indices)
+
+
+def _get_nwp_dropout_available_time(
```
dfulu, line 100: This helper inherits and extends on all the input checking that was being done in the main `select_time_slice_nwp()` function and keeps that function cleaner

tests/select/test_time_periods.py
```diff
@@ -94,7 +94,7 @@ def test_find_contiguous_t0_periods_nwp():
         .values
     )
 
-    first_forecast_step = np.timedelta64(0)
+    first_forecast_step = np.timedelta64(0, "h")
```
dfulu, line 97: There was a deprecation warning for not supplying units on these

peterdudfield (approved):

PR #427, opened 2026-07-22
Remove old scripts and refactor t0 valid period calculations

dfulu:
# Pull Request

## Description

In this PR
- Old scripts that we haven't used in a long time are removed
  - The download gsp location script has an equivalent script in [solar-consumer](https://github.com/openclimatefix/solar-consumer/blob/main/scripts/get_pvlive_uk_pv_generation.py) and it was only in this repo since we used to store the GSP locations in this repo. The locations don't live here anymore
  - The refactor gsp or site script is very old, and isn't being used anymore
- Refactor and improve the code which finds the valid t0 periods
   - Previously this couldn't be used for the case where we wanted a single timestep from a source, like a single satellite image. This is now supported
   - Previously the filtering would incorrectly remove periods which were exactly long enough to make a sample. We now keep these 
   - I've generally cleaned up, refactored and made it more efficient
- Allow the `PVNetConcurrentDataset` to set the `use_xarray` and `include_extra_metadata` flags of its parent class
- Add explicit error that `PVNetConcurrentDataset` doesn't support using incomplete generation. This was know before but not documented anywhere
- Add a `__repr__` to the location object and use it to make nicer error messages

## Checklist:


src/ocf_data_sampler/datasets/pvnet/valid_t0s.py
```diff
@@ -63,7 +62,10 @@ def find_valid_time_periods(
                 max_staleness=max_staleness,
             )
 
-            contiguous_time_periods[f"nwp_{nwp_key}"] = time_periods
+            if len(time_periods) == 0:
```
dfulu, line 65: I've added these explicit errors to make it easier to track down when one of the datasets has gone wrong and has no contiguous periods

src/ocf_data_sampler/select/spatial_slice.py
```diff
@@ -65,6 +67,7 @@ def _get_window_bounds(
 def _validate_window_slice(
     window_slice: tuple[int, int, int, int],
     total_size: tuple[int, int],
+    locations: Location | list[Location] | None = None,
```
dfulu, line 70: Add this in to make any spatial selection error messages easier to trace

src/ocf_data_sampler/select/time_periods.py
dfulu, line 1: I refactored a lot of this code to make it more efficient and cleaner. I've kept the test-cases the same in the test directory, and added a few more

src/ocf_data_sampler/spatial.py
```diff
@@ -245,6 +245,16 @@ def __init__(self, x: float, y: float, coord_system: str, id: int | str | None =
         self._projections: dict[str, tuple[float, float]] = {coord_system: (x, y)}
         self.id = id
 
+    def __repr__(self) -> str:
```
dfulu, line 248: this makes the error messages which show the location richer and easier to debug from

tests/select/test_time_periods.py
```diff
@@ -55,6 +55,62 @@ def test_find_contiguous_t0_periods():
 
     assert periods.equals(expected_results)
 
+    # This is a stand in for where we just need a single satellite image from 5 minutes ago
```
dfulu, line 58: A test for the newly supported behaviour of just needing a single image

tests/select/test_time_periods.py
```diff
@@ -55,6 +55,62 @@ def test_find_contiguous_t0_periods():
...
+    expected_results = construct_time_periods_df(
+        start_dt=["2023-01-01 12:05", "2023-01-01 12:25"],
+        end_dt=["2023-01-01 12:15", "2023-01-01 12:25"],
+    )
+
+    assert periods.equals(expected_results)
+
+
+
+def test_find_contiguous_t0_periods_keeps_exact_length_period():
```
dfulu, line 88: A test that would have caught the bug of filtering out periods which are the exact right length

src/ocf_data_sampler/select/time_periods.py
```diff
@@ -259,46 +213,49 @@ def intersection_of_2_dataframes_of_periods(a: pd.DataFrame, b: pd.DataFrame) ->
...
     if len(a) > len(b):
         a, b = b, a
 
-    all_intersecting_periods = []
-    for a_period in a.itertuples():
-        # Five ways in which two periods may overlap:
-        # a: |----| or |---|   or  |---| or   |--|   or |-|
-        # b:  |--|       |---|   |---|      |------|    |-|
-        # In all five, `a` must always start before (or equal to) where `b` ends,
-        # and `a` must always end after (or equal to) where `b` starts.
```
peterdudfield, line 279: I quite liked this diagram, I wonder if you could add something like this back in to help explain some of the logic below

peterdudfield (approved):
Ive left one optional comment

Issue #256, opened 2025-05-12
Create config input for date/time features


Sukh-P:
As was done for the solar position inputs (shown [here](https://github.com/openclimatefix/ocf-data-sampler/commit/80f3f2e1fc09bc9857a85402168746ea578adda0)) we want more flexibility over these inputs and to control the length/resolution of these inputs through config

Issue #60, opened 2024-09-27
move input data up a level in config


AUdaltsova:
input data is currently a very small mixin in config that plugs into Configuration (also very small); input sources should just be in Configuration instead

MAYANK12SHARMA:
Hi @AUdaltsova, I would like to work on this. Can you provide me more input to work on this issue?

AUdaltsova:
Hi @MAYANK12SHARMA, sure! This is regarding our [config model](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/config/model.py), where the body of it splits into `general` and `input_data` parts. `input_data` then houses all the input sources. It might be a good idea to put the data sources directly into the config (so it would now have parts like `general`, `nwp`, `satellite`...) instead. 

This will require changes to the config model and then everywhere in the code where we use `config.input_data` or similar to be updated as well. 

Now that I think about it, it's probably a good idea to double-check there are no functions relying on cycling through the input sources (with something like `for source in config.input_data`), but that would require some digging and understanding of our code, which disqualifies this issue from the contributions-welcome tag unfortunately. I can still assign you if you want to do this, but please be aware that this will require some attention and might take you quite a bit of reading to familiarise yourself with the repo!

MAYANK12SHARMA:
## Configuration Model Refactor Proposal

Hi @AUdaltsova, I understand the issue! 

Currently, our `config_model` is structured like this:  
```python
class InputData(Base):
    """Input data model."""
    satellite: Satellite | None = None
    nwp: MultiNWP | None = None
    gsp: GSP | None = None
    site: Site | None = None

class Configuration(Base):
    """Configuration model for the dataset."""
    general: General = General()
    input_data: InputData = InputData()
```

The goal is to update it to something like this:  
```python
class Configuration(BaseModel):
    general: GeneralConfig
    nwp: NWPConfig
    satellite: SatelliteConfig
    gsp: GSPConfig
    site: SiteConfig
```

This change will affect the config model and all parts of the code that reference `config.input_data`. I’ll also double-check if any functions loop through input sources (e.g., `for source in config.input_data`) to make sure nothing breaks during the refactor.

I’d be happy to take this on — please assign me the issue! If I run into any challenges, I’ll reach out for help.

AUdaltsova:
@MAYANK12SHARMA ok great, I've assigned you now! Thanks for volunteering, let me know if you have any questions.

[...]

PR #428, opened 2026-07-29
Detangle generation data

dfulu:
# Pull Request

## Description

Currently when we run models in production we HAVE to provide generation data, even if the model does not use generation data as an input. We've tangled too much additional responsibility into the generation data like also providing the coordinates and t0-spacing.

This PR detangles the generation data inputs from the rest of the code.

The main changes are:
- Change the generation data schema. We no longer expect longitudes and latitudes in the generation data. 
- The location ID to coordinate mapping now lives in a separate metadata file which we need to point to from the config
- We split the generation data into "input" and "target" components. At inference time we can null out the target part of the config, and if a model does not use generation "input" then we don't need generation data at all.
- Various updates to processing and slicing to accommodate the 2 pieces of generation data which might be used.
- We used to tie the datetime encodings to the time interval used for the generation data. This was messy and now that we're splitting the generation in two pieces and not using it at inference, this had to be untied. So I've added a new item to configure the datetimes passed into the datetime encoding function. This is similar to what we do for the solar coords. Incidentally completes #256
- We used to tie the t0-spacing to the generation data intervals in the config. This was messy, and we can't do this if we plan to have no generation. I've moved the value used to define the t0 spacing (and also the path to the locations metadata) to a new config entry `sampling_grid`.
- The `apply_history_dropout()` was designed specifically for the old generation data slice so that the future generation wasn't dropped out but the last data was. We no longer need this since we have split the generation data already. I have modified it into `apply_dropout()` so that all values after the randomly sampled dropout time are filled with NaNs. No dropout is allowed for the target generation data.
- Previously we were using the capacity at the start of the sliced generation data interval to normalise the data. This was just a simplification. I've changed this to use the capacities at each point in the interval 

Other changes:
- Unnest the configuration. The top level of the configuration used to have `general` and `input_data` sections. We have never used the `general` section. So I've moved this top level and now the top level is everything which was previously under `input_data`. I also renamed the config class to `PVNetDataConfig` to allow us naming room for any future datasets. Completes #60
- To make things cleaner for the main changes above I did some refactoring in a few places.

## Checklist:

### DIFF
--- a/src/ocf_data_sampler/config/model.py
+++ b/src/ocf_data_sampler/config/model.py
@@ -1,8 +1,4 @@
-"""Configuration model for the dataset.
-
-Absolute or relative zarr filepath(s).
-Prefix with a protocol like s3:// to read from alternative filesystems.
-"""
+"""Configuration model for the PVNet dataset."""
 
 from collections.abc import Iterator
 from typing import Literal
@@ -18,16 +14,6 @@ class Base(BaseModel):
     """Pydantic Base model where no extras can be added."""
 
     model_config = ConfigDict(extra="forbid")
-
-
-class General(Base):
-    """General pydantic model."""
-
-    name: str = Field("example", description="The name of this configuration file")
-    description: str = Field(
-        "example configuration",
-        description="Description of this configuration file",
-    )
 
 
 class TimeWindowMixin(Base):
@@ -72,7 +58,16 @@ class TimeWindowMixin(Base):
         return self
 
 
-class DropoutMixin(Base):
+class FillValueMixin(Base):
+    """Mixin class, to add a value used for filling missing data."""
+
+    dropout_fill_value: float = Field(
+        default=0.0,
+        description="The value used to fill in dropped out data or any missing values."
+    )
+
+
+class DropoutMixin(FillValueMixin):
     """Mixin class, to add dropout minutes."""
 
     dropout_timedeltas_minutes: list[int] = Field(
@@ -85,11 +80,6 @@ class DropoutMixin(Base):
         default=0.0,
         description="Either a float(Chance of dropout being applied to each sample) or a list of "
         "floats (probability that dropout of the corresponding timedelta is applied)",
-    )
-
-    dropout_fill_value: float = Field(
-        default=0.0,
-        description="The value used to fill in dropped out data or any missing values."
     )
 
     @field_validator("dropout_timedeltas_minutes")
@@ -235,7 +225,7 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
 
     provider: str = Field(..., description="The provider of the NWP data")
 
-    accum_channels: list[str] = Field([], description="the nwp channels which need to be diffed")
+    accum_channels: list[str] = Field([], description="The NWP channels which need to be diffed")
 
     max_staleness_minutes: int | None = Field(
         None,
@@ -319,8 +309,58 @@ class MultiNWP(RootModel):
         return self.root.items()
 
 
-class Generation(TimeWindowMixin, DropoutMixin):
-    """Generation configuration model."""
+class GenerationWindow(Base):
+    """Mixin class, to add interval start and end minutes for a generation window.
+
+    Unlike `TimeWindowMixin`, the temporal resolution is not included here - it belongs to the
+    shared generation data source (`Generation.time_resolution_minutes`), not to an individual
+    window over it.
+    """
+
+    interval_start_minutes: int = Field(
+        ...,
+        description="Data interval starts at `t0 + interval_start_minutes`",
+    )
+
+    interval_end_minutes: int = Field(
+        ...,
+        description="Data interval ends at `t0 + interval_end_minutes`",
+    )
+
+    @model_validator(mode="after")
+    def validate_interval_order(self) -> "GenerationWindow":
+        """Validator for time interval fields."""
+        start = self.interval_start_minutes
+        end = self.interval_end_minutes
+        if start > end:
+            raise ValueError(
+                f"interval_start_minutes ({start}) must be <= interval_end_minutes ({end})",
+            )
+        return self
+
+
+class GenerationInputWindow(GenerationWindow, DropoutMixin):
+    """Generation input window configuration model, used for `Generation.input`.
+
+    Extends `GenerationWindow` with dropout configuration, since only the input window (not the
+    prediction target) should ever be randomly masked out.
+    """
+
+
+class GenerationTargetWindow(GenerationWindow, FillValueMixin):
+    """Generation target window configuration model, used for `Generation.target`."""
+
+
+class Generation(Base):
+    """Generation configuration model.
+
+    Bundles the shared generation data source (`zarr_path`, `time_resolution_minutes`) with its
+    `input` and `target` windows - two independently configurable time windows over the same
+    underlying data. `time_resolution_minutes` describes generation's own native data cadence
+    (used for gap detection and windowed slicing of generation's own data) - it is independent
+    of `SamplingGrid.t0_resolution_minutes`, which is the cadence t0 candidates are enumerated
+    at and may legitimately differ (e.g. generation stored every 5 minutes, sampled every 30).
+    """
 
     zarr_path: str = Field(
         ...,
@@ -328,9 +368,67 @@ class Generation(TimeWindowMixin, DropoutMixin):
         "to read from alternative filesystems.",
     )
 
+    time_resolution_minutes: int = Field(
+        ...,
+        gt=0,
+        description="The temporal resolution of the generation data in minutes",
+    )
+
+    input: GenerationInputWindow | None = None
+    target: GenerationTargetWindow | None = None
+
+    @model_validator(mode="after")
+    def validate_windows(self) -> "Generation":
+        """Validate the input/target windows are set and divisible by the shared resolution."""
+        if self.input is None and self.target is None:
+            raise ValueError(
+                "At least one of `generation.input` or `generation.target` must be configured",
+            )
+
+        for name, window in (("input", self.input), ("target", self.target)):
+            if window is None:
+                continue
+            for bound_name, bound in (
+                ("interval_start_minutes", window.interval_start_minutes),
+                ("interval_end_minutes", window.interval_end_minutes),
+            ):
+                if bound % self.time_resolution_minutes != 0:
+                    raise ValueError(
+                        f"generation.{name}.{bound_name} ({bound}) must be divisible by "
+                        f"generation.time_resolution_minutes ({self.time_resolution_minutes})",
+                    )
+        return self
+
+
+class SamplingGrid(Base):
+    """Configuration for the (location, time) grid that t0 times are sampled from.
+
+    `locations_zarr_path` points to the locations metadata (location IDs and their
+    coordinates) - see `ocf_data_sampler.load.locations.open_locations`.
+    `t0_resolution_minutes` is the cadence t0 candidates are enumerated at, needed to compute
+    valid t0 times regardless of which other input sources are configured - it is not any one
+    source's own native data resolution (see `Generation.time_resolution_minutes` for that).
+    """
+
+    locations_zarr_path: str = Field(
+        ...,
+        description="Absolute or relative zarr filepath to the locations metadata. Prefix with "
+        "a protocol like s3:// to read from alternative filesystems.",
+    )
+
+    t0_resolution_minutes: int = Field(
+        ...,
+        gt=0,
+        description="The resolution of the t0 sampling grid, in minutes.",
+    )
+
 
 class SolarPosition(TimeWindowMixin):
     """Solar position configuration model."""
+
+
+class DatetimeEncoding(TimeWindowMixin):
+    """Datetime encoding configuration model."""
 
 
 _embedding_type = list[tuple[str, Literal["cyclic", "linear"]]]
@@ -374,18 +472,13 @@ class T0Embedding(Base):
         return embeddings
 
 
-class InputData(Base):
-    """Input data model."""
+class PVNetDataConfig(Base):
+    """Configuration model for the PVNet dataset."""
 
+    sampling_grid: SamplingGrid
     satellite: Satellite | None = None
     nwp: MultiNWP | None = None
     generation: Generation | None = None
     solar_position: SolarPosition | None = None
+    datetime_encoding: DatetimeEncoding | None = None
     t0_embedding: T0Embedding | None = None
-
-
-class Configuration(Base):
-    """Configuration model for the dataset."""
-
-    general: General = General()
-    input_data: InputData = InputData()
--- a/src/ocf_data_sampler/config/model.py
+++ b/src/ocf_data_sampler/config/model.py
@@ -394,11 +394,16 @@ class Generation(Base):
 class SamplingGrid(Base):
     """Configuration for the (location, time) grid that t0 times are sampled from."""
 
-    locations_zarr_path: str = Field(
+    locations_csv_path: str = Field(
         ...,
-        description="Absolute or relative zarr filepath to the locations metadata (location IDs "
-        "and their coordinates) - see `ocf_data_sampler.load.locations.open_locations`. Prefix "
-        "with a protocol like s3:// to read from alternative filesystems.",
+        description="Absolute or relative CSV filepath to the locations metadata (location IDs "
+        "and their coordinates) - see `ocf_data_sampler.load.locations.open_locations`.",
+    )
+
+    exclude_location_ids: list[int] = Field(
+        default=[],
+        description="Location IDs from the locations metadata to drop from the sampling grid. "
+        "Every ID listed must be present in the locations data.",
     )
 
     t0_resolution_minutes: int = Field(
@@ -407,6 +412,14 @@ class SamplingGrid(Base):
         description="The cadence t0 candidates are enumerated at, needed to compute valid t0 "
         "times regardless of which other input sources are configured.",
     )
+
+    @field_validator("exclude_location_ids")
+    def validate_exclude_location_ids_unique(cls, v: list[int]) -> list[int]:
+        """Validate 'exclude_location_ids'."""
+        duplicates = {i for i in v if v.count(i) > 1}
+        if duplicates:
+            raise ValueError(f"exclude_location_ids contains duplicates: {sorted(duplicates)}")
+        return v
 
 
 class SolarPosition(TimeWindowMixin):
--- a/src/ocf_data_sampler/config/model.py
+++ b/src/ocf_data_sampler/config/model.py
@@ -310,11 +310,10 @@ class MultiNWP(RootModel):
 
 
 class GenerationWindow(Base):
-    """Mixin class, to add interval start and end minutes for a generation window.
+    """Interval bounds for a generation window.
 
-    Unlike `TimeWindowMixin`, the temporal resolution is not included here - it belongs to the
-    shared generation data source (`Generation.time_resolution_minutes`), not to an individual
-    window over it.
+    Unlike `TimeWindowMixin`, no temporal resolution here - it belongs to the shared
+    generation data source (`Generation.time_resolution_minutes`), not to an individual window.
     """
 
     interval_start_minutes: int = Field(
@@ -340,27 +339,19 @@ class GenerationWindow(Base):
 
 
 class GenerationInputWindow(GenerationWindow, DropoutMixin):
-    """Generation input window configuration model, used for `Generation.input`.
+    """A generation window with dropout configuration.
 
-    Extends `GenerationWindow` with dropout configuration, since only the input window (not the
-    prediction target) should ever be randomly masked out.
+    Dropout is configurable here since only the input window (not the prediction target)
+    should ever be randomly masked out.
     """
 
 
 class GenerationTargetWindow(GenerationWindow, FillValueMixin):
-    """Generation target window configuration model, used for `Generation.target`."""
+    """Generation target window configuration model."""
 
 
 class Generation(Base):
-    """Generation configuration model.
-
-    Bundles the shared generation data source (`zarr_path`, `time_resolution_minutes`) with its
-    `input` and `target` windows - two independently configurable time windows over the same
-    underlying data. `time_resolution_minutes` describes generation's own native data cadence
-    (used for gap detection and windowed slicing of generation's own data) - it is independent
-    of `SamplingGrid.t0_resolution_minutes`, which is the cadence t0 candidates are enumerated
-    at and may legitimately differ (e.g. generation stored every 5 minutes, sampled every 30).
-    """
+    """Generation configuration model."""
 
     zarr_path: str = Field(
         ...,
@@ -401,25 +392,20 @@ class Generation(Base):
 
 
 class SamplingGrid(Base):
-    """Configuration for the (location, time) grid that t0 times are sampled from.
-
-    `locations_zarr_path` points to the locations metadata (location IDs and their
-    coordinates) - see `ocf_data_sampler.load.locations.open_locations`.
-    `t0_resolution_minutes` is the cadence t0 candidates are enumerated at, needed to compute
-    valid t0 times regardless of which other input sources are configured - it is not any one
-    source's own native data resolution (see `Generation.time_resolution_minutes` for that).
-    """
+    """Configuration for the (location, time) grid that t0 times are sampled from."""
 
     locations_zarr_path: str = Field(
         ...,
-        description="Absolute or relative zarr filepath to the locations metadata. Prefix with "
-        "a protocol like s3:// to read from alternative filesystems.",
+        description="Absolute or relative zarr filepath to the locations metadata (location IDs "
+        "and their coordinates) - see `ocf_data_sampler.load.locations.open_locations`. Prefix "
+        "with a protocol like s3:// to read from alternative filesystems.",
     )
 
     t0_resolution_minutes: int = Field(
         ...,
         gt=0,
-        description="The resolution of the t0 sampling grid, in minutes.",
+        description="The cadence t0 candidates are enumerated at, needed to compute valid t0 "
+        "times regardless of which other input sources are configured.",
     )
 
 
--- a/src/ocf_data_sampler/datasets/pvnet/dataset.py
+++ b/src/ocf_data_sampler/datasets/pvnet/dataset.py
@@ -45,29 +45,36 @@ logger = logging.getLogger(__name__)
 
 
 
-def get_locations(zarr_path: str) -> list[Location]:
+def get_locations(csv_path: str, exclude_ids: list[int] | None = None) -> list[Location]:
     """Load the locations metadata and build the list of all locations.
 
     Args:
-        zarr_path: Path to the locations zarr data
+        csv_path: Path to the locations CSV data
+        exclude_ids: Location IDs to drop from the returned locations
     """
-    locations_data = open_locations(zarr_path)
+    locations_data = open_locations(csv_path)
 
-    locations = []
-    location_ids = locations_data["location_id"].values
+    if exclude_ids:
+        missing_ids = np.setdiff1d(exclude_ids, locations_data["location_id"].values)
+        if len(missing_ids) > 0:
+            raise ValueError(
+                f"Cannot exclude location IDs which are not in the locations data: {missing_ids}",
+            )
 
-    for location_id in location_ids:
-        loc_data = locations_data.sel(location_id=location_id)
-        locations.append(
-            Location(
-                x=loc_data["longitude"].values,
-                y=loc_data["latitude"].values,
-                coord_system="lon_lat",
-                id=int(location_id),
-            ),
+        locations_data = locations_data[~locations_data["location_id"].isin(exclude_ids)]
+
+        if len(locations_data) == 0:
+            raise ValueError("All location IDs in the locations data have been excluded")
+
+    return [
+        Location(
+            x=row.longitude,
+            y=row.latitude,
+            coord_system="lon_lat",
+            id=int(row.location_id),
         )
-
-    return locations
+        for row in locations_data.itertuples()
+    ]
 
 
 def xarray_to_lightarray_dict(
@@ -255,7 +262,10 @@ class AbstractPVNetDataset(PickleCacheMixin, Dataset):
 
         config = load_yaml_configuration(config_filename)
 
-        locations = get_locations(config.sampling_grid.locations_zarr_path)
+        locations = get_locations(
+            config.sampling_grid.locations_csv_path,
+            config.sampling_grid.exclude_location_ids,
+        )
 
         datasets_dict = get_dataset_dict(config)
 
--- a/src/ocf_data_sampler/datasets/pvnet/dataset.py
+++ b/src/ocf_data_sampler/datasets/pvnet/dataset.py
@@ -265,8 +265,8 @@ class AbstractPVNetDataset(PickleCacheMixin, Dataset):
             if len(missing) > 0:
                 raise ValueError(f"Generation data is missing for location IDs: {missing}")
 
-            # Restrict to the catalog's locations - generation may have extra ids that aren't
-            # real samplable points (e.g. a summation-model placeholder).
+            # Slice the generation data to only include the specified locations. This allows us to
+            # quality check the generation data for nans and find valid t0 times for each location.
             datasets_dict["generation"] = datasets_dict["generation"].sel(location_id=location_ids)
 
         # Check if generation data has nans. If generation isn't configured at all, there's no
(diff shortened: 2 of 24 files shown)


src/ocf_data_sampler/config/model.py
```diff
@@ -1,8 +1,4 @@
-"""Configuration model for the dataset.
-
-Absolute or relative zarr filepath(s).
```
dfulu, line 3: I don't know what this comment was doing here

src/ocf_data_sampler/config/model.py
```diff
@@ -72,7 +58,16 @@ def validate_intervals(self) -> "TimeWindowMixin":
         return self
 
 
-class DropoutMixin(Base):
+class FillValueMixin(Base):
```
dfulu, line 61: I separated this from the `DropoutMixin` because I thought we might want to allow missing values in the target generation to be infilled. However, I explicitly don't want to allow dropout in the target generation. We could be tighter on the target generation and assume that we should never have any NaNs in the sampled future slice. In that case this could go back into `DropoutMixin`. I wasn't sure on this one

src/ocf_data_sampler/config/model.py
```diff
@@ -319,20 +309,128 @@ def items(self) -> Iterator[tuple[str, NWP]]:
         return self.root.items()
 
 
-class Generation(TimeWindowMixin, DropoutMixin):
-    """Generation configuration model."""
+class GenerationWindow(Base):
```
dfulu, line 312: This class is used for both the input and target generation slices. The full config paths are `config.generation.[input/target]`

This class doesn't have `time_resolution_minutes` or `zarr_path` since those are assumed to be shared and they live directly under  `config.generation`

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -161,6 +162,69 @@ def add_alternate_coordinate_projections(
     return locations
 
 
+def build_numpy_sample(
```
dfulu, line 172: This is a refactor. Most of the code under here was copy-pasted from the `.process_and_combine_datasets()` method deleted below. Although a some of lines from that method were moved to a different function. This function now JUST builds the numpy sample from already processed DataArray objects

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -232,12 +308,6 @@ def __init__(
         else:
             self.datasets_dict = xarray_to_lightarray_dict(datasets_dict)
 
-        # Assign t0 idx value
-        self.t0_idx = (
```
dfulu, line 236: Now that we have input and target generation slices we don't need this anymore

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -451,10 +444,13 @@ def _get_sample(self, t0: np.datetime64, location: Location) -> NumpySample:
         sample_dict = slice_datasets_by_space(self.datasets_dict, location, self.config)
         sample_dict = slice_datasets_by_time(sample_dict, t0, self.config)
         sample_dict = load_data_dict(sample_dict)
-        # Apply dropout to the data sources in-place
-        apply_dropout_to_datasets(sample_dict, t0, self.config)
-        sample_dict = diff_nwp_data(sample_dict, self.config)
-        return self.process_and_combine_datasets(sample_dict, t0, location)
+        sample_dict = preprocess_dataset_dict(
```
dfulu, line 461: This new function holds steps which were in the `.process_and_combine_datasets()` method, which aren't in `build_numpy_sample()`.  Dropout and NWP diffing were also moved into the new `preprocess_dataset_dict()` function

src/ocf_data_sampler/datasets/pvnet/loading.py
```diff
@@ -1,54 +1,46 @@
...
+        config: PVNetDataConfig configuration object
     """
     datasets_dict = {}
 
-    # Load generation data unless the path is None
-    if input_config.generation and input_config.generation.zarr_path:
-        da_generation = open_generation(zarr_path=input_config.generation.zarr_path)
-
-        # Remove location_id 0 if more than one location present
-        if len(da_generation["location_id"]) > 1 and 0 in da_generation["location_id"].values:
```
dfulu, line 27: The generation data no longer defines the locations we sample from. So this isn't needed

src/ocf_data_sampler/datasets/pvnet/preprocess.py
```diff
@@ -172,3 +240,42 @@ def fill_nans(da: TArray, fill_value: float) -> TArray:
     if np.isnan(da.data).any():
         da.data = np.nan_to_num(da.data, copy=True, nan=fill_value)
     return da
+
+
+def preprocess_dataset_dict(
+    dataset_dict: SourceDict,
+    t0: np.datetime64,
+    config: PVNetDataConfig,
+    mean_dict: dict[str, np.ndarray | dict[str, np.ndarray]],
```
dfulu, line 257: These inputs are horrible and I'll refactor this all at some point

src/ocf_data_sampler/load/locations.py
```diff
@@ -0,0 +1,47 @@
...
+def open_locations(zarr_path: str) -> xr.Dataset:
+    """Open and eagerly load the locations metadata and validate its data types.
+
+    Args:
+        zarr_path: Path to the locations zarr data
+
+    Returns:
+        xr.Dataset: The opened locations metadata
+    """
+    ds = xr.open_zarr(zarr_path, chunks=None)
```
dfulu: This could easily be a CSV, but it would be our only one. So I think having less data formats wins here

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -262,113 +332,30 @@ def _sanitise_index(self, idx: int) -> int:
 
         return index
 
-    def process_and_combine_datasets(
```
dfulu, line 265: This method has been removed and distributed to the new `preprocess_dataset_dict` and `build_numpy_sample` functions

src/ocf_data_sampler/datasets/pvnet/slicing.py
```diff
@@ -17,29 +17,24 @@
...
     Args:
         datasets_dict: Dictionary of the input data sources
         location: The location to sample around
-        config: Configuration object.
+        config: PVNetDataConfig object.
 
     Returns:
         A dictionary of the sliced input data sources.
     """
-    if not set(datasets_dict.keys()).issubset({"nwp", "sat", "generation"}):
```
dfulu, line 32: This just isn't needed. We only get to this point via internal functions

src/ocf_data_sampler/config/model.py
```diff
@@ -374,18 +458,13 @@ def validate_embeddings(cls, embeddings: _embedding_type) -> _embedding_type:
         return embeddings
 
 
-class InputData(Base):
-    """Input data model."""
+class PVNetDataConfig(Base):
+    """Configuration model for the PVNet dataset."""
 
+    sampling_grid: SamplingGrid
```
dfulu, line 477: Unlike the other inputs, this is not optional

src/ocf_data_sampler/config/model.py
```diff
@@ -328,11 +359,64 @@ class Generation(TimeWindowMixin, DropoutMixin):
...
+    input: GenerationInputWindow | None = None
+    target: GenerationTargetWindow | None = None
+
+    @model_validator(mode="after")
+    def validate_windows(self) -> "Generation":
+        """Validate the input/target windows are set and divisible by the shared resolution."""
+        if self.input is None and self.target is None:
+            raise ValueError(
+                "At least one of `generation.input` or `generation.target` must be configured",
+            )
```
AUdaltsova, line 377: I feel like I'm missing something obvious, but I thought the point of the detanglement was that we can not supply any generation data and still run? E g at inference time, if history is not an input & we wouldn't have target data either?

src/ocf_data_sampler/config/model.py
```diff
@@ -328,11 +359,64 @@ class Generation(TimeWindowMixin, DropoutMixin):
...
+
 
 class SolarPosition(TimeWindowMixin):
     """Solar position configuration model."""
 
 
+class DatetimeEncoding(TimeWindowMixin):
+    """Datetime encoding configuration model."""
+
+
```
AUdaltsova, line 432: I thought we were getting rid of these in favour of t0 embeddings?

AUdaltsova: Also, it's bothered me for a while that solar position and now this have to be the the same as target, as in, have the same interval start, interval end, and resolution, but we define them separately. I don't know if it's worth it, or if we should make an update to PVNet to allow them to be truly decoupled, or something

tests/config/test_config.py
```diff
@@ -2,42 +2,45 @@
...
     return config, provider
 
 
 def _validate_configuration(config):
     """Recreate config instance from dict to trigger validation."""
-    return Configuration(**config.model_dump())
+    return PVNetDataConfig(**config.model_dump())
+
+
+_MINIMAL_SAMPLING_GRID = {"locations_zarr_path": "locations.zarr", "t0_resolution_minutes": 30}
```
AUdaltsova: This is the nittest pick possible, but is there a reason this is not at the top after the inputs? I see that this brings the variable closer to where it's used, but if e.g. you're looking at the `test_extra_field_error` it's not immediately obvious that this is defined one test above, and I would go look for it at the top of the file. But might be preference/new best practice I'm not aware of

src/ocf_data_sampler/config/model.py
dfulu: This only gets triggered if `config.generation` is not None. So if the `config.generation` is not None then at leats one of the windows needs to be configured

src/ocf_data_sampler/config/model.py
dfulu: I agree with you, my plan is to update PVNet to allow us to configure it there. I also am in favour of getting rid of the datetime encoding, but I wanted to leave that as a separate issue after we talk about it in an ML chat. For now I just wanted to keep it working in some way rather than me make the decision to cut them here

tests/config/test_config.py
dfulu: That is a good point. I'll move it

src/ocf_data_sampler/config/model.py
AUdaltsova: Oh yeah makes sense! Thank you

src/ocf_data_sampler/config/model.py
AUdaltsova: Yeah fair, thanks!

src/ocf_data_sampler/config/model.py
peterdudfield: I think assume (and checking) that there is no Nans in target data would be good

However in the input data we definitely should have options `dropouts` and or `fillvalue`.

`Dropout` has always confused me a bit here. I wonder if it should be renamed to `T0Delay` or `LiveDelay`. For me dropout is about randomly dropping bits of data. I can see that there is some cross over if we are assuming that with prob=1.0 we drop data out.

src/ocf_data_sampler/config/model.py
```diff
@@ -328,11 +359,64 @@ class Generation(TimeWindowMixin, DropoutMixin):
...
+            ):
+                if bound % self.time_resolution_minutes != 0:
+                    raise ValueError(
+                        f"generation.{name}.{bound_name} ({bound}) must be divisible by "
+                        f"generation.time_resolution_minutes ({self.time_resolution_minutes})",
+                    )
+        return self
+
+
+class SamplingGrid(Base):
```
peterdudfield, line 394: Would this ever be different from the Generation data?

src/ocf_data_sampler/config/model.py
```diff
@@ -374,18 +458,13 @@ def validate_embeddings(cls, embeddings: _embedding_type) -> _embedding_type:
...
+    """Configuration model for the PVNet dataset."""
 
+    sampling_grid: SamplingGrid
     satellite: Satellite | None = None
     nwp: MultiNWP | None = None
     generation: Generation | None = None
     solar_position: SolarPosition | None = None
+    datetime_encoding: DatetimeEncoding | None = None
     t0_embedding: T0Embedding | None = None
-
```
peterdudfield, line 385: I wonder as this gets a bit more complicated we should make a little diagram of the config. `pydantic-mermaid` might be able to do this? What do you think @dfulu ?

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -191,10 +255,25 @@ def __init__(
...
-        datasets_dict = get_dataset_dict(config.input_data)
+        locations = get_locations(config.sampling_grid.locations_zarr_path)
+
+        datasets_dict = get_dataset_dict(config)
 
-        # Check if generation data has nans
-        self.complete_generation = not datasets_dict["generation"].isnull().any()
+        if "generation" in datasets_dict:
+            location_ids = [loc.id for loc in locations]
+            missing = np.setdiff1d(location_ids, datasets_dict["generation"]["location_id"].values)
```
peterdudfield, line 274: are you sure you want this? Could be some cases where this is not true? Maybe if we come across them, we can adjust this line

src/ocf_data_sampler/datasets/pvnet/dataset.py
peterdudfield: That is nicer

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -378,41 +368,48 @@ def find_valid_t0_and_location_ids(
...
         datasets_without_generation = {k: v for k, v in datasets_dict.items() if k != "generation"}
         valid_time_periods = find_valid_time_periods(datasets_without_generation, config)
 
         # Loop over each location in system id and obtain valid periods
-        generations = datasets_dict["generation"]
-        location_ids = generations.location_id.values
-        generation_config = config.input_data.generation
+        generation_windows = [
+            w for w in (config.generation.input, config.generation.target) if w is not None
+        ]
```
peterdudfield, line 391: Im a bit confused by this, what is this doing?

src/ocf_data_sampler/datasets/pvnet/dataset.py
peterdudfield: nice

src/ocf_data_sampler/config/model.py
```diff
@@ -374,18 +458,13 @@ def validate_embeddings(cls, embeddings: _embedding_type) -> _embedding_type:
         return embeddings
 
 
-class InputData(Base):
-    """Input data model."""
+class PVNetDataConfig(Base):
```
peterdudfield, line 474: Are you sure you want to call it `PVNetDataConfig`. My vote would probably just call it `Configuration` as its got nothing to do with PVNet, just we happen to use it in PVNet. I appreciate that this is the only thing we use it for at the moment

src/ocf_data_sampler/config/model.py
```diff
@@ -328,11 +359,64 @@ class Generation(TimeWindowMixin, DropoutMixin):
...
+
+class SamplingGrid(Base):
+    """Configuration for the (location, time) grid that t0 times are sampled from."""
+
+    locations_zarr_path: str = Field(
+        ...,
+        description="Absolute or relative zarr filepath to the locations metadata (location IDs "
+        "and their coordinates) - see `ocf_data_sampler.load.locations.open_locations`. Prefix "
+        "with a protocol like s3:// to read from alternative filesystems.",
+    )
```
peterdudfield, line 407: How come we dont use the generation zarr file here?

tests/config/test_config.py
```diff
@@ -165,15 +168,41 @@ def test_accum_channels_validation(config_filename):
...
+        rf"nwp.{nwp_name}\n"
         fr"  Value error, NWP provider '{nwp_name}': all values in 'accum_channels' "
         r"should be present in 'channels'\. "
         r"Extra values found: {'invalid_channel'}.*"
     )
     with pytest.raises(ValidationError, match=expected_error):
         _validate_configuration(invalid_config)
+
+
+def test_generation_interval_divisibility_raises(config_filename):
```
peterdudfield, line 184: nice

tests/datasets/pvnet/test_dataset.py
```diff
@@ -217,6 +238,29 @@ def test_solar_position_decoupling(tmp_path, pvnet_config_filename):
...
+    assert dataset.complete_generation
+
+    # All locations from the locations catalog are available - none to filter out
+    assert len(dataset.locations) == 317
+
+    sample = dataset[0]
+    assert "generation_input" not in sample
+    assert "generation_target" not in sample
+    assert "nwp_ukv" in sample
+    assert "satellite" in sample
```
peterdudfield, line 287: good test, and I assume this is how we would run in live

peterdudfield (approved):
Looks great ive left a few comments but nothing big

Do you want to add a migrations script in this PR too? Or perhaps that will come along a bit later

src/ocf_data_sampler/config/model.py
dfulu: We have example configs which for me seem more helpful than a diagram

src/ocf_data_sampler/datasets/pvnet/dataset.py
dfulu: I think its important to have this in. Otherwise you might think you're training on a certain set of locations (as defined in the locations file) but some or many of them were missing. So when it comes to inference time you end up showing your model locations it has never seen during training

src/ocf_data_sampler/datasets/pvnet/dataset.py
dfulu: Its just making a list of the generation windows which are requested. It might include an input generation window, and/or a target generation window or netiher

src/ocf_data_sampler/config/model.py
dfulu: Yeh I think I want to call it that. It creates sampled with the input format tailored to PVNet. We've never used it for any other model. I also want to leave naming space so that some day I might have a cloudcasting config in here too so we can use components of this library for that

src/ocf_data_sampler/config/model.py
dfulu: As in why don't we use the generation zarr to define the locations? If so then the point of this PR specifically to remove that dependency. We don't want to require generation data to be constructed just to specify the locations

tests/datasets/pvnet/test_dataset.py
dfulu: Yeh exactly

dfulu:
> Looks great ive left a few comments but nothing big
> 
> Do you want to add a migrations script in this PR too? Or perhaps that will come along a bit later

Typically we put migration scripts in PVNet when we upgrade it to newer versions of ocf-data-sampler. I think we should do the same. Plus we should keep an eye on how many breaking changes we're introducing. I think we should consider a clean break

src/ocf_data_sampler/config/model.py
dfulu: We already do check the generation data for NaNs and only choose t0 times which don't slice out a NaN. I don't think I want to change the dropout name here. There's a related issue to rename the NWP dropout #18 and any changes would sit better there

src/ocf_data_sampler/config/model.py
dfulu: Do you mean would the time resolution of t0 values ever be different from the time resolution of the generation data? If so, then yes it could. For one where we don't use any generation data - like doing inference for a model which doesn't use generation inputs. And two for the 5-minutely updates in PVNet

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -46,21 +45,23 @@
 
 
 
-def get_locations(generation_data: xr.DataArray) -> list[Location]:
-    """Get list of locations of all locations.
+def get_locations(zarr_path: str) -> list[Location]:
```
AUdaltsova: I wonder if this should live in `loading.py`? It feels like it's doing a very similar sort of prep-work to constructing the datasets dict.

dfulu: I definitely agree. I've got another PR in mind for later in which I'll move a bunch of the functions out of this file and into its sibling files. Since this function is already in here I didn't want to move it in this PR

src/ocf_data_sampler/datasets/pvnet/loading.py
AUdaltsova: Sorry I'm confused a bit here, has this been moved somewhere I can't see? Or are we assuming that you've removed your national data from your files beforehand?

dfulu: My assumption was that we shouldn't have national in our `locations.zarr`. There is a line where we filter the generation data to just the location IDs in the `locations.zarr`. 

If we want this filter I guess we could put it in `open_locations`? We could even make that configurable from the config?

dfulu: I just don't like this if-else statement which filters ID 0 if there is more than 1 location. It feels brittle to me, like its set up to catch you out

src/ocf_data_sampler/datasets/pvnet/dataset.py
AUdaltsova: Yeah fair! Cool beans, & definitely agree that a lot of helpers in this file need to move out. Thank you!

src/ocf_data_sampler/datasets/pvnet/loading.py
AUdaltsova: Yeah I agree about if/else! But it also feels strange to me to basically have something like `NL_all_locations.zarr` and then `NL_locations_no_national.zarr` and juggle that (unless I misunderstand the idea here?) Config-configurable sounds good! I initially thought it should let you filter out any location, but I think that would be bad (more complicated traceability of what the model was trained on), so maybe like a binary flag `include_national_location`? Is that along the lines of what you were thinking?

dfulu: I'd be tempted to make it more flexible than `include_national_location`. I think I'd have something like `exclude_ids: list[int]` which would allow us to exclude whatever we want. The benefit of that is if we ever changed the convention that ID 0 is national, then we'd not need any code changes. And similar if we updated to use string IDs instead of integers, we'd just have to update the type hint. 

It might make it more complex to see what a model trained on, but I think we can get around that. Though now I'm actually thinking that a CSV might be the way to go for the locations instead of zarr. Its probably easier if the locations are human readable and editable by hand

AUdaltsova: Yeah fair. And agree about the csv

src/ocf_data_sampler/datasets/pvnet/valid_t0s.py
```diff
@@ -83,19 +81,21 @@ def find_valid_time_periods(
...
+                datasets_dict["generation"]["time_utc"].values,
+                time_resolution=minutes(config.generation.time_resolution_minutes),
+                interval_start=minutes(window_config.interval_start_minutes),
+                interval_end=minutes(window_config.interval_end_minutes),
+            )
 
-        if len(time_periods) == 0:
-            raise ValueError("No valid t0 periods found for generation data")
+            if len(time_periods) == 0:
+                raise ValueError("No valid t0 periods found for generation data")
```
AUdaltsova: This is also a nitpick, but I'd maybe prefer if this said `{key} data` instead of `generation data`, so if you have both you know which one to fix.

src/ocf_data_sampler/config/model.py
```diff
@@ -319,7 +309,48 @@ def items(self) -> Iterator[tuple[str, NWP]]:
...
+        start = self.interval_start_minutes
+        end = self.interval_end_minutes
+        if start > end:
+            raise ValueError(
+                f"interval_start_minutes ({start}) must be <= interval_end_minutes ({end})",
+            )
+        return self
+
+
+class GenerationInputWindow(GenerationWindow, DropoutMixin):
```
AUdaltsova, line 341: Only just realised that this is called "input". I don't have strong feelings about this if you think this is clearer, but I think we usually call this "history". I'm happy either way, but I think if this is "input" we might want to rename the parameter in pvnet to `include_generation_as_an_input` or some such instead of `include_generation_history` - just for consistency

src/ocf_data_sampler/datasets/pvnet/valid_t0s.py
dfulu: Sure, added

src/ocf_data_sampler/config/model.py
dfulu: Yeh input and target seemed like cleaner names to me. I think I mentioned to you last week that I'd like to experiment with PVNet to predict a few past generation timestamps as well as the future timestamps to see if it helps training the satellite encoder. I think input and target are cleaner names which cover that kind of experiment as well as our regular configuration.

And now the input and target period are freely configurable (if we wanted we could configure the input to be after the target period) so I think input and target are better general names than history and future.

And yes I agree that we should change that in PVNet

src/ocf_data_sampler/load/locations.py
dfulu: I actually changed this to a CSV after this comment. I think its better for the locations metadata to be human readable and more easily editable, and I think that beats having a single data format.

tests/conftest.py
```diff
@@ -192,23 +199,46 @@ def nwp_cloudcasting_zarr_path(session_tmp_path, session_rng):
...
+    return _locations_dataframe(session_rng, np.arange(1, 11))
+
+
+@pytest.fixture(scope="session")
+def ds_generation(session_rng, df_locations):
+    """Generation for every catalogued location, plus ID 0.
+
+    ID 0 is a summation-model placeholder which the catalog deliberately does not list, so this
+    exercises the filtering of generation IDs down to the catalog's locations.
+    """
```
AUdaltsova: I think this is not true now? Should it have 0 in the catalogue and then exclude through config?

tests/datasets/pvnet/test_dataset.py
```diff
@@ -114,6 +136,24 @@ def test_pvnet_dataset(pvnet_config_filename):
     _pvnet_dataset_sample_check(sample, dataset.config)
 
 
+def test_get_locations_exclude_ids(locations_csv_path):
+    excluded_ids = [1, 5, 317]
+    locations = get_locations(locations_csv_path, exclude_ids=excluded_ids)
+
+    assert len(locations) == 317 - len(excluded_ids)
+    assert not set(excluded_ids) & {loc.id for loc in locations}
```
AUdaltsova, line 150: Nice, thanks for adding test coverage!

tests/conftest.py
dfulu: Oh that is true, I missed this part! Thanks for catching this!

src/ocf_data_sampler/select/dropout.py
```diff
@@ -5,15 +5,16 @@
...
     t0: np.datetime64,
     dropout_timedeltas: list[np.timedelta64],
     dropout_frac: float | list[float],
 ) -> TArray:
-    """Apply in-place random dropout to the historical part of some sequence data.
+    """Apply in-place random dropout to some sequence data.
 
-    Dropped out data is replaced with NaNs.
+    A timedelta relative to t0 is randomly sampled, and all data after that point in time is
+    dropped out (replaced with NaNs).
```
AUdaltsova, line 17: I've only now realised that this makes dropout relative to t0, which is a bit counter-intuitive to me: I would've assumed dropout is applied from the interval end. Is that by design?

dfulu: Dropout has actually always been relative to t0, and it is [documented to be that way](https://github.com/openclimatefix/ocf-data-sampler/blob/main/ocf_data_sampler/config/model.py#L86-L90) even on the main branch

dfulu: I can see your point in doing it relative to the end of the period. But one thing that's nice about the way we do it now is that if I wanted to train a new model which and change the amount of satellite data it uses, then all I need to do is update the `interval_end_minutes` and I don't need to update the dropout values. To me that seems easier

tests/conftest.py
dfulu: I've updated and also did some clean up in the tests to avoid the numbers 317/318 (for the number of GSPs) being written everywhere

src/ocf_data_sampler/select/dropout.py
AUdaltsova: Ok that makes sense! Thank you for explaining :)

tests/conftest.py
AUdaltsova: Nice, thank you!

src/ocf_data_sampler/datasets/pvnet/valid_t0s.py
AUdaltsova: Thank you!

src/ocf_data_sampler/datasets/pvnet/preprocess.py
```diff
@@ -104,61 +179,62 @@ def diff_nwp_data(dataset_dict: SourceDict, config: Configuration) -> SourceDict
 def apply_dropout_to_datasets(
     datasets_dict: SourceDict,
     t0: np.datetime64,
-    config: Configuration,
+    config: PVNetDataConfig,
 ) -> None:
     """Apply dropout in-placeto the dictionary of input data sources around a given t0 time.
```
AUdaltsova:
```suggestion
    """Apply dropout in-place to the dictionary of input data sources around a given t0 time.
```
Sorry, I know it's not changed by your PR! But the docstring is slightly. Feel free to not piggy-back though

tests/datasets/pvnet/test_preprocess.py
```diff
@@ -53,27 +100,38 @@ def test_apply_dropout_to_datasets(pvnet_config_filename):
         ],
         dtype="datetime64[m]",
     )
+    generation_mw = np.arange(4 * 2, dtype=float).reshape(4, 2)
+    capacity_mwp = np.ones((4, 2))
```
AUdaltsova: I am a bit in two minds about this test; I think it duplicates a lot of things done by tests in `tests/select/test_dropout.py`, as all this function does is check for keys and call `apply_dropout`, which is already extensively tested in the dropout section. I guess it can stay to catch if someone accidentally makes the `if` statements nested or something... But I think it can maybe use  `_generation_da()` instead of re-making generation DataArray?

AUdaltsova: Okay actually, I've noticed it's not even testing satellite dropout, so it will not tell you anything useful that `apply_dropout` wouldn't I think?

src/ocf_data_sampler/load/locations.py
AUdaltsova: Yeah, agree! Thanks for changing. It is kinda funny to me that we've come full circle back to having split generation and metadata files though.

tests/datasets/pvnet/test_preprocess.py
dfulu: Yeh I see what you mean, and I _am_ a fan of less code. 

This test does catch if everything in the configs and datasets are unpacked and repacked correctly. Like did we take the right settings from the config and apply the correct dropout to each data source. The generation settings are under `config.generation.input` but the satellite setting are shallower under `config.satellite`.  But the tests in `test_dataset.py` should catch that too, so that seems like a strike against it. This additional test would only guard against us going something silly like using the satellite dropout settings into the generation.

The `apply_dropout_to_datasets()` function is advertised as an in-place operation, and this test maybe does help lock that in. It could help us catch errors if we ever made (accidentally or otherwise) `apply_dropout_to_datasets()` out-of-place.

Note: I do plan to come back and make us more consistent with what we return for in-place and out-of-place functions. Sometimes we return an object for an in-place operation and sometimes we don't.

I don't feel strongly either way. What do you think?

AUdaltsova: Okay yeah I think it does make sense to check that we've passed the right config along. I think in that case it should check that dropout is applied to satellite (currently satellite has settings that don't result in dropout; I think the result of that is ambiguous - if nothing happened it can mean `apply_dropout()` wasn't called at all), and also maybe check that nothing happens to `generation_target`? That feels like something a silly update might end up breaking. 

I think the fact that it's in-place is locked in `apply_dropout()` as that's already in-place? You can't really change that by only changing `apply_dropout_to_datasets()`. 

I also still think the `_generation_da()` from this file can probably be reused here instead of LL 105-113, but maybe I'm missing something?

Also, I do realise I am nitpicking over a test that was already kinda bad before this PR. If you think this is getting silly, go ahead and merge as is, LGTM otherwise!

dfulu: Yes it is locked in, because of `apply_dropout()`, but there is always the chance we might decide to make `apply_dropout()` out-of-place later and forget to update this function. If we did that we'd be silently not apply dropout anywhere and I think this is the only test that might catch that. But that is an real edge case, we shouldn't be so silly. 

So how about this this then. We switch to use `_generation_da()` as a helper, we add `generation_target` to the sample and check it isn't touched, we add dropout to the satellite and we merge this? I'm going to do a sweep through the tests in a later PR anyay

PR #429, opened 2026-07-31
Upgrade xarray-tensor and add tenorstore concat function

dfulu:
# Pull Request

## Description

This PR improves how we use xarray-tensorstore with multiple zarr by adding a concat function.

Currently we use a custom function which opens a bunch of zarrs using xarray and concatenates the data variables using tensorstore as a backend. We can never concatenate xarray-tensorstore objects after we have loaded them since `xarray.concat` doesn't support tensorstore backends and it tries to load the data eagerly to concat it.

The new `concat_tensorstore()` function in this PR introduces a way for us to lazily concat xarray-tensorstore objects. Now that we have this function, we no longer need the `_tensorstore_open_zarrs()` function which can only concat tensorstore objects whilst opening the zarrs. This simplifies our code and is more generally useful. 

This has the additional benefit of allowing us to remove dask as a requirement. Previously we needed dask for [one step](https://github.com/openclimatefix/ocf-data-sampler/blob/dev_feb2026_speedups/src/ocf_data_sampler/common/xr_tensorstore.py#L187-L194) in the original `_tensorstore_open_zarrs()` function. The dask step was also incredibly slow because it would build a dask graph over all the data variables. When opening 5 years of highly chunked satellite data this was taking ~25 seconds to open and concat using `open_zarr_paths()` which internally called `_tensorstore_open_zarrs()`. Now this takes under a second.

I've also taken the opportunity to upgrade us to the latest release of `xarray-tensorstore` which supports zarr3 and allows us to remove a lot of the custom code we had in the `xr_tensorstore` module.

I've also removed the Apache License and attribution from `xarray-tensorstore` since we are no longer redistributing any copied code from `xarray-tensorstore`, only importing it.

NOTE: This PR removes some functionality

[Previously](https://github.com/openclimatefix/ocf-data-sampler/blob/dev_feb2026_speedups/src/ocf_data_sampler/common/xr_tensorstore.py#L200-L207) we allowed zarrs to be concatenated even if they didn't share the same coordinates. I think this might have been motivated by the Meteosat satellite position over the indian ocean shifting. This PR removes support for this and the coordinates now must match. 

I strongly believe we should discourage ourselves from doing this kind of thing since it could affect model performance and I believe there should be other ways to solve satellite shifts better. But if absolutely required, we could put that functionality back into this new version later

## Checklist:

### DIFF
--- a/src/ocf_data_sampler/common/xr_tensorstore.py
+++ b/src/ocf_data_sampler/common/xr_tensorstore.py
@@ -88,7 +88,7 @@ def concat_tensorstore(datasets: Sequence[xr.Dataset], concat_dim: str) -> xr.Da
     #   copy for the data_vars below.
     # - join="exact" ensures that the coords are identical across datasets, which is a bakstop for
     #   the _validate() check above.
-    out = xr.concat(
+    ds_out = xr.concat(
         [ds.drop_vars(first.data_vars) for ds in datasets],
         dim=concat_dim,
         join="exact",
@@ -101,10 +101,10 @@ def concat_tensorstore(datasets: Sequence[xr.Dataset], concat_dim: str) -> xr.Da
                 [_tensorstore_of(ds[name]) for ds in datasets],
                 axis=da.dims.index(concat_dim),
             )
-            out[name] = xr.Variable(da.dims, xrt._TensorStoreAdapter(store), attrs=da.attrs)
+            ds_out[name] = xr.Variable(da.dims, xrt._TensorStoreAdapter(store), attrs=da.attrs)
         else:
-            out[name] = da.variable  # attrs travel with the Variable; xarray copies on assign
-    return out
+            ds_out[name] = da.variable  # attrs travel with the Variable; xarray copies on assign
+    return ds_out
 
 
 def open_zarr_paths(zarr_path: ZarrSource, concat_dim: str | None = None) -> xr.Dataset:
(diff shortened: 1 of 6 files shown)


src/ocf_data_sampler/common/xr_tensorstore.py
```diff
@@ -1,96 +1,110 @@
...
-
-
-def _tensorstore_open_zarrs(
-    paths: list[str],
-    data_vars: list[str],
-    concat_axes: list[int],
-    context: ts.Context,
-) -> dict[str, ts.TensorStore]:
-    """Open multiple zarrs with TensorStore.
+def _tensorstore_of(da: xr.DataArray) -> ts.TensorStore:
```
Sukh-P: I am not super sure about the name of this function, maybe something like `_as_tensorstore` would be a bit clearer? I might be being a bit pedantic here but this follows a bit for example how `numpy` does naming of functions like this e.g. https://numpy.org/doc/stable/reference/generated/numpy.asarray.html

Sukh-P (approved):
Looks great, really nice changes!

src/ocf_data_sampler/common/xr_tensorstore.py
dfulu: I see what you mean, `_tensorstore_of()` isn't great. However, to me `_as_` implies that it might be converting or copying, but its just pulling out the tensorstore. How do you feel about `_extract_tensostore()` since "extract" is the first word in the docstring?

PR #430, opened 2026-07-31
9 various tidy-ups and loud failing

dfulu:
# Pull Request

## Description

This PR wraps up a lot of small changes mainly aimed at the code failing early and raising helpful error messages

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ It includes and adapts the `ephemeris()` function from the [pvlib-python](https:
 
 ## Documentation
 
-**ocf-data-sampler** doesn't have external documentation _yet_; you can read a bit about how our torch datasets work in the README [here](ocf_data_sampler/torch_datasets/README.md).
+**ocf-data-sampler** doesn't have external documentation _yet_; you can read a bit about how our torch datasets work in the README [here](src/ocf_data_sampler/datasets/pvnet/README.md).
 
 ## FAQ
 
--- a/src/ocf_data_sampler/config/model.py
+++ b/src/ocf_data_sampler/config/model.py
@@ -63,7 +63,8 @@ class FillValueMixin(Base):
 
     dropout_fill_value: float = Field(
         default=0.0,
-        description="The value used to fill in dropped out data or any missing values."
+        description="The value used to fill in dropped out data or any missing values. This is "
+        "applied after normalisation, so it is in normalised units."
     )
 
 
--- a/src/ocf_data_sampler/config/model.py
+++ b/src/ocf_data_sampler/config/model.py
@@ -1,6 +1,6 @@
 """Configuration model for the PVNet dataset."""
 
-from collections.abc import Iterator
+from collections.abc import ItemsView, Iterator, KeysView
 from typing import Literal
 
 from pydantic import BaseModel, ConfigDict, Field, RootModel, field_validator, model_validator
@@ -281,13 +281,12 @@ class NWP(TimeWindowMixin, DropoutMixin, SpatialWindowMixin, NormalisationConsta
 
 
 class MultiNWP(RootModel):
-    """Configuration for multiple NWPs."""
+    """Configuration for multiple NWPs.
+
+    The NWP sources are accessed dict-style - `config.nwp["ukv"]`, not `config.nwp.ukv`.
+    """
 
     root: dict[str, NWP]
-
-    @override
-    def __getattr__(self, item: str) -> NWP:
-        return self.root[item]
 
     @override
     def __getitem__(self, item: str) -> NWP:
@@ -301,11 +300,11 @@ class MultiNWP(RootModel):
     def __iter__(self) -> Iterator:
         return iter(self.root)
 
-    def keys(self) -> Iterator[str]:
+    def keys(self) -> KeysView[str]:
         """Returns dictionary-like keys."""
         return self.root.keys()
 
-    def items(self) -> Iterator[tuple[str, NWP]]:
+    def items(self) -> ItemsView[str, NWP]:
         """Returns dictionary-like items."""
         return self.root.items()
 
--- a/src/ocf_data_sampler/datasets/cache.py
+++ b/src/ocf_data_sampler/datasets/cache.py
@@ -6,13 +6,20 @@ import pickle
 class PickleCacheMixin:
     """A mixin for classes that need to cache their state using pickle."""
 
-    def __init__(self, *args: list, **kwargs: dict) -> None:
+    def __init__(self, *args: object, **kwargs: object) -> None:
         """Initialize the pickle path and call the parent constructor."""
         self._pickle_path = None
         super().__init__(*args, **kwargs)  # cooperative multiple inheritance
 
     def presave_pickle(self, pickle_path: str) -> None:
-        """Save the full object state to a pickle file and store the pickle path."""
+        """Save the full object state to a pickle file and store the pickle path.
+
+        The object is then pickled by reference, so the file must be readable wherever it is
+        unpickled. The state is a snapshot - call this again after mutating the object.
+
+        Args:
+            pickle_path: Where to write the object state to.
+        """
         self._pickle_path = pickle_path
         with open(pickle_path, "wb") as f:
             pickle.dump(self.__dict__, f)
@@ -22,12 +29,23 @@ class PickleCacheMixin:
         if self._pickle_path:
             return {"_pickle_path": self._pickle_path}
         else:
-            return self.__dict__
+            # Copied so that the pickled state can't be mutated through the live object
+            return dict(self.__dict__)
 
     def __setstate__(self, state: dict) -> None:
         """Restore object from pickle, reloading from presaved file if possible."""
         self.__dict__.update(state)
-        if self._pickle_path and os.path.exists(self._pickle_path):
-            with open(self._pickle_path, "rb") as f:
-                saved_state = pickle.load(f)  # noqa: S301
-                self.__dict__.update(saved_state)
+
+        if not self._pickle_path:
+            return
+
+        if not os.path.exists(self._pickle_path):
+            raise FileNotFoundError(
+                f"Presaved state file not found: {self._pickle_path}. This object was pickled by "
+                "reference to that path - see `presave_pickle`.",
+            )
+
+        with open(self._pickle_path, "rb") as f:
+            saved_state = pickle.load(f)  # noqa: S301
+
+        self.__dict__.update(saved_state)
--- a/src/ocf_data_sampler/datasets/cache.py
+++ b/src/ocf_data_sampler/datasets/cache.py
@@ -14,8 +14,10 @@ class PickleCacheMixin:
     def presave_pickle(self, pickle_path: str) -> None:
         """Save the full object state to a pickle file and store the pickle path.
 
-        The object is then pickled by reference, so the file must be readable wherever it is
-        unpickled. The state is a snapshot - call this again after mutating the object.
+        The object will be pickled by reference after calling this function, so the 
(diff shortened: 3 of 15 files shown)


src/ocf_data_sampler/config/model.py
```diff
@@ -280,13 +281,12 @@ def check_all_channel_have_normalisation_constants(self) -> "NWP":
...
 class MultiNWP(RootModel):
-    """Configuration for multiple NWPs."""
+    """Configuration for multiple NWPs.
 
-    root: dict[str, NWP]
+    The NWP sources are accessed dict-style - `config.nwp["ukv"]`, not `config.nwp.ukv`.
+    """
 
-    @override
-    def __getattr__(self, item: str) -> NWP:
```
dfulu, line 288: we don't use this and don't need it

src/ocf_data_sampler/config/model.py
```diff
@@ -300,11 +300,11 @@ def __len__(self) -> int:
     def __iter__(self) -> Iterator:
         return iter(self.root)
 
-    def keys(self) -> Iterator[str]:
+    def keys(self) -> KeysView[str]:
```
dfulu, line 303: the old type hints were wrong

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -291,6 +298,8 @@ def __init__(
             # Filter t0 times to given range
             if time_periods is not None:
                 mask = get_time_periods_mask(valid_t0_times, time_periods)
+                if not mask.any():
+                    raise ValueError(f"`time_periods` {time_periods} excluded all valid t0 times")
```
dfulu, line 302: We raise this here instead of letting the dataset be empty and raise later further away from the source

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -362,6 +373,13 @@ def find_valid_t0_times(
             valid_time_periods,
             freq=minutes(config.sampling_grid.t0_resolution_minutes),
         )
+
+        if len(valid_t0_times) == 0:
+            raise ValueError(
```
dfulu, line 378: This catches a separate error where there are some time periods available but none intersect with the t0 resolution. As in we could have a time period 00:10 -> 00:20 but if we have time resolution 30mins then we get no valid t0s

src/ocf_data_sampler/datasets/pvnet/dataset.py
dfulu, line 386: I just did a bit of refactoring and cleaning up in this function

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -381,15 +399,19 @@ def find_valid_t0_and_location_ids(
             locations: The locations to find valid t0 times for
             config: PVNetDataConfig file
         """
-        # Get valid time period for nwp and satellite
-        datasets_without_generation = {k: v for k, v in datasets_dict.items() if k != "generation"}
-        valid_time_periods = find_valid_time_periods(datasets_without_generation, config)
+        # Get valid time period for inputs other than generation
```
dfulu:
```suggestion
        # Get valid time periods for inputs other than generation
```

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -408,22 +430,29 @@ def find_valid_t0_and_location_ids(
...
+            valid_t0_times = fill_time_periods(
+                valid_time_periods,
                 freq=minutes(config.sampling_grid.t0_resolution_minutes),
             )
 
-            valid_t0_per_location = pd.DataFrame(index=valid_t0_times_per_location)
-            valid_t0_per_location["location_id"] = location.id
-            valid_t0_and_location_ids.append(valid_t0_per_location)
+            if len(valid_t0_times) == 0:
+                logger.warning(f"No valid t0 times found for location {location.id}")
```
dfulu, line 444: I think this is good to warn about since the model could silently train without ever seeing this location

dfulu: This could have been an error but maybe that's too strict

README.md
```diff
@@ -22,7 +22,7 @@ It includes and adapts internal functions from the Google xarray-tensorstore pro
 
 ## Documentation
 
-**ocf-data-sampler** doesn't have external documentation _yet_; you can read a bit about how our torch datasets work in the README [here](ocf_data_sampler/torch_datasets/README.md).
+**ocf-data-sampler** doesn't have external documentation _yet_; you can read a bit about how our torch datasets work in the README [here](src/ocf_data_sampler/datasets/pvnet/README.md).
```
dfulu, line 25: This was out of date

src/ocf_data_sampler/datasets/cache.py
```diff
@@ -22,12 +29,23 @@ def __getstate__(self) -> dict:
...
             return {"_pickle_path": self._pickle_path}
         else:
-            return self.__dict__
+            # Copied so that the pickled state can't be mutated through the live object
+            return dict(self.__dict__)
 
     def __setstate__(self, state: dict) -> None:
         """Restore object from pickle, reloading from presaved file if possible."""
         self.__dict__.update(state)
-        if self._pickle_path and os.path.exists(self._pickle_path):
```
dfulu, line 30: There was a chance here that this would fail weirdly if the pickled dataset was deleted or unavailable. So we handle this more explicitly now

src/ocf_data_sampler/datasets/pvnet/valid_t0s.py
```diff
@@ -62,9 +62,6 @@ def find_valid_time_periods(
                 max_staleness=max_staleness,
             )
 
-            if len(time_periods) == 0:
```
dfulu, line 65: The `find_contiguous_t0_periods[_nwp]` functions raise their own errors if no periods are found, so we never reach  these

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -109,10 +111,15 @@ def get_time_periods_mask(
 
     for start_time, end_time in time_periods:
 
-        start_time = times[0] if start_time is None else np.datetime64(start_time)
-        end_time = times[-1] if end_time is None else np.datetime64(end_time)
```
dfulu, line 113: The new version is slightly more robust and general since it doesn't rely on `times` being sorted and the `[-1]` indexer working on whatever input type goes in. This is so marginal that I'd wonder if its worth the code churn, so I'm happy to revert if we want. 

One change here that we should have had is to be exclusive of the end point. A common pattern would be 
```
train_period=[(None, "2025-01-01 00:00")]
val_period=[("2025-01-01 00:00", None)]
```

This stops both train and val from containing the `"2025-01-01 00:00"` t0 time

src/ocf_data_sampler/datasets/pvnet/valid_t0s.py
dfulu: Not reaching these was a bug, but perhaps we want to reach these. But that would require us to define what `find_contiguous_t0_periods[_nwp]` returns if there are no time periods. Is it just an empty dataframe?

src/ocf_data_sampler/features/time_encodings.py
```diff
@@ -27,14 +27,16 @@ def encode_datetimes(datetimes: NDArray[np.datetime64]) -> dict[str, NDArray[np.
...
     time_in_radians = (2 * np.pi) * day_fraction
-    date_in_radians = (2 * np.pi) * (day_of_year / 365)
+    date_in_radians = (2 * np.pi) * ((day_of_year - 1) / days_in_year)
 
     return {
-        "date_sin": np.sin(date_in_radians),
-        "date_cos": np.cos(date_in_radians),
-        "time_sin": np.sin(time_in_radians),
-        "time_cos": np.cos(time_in_radians),
+        "date_sin": np.sin(date_in_radians).astype(np.float32),
```
dfulu, line 36: These were returning float64 which is not what the function signature advertises. We also cast to float32 for the TensorBatch passed to the model

tests/datasets/pvnet/test_dataset.py
```diff
@@ -104,16 +106,27 @@ def test_get_time_periods_mask():
             ("2023-01-01 12:00", "2023-01-01 13:00"),
         ],
     )
-    expected_mask = np.array([False, True, True, True, False, True, True, True])
```
dfulu, line 108: The end point is now exclusive

src/ocf_data_sampler/load/nwp.py
```diff
@@ -68,6 +70,22 @@ def open_nwp(zarr_path: ZarrSource, provider: str) -> xr.DataArray:
     return da
 
 
+def _assert_steps_uniformly_spaced(steps: NDArray[np.timedelta64], provider: str) -> None:
```
felix-e-h-p, line 73: This will cause a hard fail with uneven steps say for MO right?

src/ocf_data_sampler/datasets/pvnet/dataset.py
```diff
@@ -305,8 +314,10 @@ def __init__(
 
             # Filter t0 times to given range
             if time_periods is not None:
-                mask = get_time_periods_mask(valid_t0_and_location_ids["t0"], time_periods)
-                valid_t0_and_location_ids = valid_t0_and_location_ids[mask]
+                mask = get_time_periods_mask(valid_t0_and_location_ids["t0"].values, time_periods)
+                if not mask.any():
```
felix-e-h-p, line 318: Nice one!

src/ocf_data_sampler/datasets/pvnet/dataset.py
felix-e-h-p: Was always silent / empty before right? Good one!

src/ocf_data_sampler/select/time_periods.py
```diff
@@ -150,6 +150,14 @@ def find_contiguous_t0_periods_nwp(
     else:
         init_end_timedelta = min(last_forecast_step - interval_end, max_staleness)
 
+    # Each init-time is unusable before init_start_timedelta and after init_end_timedelta. If they
+    # are the wrong way round then no t0 can ever use it, and every period below will be empty
+    if init_end_timedelta < init_start_timedelta:
```
felix-e-h-p, line 155: Solid one, clear flagging

src/ocf_data_sampler/select/spatial_slice.py
```diff
@@ -166,6 +166,9 @@ def select_spatial_slice_pixels_multiple(
     Returns:
         The selected DataArray-like slice.
     """
+    if len(locations) == 0:
```
felix-e-h-p, line 169: For any window inversion errors right?

felix-e-h-p (changes requested):
Thanks for this! Looking great. Think the one thing from me would be if the uniformly spaced steps assertion could reject archives, but all good asides that.

src/ocf_data_sampler/load/nwp.py
dfulu: Yeh, that's the intention. We don't support uneven NWP steps in data-sampler so I wanted to make that explicit. It's caused a bunch of hard-to-trace errors in the past so I think its worth failing early and explicitly instead

src/ocf_data_sampler/select/spatial_slice.py
dfulu: Well I'm intending it to be a public function so its to just catch any user errors in the input

felix-e-h-p (approved):

